PR #216, opened 2025-02-25
support sp4&sp8 for sta

GOATmessi8:
Optimize the current sp implementation to support sp4&sp8


SolitaryThinker:
Thanks, however closing this as we have now migrated to V1

PR #280, opened 2025-03-24
Attn Abstraction

jzhang38:


SolitaryThinker:
included in https://github.com/hao-ai-lab/FastVideo/pull/282. 
reopen if this is not the case

PR #291, opened 2025-03-27
Add Wan VAE & T5 Text Encoder

JerryZhou54:


JerryZhou54:
A couple changes to the "common" files used by all models:
1. Add NewGELU back to the layers/activation.py file, as it is needed by the T5 encoder.
2. Change the "blend_num_frames" variable to a global variable and set it during VAE initialization
3. Call the init function of ParallelTiledVAE at the end of hunyuanVAE's init function in order to set the blend_num_frames variable

PR #302, opened 2025-03-30
Set up text encoder tests to work with pytest and Github Actions

kevin314:
Changed test_clip_encoder.py and test_llama.encoder.py to work with pytest:
- Adds local_dir param to maybe_download_model to download model during test execution
- Sets MASTER_ADDR and MASTER_PORT as environment vars in test script, avoid need for torchrun
- Hardcode inference args
- Adds fixture for distributed env setup/teardown between tests

Usage:
`pytest ./fastvideo/v1/tests/test_clip_encoder.py -s`
`pytest ./fastvideo/v1/tests/test_llama_encoder.py -s`

Both tests:
`pytest ./fastvideo/v1/tests/ --ignore=./fastvideo/v1/tests/old_tests/ -s`
\
\
Added a Github Actions workflow for running tests on PRs with RunPod (only works with the two encoder tests currently):

- Makes requests to RunPod to create a pod with specified resources (GPU type, count, volume/disk size, etc.) for the duration of the workflow job
- Created instance will run commands specified in the job

image


SolitaryThinker (approved):
We might be running in to some consistency issues with secrets. Going to merge and open new PR to test

zhisbug:
Just seeking clarification: So basically the GPU runner pod is created on-the-fly when a CI job is launched?

kevin314:
@zhisbug When a CI job is launched, the Github "runner" instance that executes the job is created first for free by Github, and then that runner creates the RunPod instance with GPUs which are shut down after the job finishes

PR #309, opened 2025-04-02
refactor the env setup and install of fastvideo

PorridgeSwim:
Support optional lint install.
Remove sta_attn dependency.
Debug env_setup.py (flash_attn should has wheel preinstalled).


SolitaryThinker:
Thank you for the effort!
Quick question, is it not possible to completely remove the env_setup.sh script? why?

SolitaryThinker:
I don't like the use of scripts to install flash-attn or sta. I am going to take a look at whether we just put these two dependencies as optional under the `.toml` file. I will directly modify this PR and work on this.

SolitaryThinker:
Okay, I checked xDit/xfuers, which depends on flash-attn package, and their instructions in README actually fails if you start with a clean conda env. 

The best way forward is to add flash-attn instruction in our README if people want to install it and fall back to torch sdpa backend if flash-attn package is not available (already being done) 

cc @jzhang38 @PorridgeSwim

zhisbug:
> Okay, I checked xDit/xfuers, which depends on flash-attn package, and their instructions in README actually fails if you start with a clean conda env.
> 
> The best way forward is to add flash-attn instruction in our README if people want to install it and fall back to torch sdpa backend if flash-attn package is not available (already being done)
> 
> cc @jzhang38 @PorridgeSwim

agree. Let's do that then. To summarize:
1. README should be explict in suggesting to `pip install flash-attn`
2. The attention backend will check fa existence and warn if not existed
3. if fa is not installed, fall back to use torch attention.

jzhang38:
Yes, I will set the default backend to SDPA in my PR: https://github.com/hao-ai-lab/FastVideo/pull/312

SolitaryThinker (approved):
merging to unblock ci/cd work

PR #311, opened 2025-04-03
[CI] Add test workflow improvements

kevin314:
To avoid triggering all workflows and jobs for all PRs, this change adds filters for triggering the pr-test workflow, as well as job level filters for running the encoder test job only if the v1/models/encoder/ or v1/models/loader/ files are changed.

- Workflow will trigger if the workflow file itself changes

\
\
Additionally, a new job `runpod-cleanup` is added as a failsafe in the event that any test jobs are unable to terminate their RunPod instances, and pods become stranded. This may happen due to the new timeout for the encoder test, which was added to prevent indefinitely running tests or hanging Github runners.

- This job will not touch any pods other than the ones started in the same workflow--pods are identified by their {JOB ID}-{WORKFLOW ID}
- This job only runs after all other test jobs have passed/failed/skipped/timed out, and will not interrupt running tests

- Normally this cleanup job will not find any pods to shutdown, as the test jobs will try to terminate their pods themselves


SolitaryThinker (approved):
Lgtm

PR #312, opened 2025-04-04
Add STA to V1

jzhang38:
- [x] add fa3 backend
- [x] remove all attn dropout args (no one use dropout in 2025, fa3 do not even support dropout)
- [x] add supported_attn_backend attribute for each model(so text encoder will avoid using STA even if we set the attn backend to STA & WAN will not call STA if we did not figure out a good mask strategy)
- [x] add prefix for dit module, so each attn knows which layer is it located at.
- [x] add sta
- [x] make sure changes does not break other functions (such as Wan)
- [x] make sure there is no lining error.


SolitaryThinker:
pretty sure hf is just down....

SolitaryThinker (approved):

PR #314, opened 2025-04-04
Add ssim test

kevin314:
Costs to run this job with different GPU setups (large percentage is from installing dependencies):
- 2 x A40: $0.33
- 1 x A100: $0.67

This was brought up previously--output between different devices varies slightly, leading to a SSIM of <1.0 with other generation input args being the same.


SolitaryThinker:
Can you please add a readme.md to the `generated_videos/` to track which version(commit) of FastVideo and configs were used to generate the reference videos

.github/workflows/pr-test.yml
```diff
@@ -71,18 +73,56 @@ jobs:
           --gpu-type "NVIDIA A40"
           --gpu-count 1
           --volume-size 100
-          --test-command "./env_setup.sh &&
-          pip install vllm &&
-          pip install pytest &&
-          pytest ./fastvideo/v1/tests/ --ignore=./fastvideo/v1/tests/old_tests/ -s"
+          --test-command "pip install -e .[test] &&
+          pip install flash-attn==2.7.0.post2 &&
```
zhisbug: I think at some point we would need use docker. Every time running `pip install flash-attn` takes very long

README.md
```diff
@@ -38,18 +38,25 @@ Dev in progress and highly experimental.
...
 ```
-./env_setup.sh fastvideo
+# Clone FastVideo
+git clone https://github.com/hao-ai-lab/FastVideo.git && cd FastVideo
+
+# Install FastVideo
+pip install -e .
+
+# Install Flash Attention (optional)
+pip install flash-attn==2.7.0.post2
```
zhisbug, line 51: is the version right? please double check

zhisbug (approved):

.github/workflows/pr-test.yml
kevin314: If we find a suitable image, it will be simple to have RunPod use it for creating a pod:
https://github.com/hao-ai-lab/FastVideo/blob/4324c1c21d121e179032c97ad2738d5dfbe0ce8a/.github/scripts/runpod_api.py#L27-L31

README.md
SolitaryThinker: This is the same as what was in `env_setup.sh` previously.

.github/workflows/pr-test.yml
SolitaryThinker: There a set of pre-built wheels for each version of flash-attn release uploaded to github. If the environment (torch version, python version, cuda version) matches one of the pre-built wheel configurations, pip only needs to download the wheel and package it. However if installing in an environment that is not pre-built, it will require full compilation. We explicitly pin torch version so that the version of flash-attn we use does not need to be compiled from source. See below:
image

PR #315, opened 2025-04-06
Fix sdpa

jzhang38:
A small bug in sdpa backend. Needs to toggle on GQA manually,


SolitaryThinker (approved):
tested manually, lgtm. Will add this backend to ssim test

PR #307, opened 2025-04-01
Add CFG & Wan Pipeline

JerryZhou54:

PR #318, opened 2025-04-07
Add tests for wan

JerryZhou54:


JerryZhou54:
Only new T5 encoder, Wan VAE, WanTransformer from #307 can pass these tests

JerryZhou54:
Close this as the tests for wan are contained in #326 as well.

PR #321, opened 2025-04-07
[CI] Use pre-commit to run linter

SolitaryThinker:
Docs and details will be added in next PR


zhisbug:
good! So when we do "git commit" it will trigger precommit in local?

SolitaryThinker:
Yes linters will automatically run and needs to pass before commits are made, devs will need to run `pre-commit install --hook-type pre-commit --hook-type commit-msg` to install the hooks. 

I will be working on docs after this as CI should be green after this and will add details to doc

SolitaryThinker:
Going to merge to unblock people, but will continue looking into flaky ssim test

kevin314 (approved):

PR #329, opened 2025-04-09
[CI] Set allowedCudaVersions

kevin314:
It seems not all RunPod GPUs support CUDA 12.4, and you have to specify which versions you allow when creating an instance to guarantee support.

Also disables output buffering on subprocess for more frequent logging.


SolitaryThinker (approved):
thanks

PR #333, opened 2025-04-10
Port tests to v1

kevin314:
TODO:

- [x] Run loader test with adequate GPU to find latent and fix test
- [x] Add jobs to PR test workflow to run transformer, loader, and vae tests
- [ ] Mark reference latents with commit hash


SolitaryThinker:
please add `# SPDX-License-Identifier: Apache-2.0` to the top of all files

SolitaryThinker:
One you add the reference commit has it should be good to go!
cc @jzhang38 please take a pass

.github/workflows/pr-test.yml
```diff
@@ -49,6 +61,12 @@ jobs:
               - 'fastvideo/v1/models/encoders/**'
               - 'fastvideo/v1/models/loaders/**'
               - 'fastvideo/v1/tests/encoders/**'
+            vae-test:
+              - 'fastvideo/v1/models/vaes/**'
+              - 'fastvideo/v1/tests/vaes/**'
+            transformer-test:
+              - 'fastvideo/v1/models/dits/**'
+              - 'fastvideo/v1/tests/transformers/**'
```
SolitaryThinker: similar to the encoder test, also run these if fastvideo/v1/models/loaders/**' change

SolitaryThinker (changes requested):

SolitaryThinker:
Note: test_hunyuanvideo is redundant since we have test_hunyuanvideo_load, but lets keep it for this PR and remove it later so it is in the git history as a reference

PR #335, opened 2025-04-11
V1 wan rebased

SolitaryThinker:


SolitaryThinker:
Thanks for the effort. Just waiting for @kevin314 to take a look at the CI and test changes and we are ready to merge!

kevin314 (changes requested):
I've noticed that setting the seed at the module level of a test file can affect the result of subsequent test files, when multiple test files are run together. The result is not affected when the seed is set within the test function, so I've since moved the hardcoded seed to the conftest.py fixture, and you should be able to remove
`torch.manual_seed(42)
np.random.seed(42)`
from the t5 encoder and wan vae tests

JerryZhou54:
> I've noticed that setting the seed at the module level of a test file can affect the result of subsequent test files, when multiple test files are run together. The result is not affected when the seed is set within the test function, so I've since moved the hardcoded seed to the conftest.py fixture, and you should be able to remove `torch.manual_seed(42) np.random.seed(42)` from the t5 encoder and wan vae tests
> 
> Other than that the tests/CI changes look good to me

Addressed

kevin314 (approved):

PR #338, opened 2025-04-14
[Model] Remove RMSNorm's forward_native hardocode from Wan

SolitaryThinker:
Currently RMSNorm is hardcoded to always use forward_native instead of letting customOps pick (will use cuda_forward instead)

- [ ] Pass lint
- [ ] Update SSIM test


SolitaryThinker:
replaced by https://github.com/hao-ai-lab/FastVideo/pull/339

PR #339, opened 2025-04-14
[Model] Remove RMSNorm's forward_native hardcode from Wan

SolitaryThinker:
Currently RMSNorm is hardcoded to always use forward_native instead of letting customOps pick (will use cuda_forward instead)

- [x] Turns out the linters were skipped a lot of files and adding a `__init__.py` to `fastvideo/` caused it to run linting on more files.
- [x] Pass lint
- [x] Update SSIM test


JerryZhou54 (approved):
No issues found

PR #342, opened 2025-04-15
[CI] Support custom Docker image

kevin314:
Adds manual workflow for building and pushing images to the GitHub Container Registry. PR test jobs will now have RunPod pull these images to create containers. Uncached, this takes around 6-7 minutes. Apparently, RunPod supports local layer caching in all its data centers, so hopefully subsequent pods using the same image in the same region/machine will be able to start up much faster.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -14,6 +14,11 @@ on:
       - ".github/workflows/pr-test.yml"
   workflow_dispatch:
     inputs:
+      custom_image:
+        description: "Custom image from this repository (default: fastvideo-dev:latest)"
+        required: false
+        default: "fastvideo-dev:latest"
+        type: string
       run_encoder_test:
         description: "Run encoder-test"
         required: false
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-test.yml
```diff
@@ -107,9 +115,8 @@ jobs:
           --gpu-type "NVIDIA A40"
           --gpu-count 1
           --volume-size 100
-          --test-command "pip install -e .[test] &&
-          pip install flash-attn==2.7.0.post2 --no-build-isolation &&
-          pytest ./fastvideo/v1/tests/encoders -s"
+          --image "ghcr.io/${{ github.repository }}/${{ github.event.inputs.custom_image || 'fastvideo-dev:latest' }}"
+          --test-command "pip install -e . && pytest ./fastvideo/v1/tests/encoders -s"
```
SolitaryThinker, line 119: shouldn't this be `pip install -e .[dev]` and then all of these dependencies should already be installed in the docker image right?

SolitaryThinker (approved):
thanks

PR #350, opened 2025-04-16
[CI] Docker image improvements

kevin314:
This change makes the image a lot more ready-to-use as a standalone image for FastVideo development--shells created in a container will automatically have the correct conda environment activated, and the copied repo is ready to use after a pull from remote for the latest changes.

The repo will be available at /FastVideo on container start, which may be then moved to a persistent volume directory like /workspace on RunPod if needed. Of course, you can still clone the repo from remote where you need it if preferred. If a pod is stopped, all dependencies will remain once resumed, even with no volume storage at all, allowing for quick/easy access to a dev environment.

The change to pr-test.yml is more of an extra safeguard against an outdated latest image when running CI, if for some reason the test dependencies changed but a new image hasn't been built/pushed, the test will still use the version it expects.

If you want to use the image (`ghcr.io/hao-ai-lab/fastvideo/fastvideo-dev:latest`) yourself on RunPod, be sure to add `bash -c "apt update;DEBIAN_FRONTEND=noninteractive apt-get install openssh-server -y;mkdir -p ~/.ssh;cd $_;chmod 700 ~/.ssh;echo \"$PUBLIC_KEY\" >> authorized_keys;chmod 700 authorized_keys;service ssh start;sleep infinity"` to the docker start cmd to be able to SSH into the container properly.


SolitaryThinker (approved):
Could you add the Tip for Runpod Development to a new section in the Docs for  Contributor Guide?

SolitaryThinker:
looks great! Thanks!

PR #356, opened 2025-04-19
[V1] Process aware logging; improve logging msg

SolitaryThinker:
After this PR, by default, `logger.info` will only print from main process after printing a warning. 
Requires `LOCAL_RANK` and `RANK` envar to be set -- available for both MP backend and `torchrun`.

Also disables tqdm progress bar for none rank 0 processes.

MP backend:
image

torchrun script:
image


fastvideo/v1/logger.py
```diff
@@ -68,6 +75,69 @@ def _print_warning_once(logger: Logger, msg: str) -> None:
...
+    
+    Note:
+        - When both main_process_only and local_main_process_only are True, 
+          the message will be logged only if both conditions are met
+        - When both are False, the message will be logged from all processes
+        - By default, only logs from processes with LOCAL_RANK=0
+    """
+    try:
+        # TODO(will): don't use env variable
+        local_rank = int(os.environ["LOCAL_RANK"])
```
jzhang38, line 107: get_world_group().local_rank？

jzhang38 (approved):

jzhang38:
overall lgtm

fastvideo/v1/logger.py
SolitaryThinker: Took another look. Using envars is probably the cleanest since importing from `distributed` cause circular import issues. Going to leave it as it is and remove the TODOs.

PR #358, opened 2025-04-20
Model config

JerryZhou54:
RFC [here](https://docs.google.com/document/d/1oLNrt-tPuyJVNZwWg3RSNuiGLyE2QJnXUz63Zr68OJY/edit?usp=sharing)


fastvideo/v1/configs/models/base.py
```diff
@@ -0,0 +1,47 @@
...
+    ) -> None:
+        arch_config = self.arch_config
+        valid_fields = {f.name for f in fields(arch_config)}
+        
+        for key, value in source_model_dict.items():
+            if key in valid_fields:
+                setattr(arch_config, key, value)
+            else:
+                raise AttributeError(f"{type(arch_config).__name__} has no field '{key}'")
+
```
SolitaryThinker: One way to enforce the fact that users shouldn't modify `ArchConfig` is to use `@dataclass(frozen=True)` (inherited classes from this also needs `frozen`) And then in this method use `dataclasses.replace()`. If you expect this to be modified internally by fastvideo then this may not work

fastvideo/v1/configs/models/vaes/base.py
```diff
@@ -0,0 +1,36 @@
...
+import torch
+
+from fastvideo.v1.configs.models import ArchConfig, ModelConfig
+
+@dataclass
+class VAEArchConfig(ArchConfig):
+    scaling_factor: Union[float, torch.tensor] = 0
+
+    temporal_compression_ratio: int = 4
+    spatial_compression_ratio: int = 8
```
SolitaryThinker, line 14: where are these originally set for each of the vae for each pipeline? I don't see these in the `config.json` on HF

fastvideo/v1/configs/models/vaes/wanvae.py
```diff
@@ -0,0 +1,72 @@
...
+    attn_scales: Tuple[float, ...] = ()
+    temperal_downsample: Tuple[bool, ...] = (False, True, True)
+    dropout: float = 0.0
+    latents_mean: Tuple[float, ...] = (
+        -0.7571,
+        -0.7089,
+        -0.9113,
+        0.1075,
+        -0.1745,
+        0.9653,
```
SolitaryThinker, line 24: For these, is there a reason we also need to store these values in FastVideo's repo? Can we not just load from the HF config?

SolitaryThinker: Do these vary between different weight versions? ie hunyuan vs fasthunyuan?

fastvideo/v1/configs/pipelines/base.py
```diff
@@ -0,0 +1,103 @@
...
+    disable_autocast: bool = False
+
+    # Model configuration
+    precision: str = "bf16"
+
+    # VAE configuration
+    vae_precision: str = "fp16"
+    vae_tiling: bool = True
+    vae_sp: bool = True
+    # vae_scale_factor: Optional[int] = None # Deprecated
```
SolitaryThinker: Shouldn't these be moved to VAEConfig() now?

fastvideo/v1/configs/models/base.py
```diff
@@ -0,0 +1,47 @@
...
+    pass
+
+@dataclass
+class ModelConfig:
+    # Every model config parameter can be categorized into either ArchConfig or everything else
+    # Diffuser/Transformer parameters
+    arch_config: ArchConfig = ArchConfig()
+
+    # FastVideo-specific parameters here
+    # i.e. STA, quantization, teacache
```
SolitaryThinker, line 24: Can you add a prototype for STA? (doesn't have to be complete or working) I just want to see how it would work with the design

fastvideo/v1/configs/pipelines/base.py
```diff
@@ -0,0 +1,103 @@
...
+    window_size: int = 16
+    stride: int = 8
+
+    # You can provide custom defaults for inherited fields
+    height: int = 576
+    width: int = 1024
+
+    # Additional configuration specific to sliding tile attention
+    pad_to_square: bool = False
+    use_overlap_optimization: bool = True
```
SolitaryThinker: Is this still the old code? Could you turn/move this to be a prototype for configuring STA? (see earlier comment)

fastvideo/v1/configs/wan_14B_i2v_480p_pipeline.json
SolitaryThinker, line 1: Are these examples of dumping the config?

fastvideo/v1/distributed/__init__.py
SolitaryThinker, line 1: errors in this file are because you also (unsuccessfully) rebased some stale commits from an old branch

SolitaryThinker (commented):
Overall this is great! We can have a sync tomorrow or early next week.

I think one thing we both may have missed is to also properly separate out generation time args (passed to `generator.generate_video()` and initialization args (passed to `generator.from_pretrained()`)

fastvideo/v1/configs/models/base.py
```diff
@@ -0,0 +1,59 @@
...
+    ) -> None:
+        assert "arch_config" not in source_model_dict.keys(), "Source model config shouldn't contain arch_config."
+        
+        valid_fields = {f.name for f in fields(self)}
+
+        for key, value in source_model_dict.items():
+            if key in valid_fields:
+                setattr(self, key, value)
+            else:
+                print(f"{type(self).__name__} does not contain field '{key}'!")
```
SolitaryThinker: Can you use logger here instead?

SolitaryThinker (approved):
- please expose config APIs under `fastvideo/__init__.py`
- disable hunyuan dit test

fastvideo/v1/configs/pipelines/wan.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # Denoising stage
+    flow_shift: int = 3
+
+    # Text encoding stage
+    text_encoder_config: EncoderConfig = T5Config()
+
+    # Precision for each component
+    precision: str = "bf16"
+    vae_precision: str = "fp16"
```
Edenzzzz, line 33: These  vae precisions are all wrong. Reviewers and PR author should verify this kind of thing against huggingface

PR #360, opened 2025-04-21
chore: Release FastVideo 0.0.2 and update python requirements

SolitaryThinker:
Tested that FastVideo also works on Python 3.11 and 3.12


BrianChen1129 (approved):

SolitaryThinker:
okay lets fix linter for minimum python version later. Reverted minimum version in pyproject to 3.8 to keep linter happy.

PR #371, opened 2025-05-02
Add Teacache to V1

SolitaryThinker:
- [x] wan2.1
- [ ] hunyuan: Leaving to later patch as it involves more work
- [x] cleanup


fastvideo/v1/configs/sample/base.py
JerryZhou54, line 1: Can we put all teacache params in a separate file called "sample/teacache.py"?

fastvideo/v1/configs/sample/hunyuan.py
JerryZhou54, line 1: maybe move "enable_teacache" into the base SamplingParam?

fastvideo/v1/configs/sample/wan.py
JerryZhou54, line 1: probably need to rename WanTeaCacheParams at some point

fastvideo/v1/entrypoints/video_generator.py
JerryZhou54, line 1: why do we need this temporary hacks?

SolitaryThinker: good catch, removing

PR #378, opened 2025-05-05
Small Fixes & Features

JerryZhou54:
- [x] Add num_frames adjustment for multi-gpu setting
- [x] T5 encoder test issues when precision is fp32
- [x] Enable torch compile for single GPU. Multi-gpu not compatible right now due to FSDP.
- [x] Clear CUDA memory between each run

### DIFF
--- a/fastvideo/v1/configs/pipelines/base.py
+++ b/fastvideo/v1/configs/pipelines/base.py
@@ -55,6 +55,8 @@ class PipelineConfig:
 
     # STA (Spatial-Temporal Attention) parameters
     mask_strategy_file_path: Optional[str] = None
+
+    # Compilation
     enable_torch_compile: bool = False
 
     @classmethod
--- a/fastvideo/v1/pipelines/stages/decoding.py
+++ b/fastvideo/v1/pipelines/stages/decoding.py
@@ -7,7 +7,6 @@ import torch
 
 from fastvideo.v1.fastvideo_args import FastVideoArgs
 from fastvideo.v1.logger import init_logger
-from fastvideo.v1.models.vaes.common import ParallelTiledVAE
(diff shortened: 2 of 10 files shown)


fastvideo/v1/pipelines/stages/decoding.py
```diff
@@ -95,4 +96,7 @@ def forward(
         if hasattr(self, 'maybe_free_model_hooks'):
             self.maybe_free_model_hooks()
 
+        self.vae.to("cpu")
+        torch.cuda.empty_cache()
+
```
SolitaryThinker, line 101: Shouldn't this be wrapped with `fastvideo_args.use_cpu_offload`?

SolitaryThinker (approved):
Thanks! please rebase and this can be merged

fastvideo/v1/configs/pipelines/base.py
```diff
@@ -55,6 +55,8 @@ class PipelineConfig:
 
     # STA (Spatial-Temporal Attention) parameters
     mask_strategy_file_path: Optional[str] = None
+
+    # Compilation
     enable_torch_compile: bool = False
```
SolitaryThinker, line 60: perhaps we can just default this to True? and then your check can catch it for mult GPU

fastvideo/v1/pipelines/stages/decoding.py
JerryZhou54: For GPU with small memory (i.e. L4), vae must be offloaded otherwise OOM issue will occur for the next run.

fastvideo/v1/configs/pipelines/base.py
JerryZhou54: yea sure

PR #380, opened 2025-05-06
[feat] add flux pipeline to v1

Eigensystem:
- [x] add pipelines and modules
- [x] debug accuracy issues
- [x] refactor some hard code in stages
- [x] refactor custom stages & remove FluxDenoiseStage & OutputStage
- [x] add a `image_generate` func in `VideoGenerator`


fastvideo/v1/tests/transformers/test_flux.py
```diff
@@ -0,0 +1,151 @@
...
+        torch_dtype=precision).to(device, dtype=precision).requires_grad_(False)
+
+    total_params = sum(p.numel() for p in model1.parameters())
+    # Calculate weight sum for model1 (converting to float64 to avoid overflow)
+    weight_sum_model1 = sum(
+        p.to(torch.float64).sum().item() for p in model1.parameters())
+    # Also calculate mean for more stable comparison
+    weight_mean_model1 = weight_sum_model1 / total_params
+    logger.info("Model 1 weight sum: %s", weight_sum_model1)
+    logger.info("Model 1 weight mean: %s", weight_mean_model1)
```
Edenzzzz, line 56: I feel there's no need to print the sums or means, only the diffs are useful

Eigensystem: Okay. I wrote this test by referring to another file, fastvideo/v1/tests/transformers/test_wanvideo.py. Do I also need to remove the sum printing in this file as well?

Edenzzzz: Possibly, it's more of a style thing, but most people just use `torch.testing.assert_close` to avoid printing too messily😂

Eigensystem:
move to #419

PR #384, opened 2025-05-06
[CLI] Update cli to support new api/model config

kevin314:


fastvideo/v1/default_configs/v1_inference_hunyuan_config.json
SolitaryThinker, line 1: Let's not provide these files anymore since we've added the model config system. If users want their cli json, they can dump their own

SolitaryThinker (approved):

PR #390, opened 2025-05-07
[Misc] Small Fixes & Features

JerryZhou54:
- [x] Adapt torch compile to teacache logic. Compile each block instead of the whole model.
- [x] Add config and sample params for Wan I2V 720P
- [x] Add the option to round down num_frames (when it's not divisible by num_gpus)


SolitaryThinker (approved):
Thanks

PR #392, opened 2025-05-09
[CI] Use python 3.10/3.11 for SSIM test

kevin314:
Also adds runpod-test.yml as reusable template for all tests


SolitaryThinker (approved):
thanks

PR #395, opened 2025-05-11
[Misc] Small fixes to Torch code

applesaucethebun:
Remove deprecated https://docs.pytorch.org/docs/stable/generated/torch.norm.html as well as `torch.cuda.amp.autocast` which is also deprecated

And minor docstrings/CLI args.

### DIFF
--- a/fastvideo/models/stepvideo/text_encoder/stepllm.py
+++ b/fastvideo/models/stepvideo/text_encoder/stepllm.py
@@ -267,7 +267,7 @@ class Step1Model(PreTrainedModel):
 class STEP1TextEncoder(torch.nn.Module):
 
     def __init__(self, model_dir, max_length=320):
-        super(STEP1TextEncoder, self).__init__()
+        super()
         self.max_length = max_length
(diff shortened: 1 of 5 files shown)


fastvideo/models/stepvideo/text_encoder/stepllm.py
```diff
@@ -276,7 +276,7 @@ def __init__(self, model_dir, max_length=320):
     @torch.no_grad
     def forward(self, prompts, with_mask=True, max_length=None):
         self.device = next(self.text_encoder.parameters()).device
-        with torch.no_grad(), torch.cuda.amp.autocast(dtype=torch.bfloat16):
+        with torch.no_grad(), torch.amp.autocast(device_type='cuda', dtype=torch.bfloat16):
```
Edenzzzz: Please remove the `no_grad` as the func has the decorator. Also use `@torch.autocast`

b8zhong: Alright 👍 done

SolitaryThinker (approved):

PR #396, opened 2025-05-11
[Misc] Replace instances of `time.time()` with `time.perf_counter()`

applesaucethebun:
For this purpose, `time.time()` should not be used to measure inference code speed, as well as benchmarks, instead we should use `time.perf_counter()`. See: https://peps.python.org/pep-0418/#rationale 👍


SolitaryThinker (approved):
Thank you for the PR! Looks good to me

PR #414, opened 2025-05-18
[Lint] Update min python to 3.10 and deprecated type hints

SolitaryThinker:
Removes deprecated Python 3.8 and python 3.9 syntax.


SolitaryThinker:
closing for now

PR #419, opened 2025-05-22
[feat] add flux pipeline to v1

Eigensystem:
recreate from #380


fastvideo/v1/tests/transformers/test_flux.py
```diff
@@ -0,0 +1,151 @@
...
+        p.to(torch.float64).sum().item() for p in model1.parameters())
+    # Also calculate mean for more stable comparison
+    weight_mean_model1 = weight_sum_model1 / total_params
+    logger.info("Model 1 weight sum: %s", weight_sum_model1)
+    logger.info("Model 1 weight mean: %s", weight_mean_model1)
+
+    # Calculate weight sum for model2 (converting to float64 to avoid overflow)
+    total_params_model2 = sum(p.numel() for p in model2.parameters())
+    weight_sum_model2 = sum(
+        p.to(torch.float64).sum().item() for p in model2.parameters())
```
Edenzzzz, line 61: I think by convention we should use `torch.testing.assert_close` for all the checks, and do it param-wise instead of summing up. Let me PR later to update tests here? https://github.com/hao-ai-lab/FastVideo/tree/main/fastvideo/v1/tests.
cc @SolitaryThinker

Eigensystem: Your are right. Of course `torch.testing.assert_close` is a better way. But for weight it might will some effort to use it directly for comparison, since we rename and split the origin weights, so it's not one-to-one. For convenience we don't use it directly.

Eigensystem: So I think it would be a bit complicated to do a param-wise weight comparison, and the original `sum` approach is sufficiently approximate to compare weight differences, so maybe we should focus on higher priorities now?

Edenzzzz: I think we should still use `assert_close`. Summing up easily underflows and makes it harder to find out which param goes wrong. I didn't get why it's not one to one, we can always do a suffix match?

Eigensystem: for original hf model, in single_block.37:
Cursor 2025-05-23 01 32 35

in our model implement:
Cursor 2025-05-23 01 32 42
We have combined some of the named parameters from the original hf model into one.

Eigensystem: Of course you can do that. I'm just saying it wasn't done before for convenience

Edenzzzz: This is a separate issue, but I wonder why do we have to change those names in the first place? It just makes it harder to add new models.
For now we can keep the sum, but I'd suggest using `torch.testing_assert` in other places.

fastvideo/v1/configs/models/dits/flux.py
```diff
@@ -0,0 +1,130 @@
...
+    pooled_projection_dim: int = 768
+    guidance_embeds: bool = False
+    axes_dims_rope: Tuple[int, ...] = (16, 56, 56)
+    rope_theta: int = 10000
+    dtype: Optional[torch.dtype] = torch.bfloat16
+
+    def __post_init__(self):
+        self.out_channels = self.out_channels or self.in_channels
+        self.hidden_size = self.num_attention_heads * self.attention_head_dim
+        self.num_channels_latents = self.in_channels // 4
```
Edenzzzz, line 123: rename to `num_latent_channels` throughout the repo?

fastvideo/v1/tests/transformers/test_flux.py
SolitaryThinker: the reason is because we are using our own layers for each model implementation, meaning we need to map the weights name to our internal layer's name. We can definitely clean up the tests in later PR

fastvideo/v1/worker/gpu_worker.py
```diff
@@ -138,7 +138,7 @@ def event_loop(self) -> None:
                     fastvideo_args = recv_rpc['kwargs']['fastvideo_args']
                     output_batch = self.execute_forward(forward_batch,
                                                         fastvideo_args)
-                    self.pipe.send({"output_batch": output_batch.output.cpu()})
+                    self.pipe.send({"output_batch": output_batch})
```
SolitaryThinker, line 141: Is there a reason this change is needed?

SolitaryThinker:
Overall this looks great, thanks for the effort. I will update the SSIM tests myself at some point and then we can merge. I will double check the quality of hunyuan before that

fastvideo/v1/tests/transformers/test_flux.py
Edenzzzz: Yeah we can skip this for now. We can use our own layers, but keep the names consistent with hf

fastvideo/v1/worker/gpu_worker.py
Eigensystem: This is because I think ForwardBatch should be used throughout the entire inference process. It is passed in by the executor and then out to the executor after the computation is done, and it is also beneficial to get something in the executor/user script other than perhaps output.

PR #423, opened 2025-05-22
Unify env report script in issue template

Edenzzzz:
Remove the old `fastvideo/utils/env_utils.py`, use the more comprehensive `collect_env.py`.


Edenzzzz:
cc @SolitaryThinker

PR #424, opened 2025-05-23
Use version.py

Edenzzzz:
Just realized we should use version.py for visibility😂, as [in SGLang ](https://github.com/sgl-project/sglang/blob/main/python/sglang/version.py)
Follows https://github.com/hao-ai-lab/FastVideo/pull/422.


SolitaryThinker (approved):

Edenzzzz:
done, there was a bug with pre-commit on my machine

PR #426, opened 2025-05-23
Use assert_close in tests & download to default cache location

Edenzzzz:


SolitaryThinker (changes requested):
Could reopen from a branch on upstream? thanks

PR #429, opened 2025-05-24
Use assert_close in tests & download to default cache location #426

Edenzzzz:
Supersedes #426


fastvideo/v1/tests/encoders/test_clip_encoder.py
```diff
@@ -15,16 +15,15 @@
...
 logger = init_logger(__name__)
 
 os.environ["MASTER_ADDR"] = "localhost"
 os.environ["MASTER_PORT"] = "29503"
 
 BASE_MODEL_PATH = "hunyuanvideo-community/HunyuanVideo"
-MODEL_PATH = maybe_download_model(BASE_MODEL_PATH,
-                                  local_dir=os.path.join(
-                                      "data", BASE_MODEL_PATH))
+MODEL_PATH = maybe_download_model(BASE_MODEL_PATH)
```
SolitaryThinker: For CI tests, we are mounting a larger disk into /workspace/ and keeping /root/ very small. This is why this PR is causing tests to fail as now `maybe_download_model` uses `/root/.cache` and runs out of space. To fix this, we can either revert the change to local_dir, or increase the disk space

SolitaryThinker:
Otherwise your assert_close changes look good. Your commend in the preprocessing PR is also correct and we shouldn't use local_dir for preprocess code

fastvideo/v1/tests/encoders/test_clip_encoder.py
Edenzzzz: Sure I will leave a comment and revert it

Edenzzzz:
Have the unrelated CIs already been failing?
image

SolitaryThinker:
yes, can ignore

PR #432, opened 2025-05-26
Add torch.compile for all small ops

Edenzzzz:
Please do NOT merge until all precision tests pass (wait for #434)

### DIFF
--- a/fastvideo/v1/layers/activation.py
+++ b/fastvideo/v1/layers/activation.py
@@ -25,6 +25,7 @@ class SiluAndMul(CustomOp):
     def __init__(self) -> None:
         super().__init__()
 
+    @torch.compile(dynamic=True)
     def forward_native(self, x: torch.Tensor) -> torch.Tensor:
         """PyTorch-native implementation equivalent to forward()."""
         d = x.shape[-1] // 2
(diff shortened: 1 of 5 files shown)


SolitaryThinker:
Do you see a speedup when compiling these? the SSIM tests just needs to be updated for it to pass if the video quality is the same.

Edenzzzz:
> Do you see a speedup when compiling these? the SSIM tests just needs to be updated for it to pass if the video quality is the same.

I believe SSIM tests can fail on their own
https://github.com/hao-ai-lab/FastVideo/actions/runs/15358666189/job/43222569090

Edenzzzz:
@SolitaryThinker @jzhang38 The speed up is quite significant (A100). I may remove a few compiles to find the one causing precision drifts
## Before compile
image
image

## After compile
image
image

jzhang38:
Just wanna double-check: does this work with model-level compilation?

https://github.com/hao-ai-lab/FastVideo/blob/8631c1b806872c344a78a460b9def494d5a7be65/fastvideo/v1/models/loader/component_loader.py#L423

Edenzzzz:
> Just wanna double-check: does this work with model-level compilation?
> 
> https://github.com/hao-ai-lab/FastVideo/blob/8631c1b806872c344a78a460b9def494d5a7be65/fastvideo/v1/models/loader/component_loader.py#L423

I think this large a compile region covering the whole transformer block may not work
1. It might cause graph breaks/recompilation 
2. Default compile (non-cuda-graph) cannot cover communication

You will need modular compilation as in XDit and in my experience, precision will drift due to triton, but the video quality will be the same

jzhang38 (approved):

Edenzzzz:
We should not merge this for now because it breaks the CI😂 I need to investigate which op is causing issues

Edenzzzz:
It's hard to test without incurring precision errors. On my A100,  `pytest -vv fastvideo/v1/tests/transformers/test_hunyuanvideo_load.py` fails on main branch

PR #435, opened 2025-05-27
misc: Add flash-attn to CI dependency

Edenzzzz:
Flash attn should be installed because [SSIM tests](https://github.com/hao-ai-lab/FastVideo/blob/12647457a784a41add9a4d93e5f13a052a8938d3/fastvideo/v1/tests/ssim/test_inference_similarity.py#L133) uses it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ test = [
     "av==14.3.0",
     "pytorch-msssim==1.0.0",
     "pytest",
+    "flash-attn==2.7.4.post1 --no-cache-dir --no-build-isolation",
 ]
 
 dev = [ "fastvideo[lint]", "fastvideo[test]", ]


SolitaryThinker:
The result is that we need manually install flash-attn after ensuring torch is available

Edenzzzz:
I think this already assumes a working pytorch env? CIs use docker before installing test deps 
https://github.com/hao-ai-lab/FastVideo/blob/657fd745e1a274980a7757f04f5cac54da93bd4b/.github/workflows/pr-test.yml#L95

SolitaryThinker:
Right, and the docker images already have flash-attn installed. The main point is that if we add flash-attn directly to pyproject, anyone using `pip install -e .[xxx]` which includes flash-attn from a clean env will fail

Edenzzzz:
I see, then it should be sufficient for CIs

PR #438, opened 2025-05-27
[Training] [1/n] Add latent datasets

SolitaryThinker:
- [x] Clean up code

cc @JerryZhou54

### DIFF
--- a/fastvideo/v1/dataset/latent_datasets.py
+++ b/fastvideo/v1/dataset/latent_datasets.py
@@ -1,0 +1,129 @@
+import json
+import os
+import random
+
+import torch
+from torch.utils.data import Dataset
+
+
+class LatentDataset(Dataset):
+
+    def __init__(
+        self,
+        json_path,
+        num_latent_t,
+        cfg_rate,
+    ) -> None:
+        # data_merge_path: video_dir, latent_dir, prompt_embed_dir, json_path
+        self.json_path = json_path
+        self.cfg_rate = cfg_rate
+        self.datase_dir_path = os.path.dirname(json_path)
+        self.video_dir = os.path.join(self.datase_dir_path, "video")
+        self.latent_dir = os.path.join(self.datase_dir_path, "latent")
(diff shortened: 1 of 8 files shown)


SolitaryThinker:
rank 0:
	1. first reads in # rows pers per parquet file, # of parquet files
	2. Generates list of indexes for each worker (dp, sp), global shuffling
	3. write this “plan” to disk

rank N:
Read plan from disk

Index into its own list of batches (file #, row #)
Read row
latent: [bs, c, t, h, w]
n = sp_size
[bs, c, t, h, w] -> [bs, c, (n s), h, w]  -> [bs, c, t // sp_size, h, w]
latent = [:, :, rank, :, :, :]

SolitaryThinker:
cc @zhisbug

fastvideo/v1/dataset/latent_datasets.py
```diff
@@ -0,0 +1,129 @@
...
+    for latent, prompt_embed, latent_attn_mask, prompt_attention_mask in dataloader:
+        print(
+            latent.shape,
+            prompt_embed.shape,
+            latent_attn_mask.shape,
+            prompt_attention_mask.shape,
+        )
+        import pdb
+
+        pdb.set_trace()
```
Edenzzzz, line 129: It's better to remove or use a separate file for testing.

fastvideo/v1/dataset/parquet_datasets.py
```diff
@@ -0,0 +1,369 @@
...
+
+    # Create DataLoader with proper settings
+    dataloader = StatefulDataLoader(
+        dataset,
+        batch_size=args.batch_size,
+        num_workers=1,  # Reduce number of workers to avoid memory issues
+        prefetch_factor=2,
+        shuffle=False,
+        pin_memory=True,
+        drop_last=True)
```
Edenzzzz, line 270: The best practice may be  `pin_memory_device=f"cuda:{torch.cuda.current_device()}`, as latents don't take much memory. The `tensor.to` is blocking on its own.

PR #442, opened 2025-05-28
[Training] [0/n] Add preprocessing pipeline

JerryZhou54:
- [x] Save to parquets periodically 
- [x] Allow resuming from the middle

### DIFF
--- a/fastvideo/data_preprocess/preprocess.py
+++ b/fastvideo/data_preprocess/preprocess.py
@@ -16,6 +16,7 @@ from fastvideo.v1.pipelines.preprocess_pipeline import PreprocessPipeline
 logger = init_logger(__name__)
 
 def main(args):
+    args.model_path = maybe_download_model(args.model_path)
     # Assume using torchrun
     local_rank = int(os.getenv("RANK", 0))
     rank = int(os.environ.get("RANK", 0))
--- a/fastvideo/v1/dataset/__init__.py
+++ b/fastvideo/v1/dataset/__init__.py
@@ -1,3 +1,5 @@
+import os
+
 from torchvision import transforms
 from torchvision.transforms import Lambda
 from transformers import AutoTokenizer
@@ -25,8 +27,8 @@ def getdataset(args, start_idx=0) -> T2V_dataset:
         *resize_topcrop,
         norm_fun,
     ])
-    # tokenizer = AutoTokenizer.from_pretrained("/storage/ongoing/new/Open-Sora-Plan/cache_dir/mt5-xxl", cache_dir=args.cache_dir)
-    tokenizer = AutoTokenizer.from_pretrained(args.text_encoder_name,
+    tokenizer_path = os.path.join(args.model_path, "tokenizer")
+    tokenizer = AutoTokenizer.from_pretrained(tokenizer_path,
                                               cache_dir=args.cache_dir)
     if args.dataset == "t2v":
         return T2V_dataset(args,


JerryZhou54:
Need to merge #438 first, because this PR requires v1/datasets

fastvideo/data_preprocess/preprocess.py
```diff
@@ -0,0 +1,122 @@
...
+from fastvideo.v1.configs.models.vaes import WanVAEConfig
+from fastvideo import PipelineConfig
+from fastvideo.v1.pipelines.preprocess_pipeline import PreprocessPipeline
+
+logger = init_logger(__name__)
+
+BASE_MODEL_PATH = "/workspace/data/Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
+MODEL_PATH = maybe_download_model(BASE_MODEL_PATH,
+                                  local_dir=os.path.join(
+                                      'data', BASE_MODEL_PATH))
```
Edenzzzz: I'm not sure why we should use `local_dir`--this can make cache sharing more complicated. I've [PRed to remove this ](https://github.com/hao-ai-lab/FastVideo/pull/429)

scripts/preprocess/preprocess_wan_data.sh
```diff
@@ -0,0 +1,25 @@
+# export WANDB_MODE="offline"
+GPU_NUM=1 # 2,4,8
+MODEL_PATH="Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
+TEXT_ENCODER_PATH="/Wan-AI/Wan2.1-T2V-1.3B-Diffusers/tokenizer"
```
SolitaryThinker: instead of hardcoding this TEXT_ENCODER_PATH, we can simply do:

```
path = maybe_download_model(args.model_path)
encoder_path = os.join(path, 'tokenizer')
```

fastvideo/data_preprocess/preprocess.py
```diff
@@ -0,0 +1,122 @@
...
+from fastvideo.v1.utils import maybe_download_model, shallow_asdict
+from fastvideo.v1.distributed import init_distributed_environment, initialize_model_parallel
+from fastvideo.v1.fastvideo_args import FastVideoArgs
+from fastvideo.v1.configs.models.vaes import WanVAEConfig
+from fastvideo import PipelineConfig
+from fastvideo.v1.pipelines.preprocess_pipeline import PreprocessPipeline
+
+logger = init_logger(__name__)
+
+BASE_MODEL_PATH = "/workspace/data/Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
```
SolitaryThinker: let's just use args.model_path here instead?

fastvideo/data_preprocess/preprocess.py
SolitaryThinker: it may have been me that added the hardcoded path? @JerryZhou54 perhaps lets just use `model_path` arg here and have the registry detect the correct pipeline config by directly using the HF string instead of a local path?

SolitaryThinker:
feel free to merge after addressing my comments

fastvideo/data_preprocess/preprocess.py
```diff
@@ -0,0 +1,118 @@
...
+    fastvideo_args.device = torch.device(f"cuda:{local_rank}")
+
+    pipeline = PreprocessPipeline(args.model_path, fastvideo_args)
+    pipeline.forward(batch=None, fastvideo_args=fastvideo_args, args=args)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    # dataset & dataloader
+    parser.add_argument("--model_path", type=str, default="data/mochi")
```
Edenzzzz, line 53: it's better to not use `data` except only for testers running on Runpod, just use hf's default cache path

PR #447, opened 2025-05-28
[Training] [5/n] Add single gpu training pipeline

SolitaryThinker:


fastvideo/v1/training/training_pipeline.py
```diff
@@ -0,0 +1,492 @@
...
+            batch_size=1,
+            num_workers=1,  # Reduce number of workers to avoid memory issues
+            prefetch_factor=2,
+            shuffle=False,
+            pin_memory=True,
+            drop_last=False)
+
+        transformer.requires_grad_(False)
+        for p in transformer.parameters():
+            p.requires_grad = False
```
Edenzzzz, line 159: Don't think we need this

PR #449, opened 2025-05-29
[Training] [7/n] gradient clipping

SolitaryThinker:


fastvideo/v1/training/training_utils.py
```diff
@@ -107,3 +110,216 @@ def normalize_dit_input(model_type, latents, args=None) -> torch.Tensor:
...
+            Maximum norm of the gradients after clipping.
+        norm_type (`float`, defaults to `2.0`):
+            Type of p-norm to use. Can be `inf` for infinity norm.
+        error_if_nonfinite (`bool`, defaults to `False`):
+            If `True`, an error is thrown if the total norm of the gradients from `parameters` is `nan`, `inf`, or `-inf`.
+        foreach (`bool`, defaults to `None`):
+            Use the faster foreach-based implementation. If `None`, use the foreach implementation for CUDA and CPU native tensors
+            and silently fall back to the slow implementation for other device types.
+        pp_mesh (`torch.distributed.device_mesh.DeviceMesh`, defaults to `None`):
+            Pipeline parallel device mesh. If not `None`, will reduce gradient norm across PP stages.
```
Edenzzzz, line 176: I'm not sure if we want to support PP, even TP is not needed in many cases. When do we need to split the model into > 8 shards?

PR #450, opened 2025-05-29
[Training] [8/n] SP Training

SolitaryThinker:


fastvideo/v1/training/training_pipeline.py
```diff
@@ -332,10 +339,25 @@ def compute_loss() -> torch.Tensor:
             absolute_errors: list[float] = []
             param_count = 0
 
+            rank = int(os.environ.get("RANK", 0))
```
Edenzzzz, line 342: better use `torch.distributed.get_rank()`

PR #451, opened 2025-05-30
[WIP] V1 LoRA inference support

Edenzzzz:
Upload the rest of it tmr morning

### DIFF
--- a/examples/inference/basic/default_args.py
+++ b/examples/inference/basic/default_args.py
@@ -1,5 +1,5 @@
 from fastvideo import VideoGenerator
-
+from fastvideo.v1.configs.pipelines.base import PipelineConfig
 
 def main():
 
--- a/fastvideo/v1/fastvideo_args.py
+++ b/fastvideo/v1/fastvideo_args.py
@@ -11,6 +11,7 @@ from typing import Any, Callable, List, Optional, Tuple
 from fastvideo.v1.configs.models import DiTConfig, EncoderConfig, VAEConfig
 from fastvideo.v1.logger import init_logger
 from fastvideo.v1.utils import FlexibleArgumentParser, StoreBoolean
+from peft import LoraConfig
 
 logger = init_logger(__name__)
 
@@ -80,6 +81,10 @@ class FastVideoArgs:
         default_factory=lambda: (preprocess_text, ))
     postprocess_text_funcs: Tuple[Callable[[Any], Any], ...] = field(
         default_factory=lambda: (postprocess_text, ))
+    
+    # LoRA parameters
+    lora_config: Optional[LoraConfig] = None
+    lora_path: Optional[str] = None
 
     # STA (Spatial-Temporal Attention) parameters
     mask_strategy_file_path: Optional[str] = None
--- a/fastvideo/v1/layers/lora/linear.py
+++ b/fastvideo/v1/layers/lora/linear.py
@@ -1,0 +1,336 @@
+# Code adapted from SGLang https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/lora/layers.py
+
+from typing import List, Tuple
+
+import torch
+from torch import nn
+
+from fastvideo.v1.distributed import (
+    get_tensor_model_parallel_rank,
+    split_tensor_along_last_dim,
+    tensor_model_parallel_all_gather,
+    tensor_model_parallel_all_reduce,
+)
+from fastvideo.v1.layers.linear import (
+    ColumnParallelLinear,
+    MergedColumnParallelLinear,
+    QKVParallelLinear,
+    RowParallelLinear,
+)
+from fastvideo.v1.layers.vocab_parallel_embedding import VocabParallelEmbedding
+
+
+class BaseLayerWithLoRA(nn.Module):
+    def __init__(
+        self,
+        base_layer: nn.Module,
+    ):
+        super().__init__()
+        self.base_layer: nn.Module = base_layer
+        self.set_lora: bool = False
+        self.lora_A: torch.Tensor = None
+        self.lora_B: torch.Tensor = None
+
+    def forward(self, x: torch.Tensor):
+        return self.base_layer.forward(x)
+
+    def set_lora_info(self, *args):
+        pass
+
+    def slice_lora_a_weights(self, A: torch.Tensor, tp_rank: int):
+        pass
+
+    def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
+        pass
+
+    def merge_lora_weights(self):
+        pass
+
+    def unmerge_lora_weights(self):
+        pass
+
+class VocabParallelEmbeddingWithLoRA(BaseLayerWithLoRA):
+    """
+    Vocab parallel embedding layer with support for LoRA (Low-Rank Adaptation).
+
+    Note: The current version does not yet implement the LoRA functionality.
+    This class behaves exactly the same as the base VocabParallelEmbedding.
+    Future versions will integrate LoRA functionality to support efficient parameter fine-tuning.
+    """
+
+    def __init__(
+        self,
+        base_layer: VocabParallelEmbedding,
+    ) -> None:
+        super().__init__(base_layer)
+        self.weight = base_layer.weight
+
+
+class ColumnParallelLinearWithLoRA(BaseLayerWithLoRA):
+    def __init__(
+        self,
+        base_layer: ColumnParallelLinear,
+    ) -> None:
+        super().__init__(base_layer)
+
+
+
+    def forward(self, input_: torch.Tensor):
+        # duplicate the logic in ColumnParallelLinear
+        bias = self.base_layer.bias if not self.base_layer.skip_bias_add else None
+        output_parallel = self.base_layer.quant_method.apply(
+            self.base_layer, input_, bias
+        )
+        if self.base_layer.gather_output:
+            output = tensor_model_parallel_all_gather(output_parallel)
+        else:
+            output = output_parallel
+        output_bias = self.base_layer.bias if self.base_layer.skip_bias_add else None
+        return output, output_bias
+
+    def slice_lora_a_weights(self, A: torch.Tensor, tp_rank: int):
+        return A
+
+    def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
+        shard_size = self.base_layer.output_partition_sizes[0]
+        start_idx = tp_rank * shard_size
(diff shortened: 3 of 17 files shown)


fastvideo/v1/models/loader/fsdp_load.py
```diff
@@ -100,42 +65,49 @@ def load_fsdp_model(
...
+        device_mesh = init_device_mesh(
+            "cuda",
+            mesh_shape=(get_sequence_model_parallel_world_size(), ),
+            mesh_dim_names=("dp", ),
+        )
+        shard_model(model,
+                    cpu_offload=cpu_offload,
+                    reshard_after_forward=True,
+                    mp_policy=mp_policy,
+                    dp_mesh=device_mesh["dp"])
```
Edenzzzz: There's a bug that the model is on CPU when not using fsdp. Should we set offload to true only for training?
https://github.com/hao-ai-lab/FastVideo/blob/b2ebaaf8656e87ee722090a8ed17f5dbd0abcaa3/fastvideo/v1/configs/pipelines/wan.py#L41

SolitaryThinker: This is intentional, as when `use_cpu_offload` is true for inference, each stage will move the model onto the device only when executing to save memory and then move it back to CPU. This means each model is loaded on the CPU at first.

Edenzzzz: Without fsdp it doesn't seem to move model back to GPU, leading to an error

fastvideo/v1/fastvideo_args.py
```diff
@@ -100,6 +107,8 @@ class FastVideoArgs:
     device_str: Optional[str] = None
     device = None
 
+    is_training: bool = False
```
SolitaryThinker: we already have training_mode, not sure why we need this as well?

fastvideo/v1/layers/lora/linear.py
```diff
@@ -0,0 +1,261 @@
...
+        self,
+        base_layer: nn.Module,
+    ):
+        super().__init__()
+        self.base_layer: nn.Module = base_layer
+        self.lora_A: torch.Tensor = None
+        self.lora_B: torch.Tensor = None
+        self.merged: bool = False
+
+        # when adapter weights don't container this layer
```
SolitaryThinker: nit: container-> contain

examples/inference/basic/basic.py
```diff
@@ -23,7 +23,7 @@ def main():
         "wide with interest. The playful yet serene atmosphere is complemented by soft "
         "natural light filtering through the petals. Mid-shot, warm and cheerful tones."
     )
-    video = generator.generate_video(prompt)
+    video = generator.generate_video(prompt, output_path=OUTPUT_PATH, save_video=True)
```
SolitaryThinker, line 28: Thanks for adding this

examples/inference/basic/default_args.py
SolitaryThinker, line 1: This file was actually accidentally included, but if this fixes it completely then lets keep it

fastvideo/v1/fastvideo_args.py
```diff
@@ -81,6 +81,13 @@ class FastVideoArgs:
     postprocess_text_funcs: Tuple[Callable[[Any], Any], ...] = field(
         default_factory=lambda: (postprocess_text, ))
 
+    # LoRA parameters
+    lora_path: Optional[str] = None
+    lora_nick_name: Optional[
```
SolitaryThinker: nit: lora_nick_name -> lora_nickname

fastvideo/v1/models/loader/component_loader.py
```diff
@@ -396,7 +396,7 @@ def load(self, model_path: str, architecture: str,
         # Load the model using FSDP loader
         logger.info("Loading model from %s, default_dtype: %s", cls_name,
                     default_dtype)
-        model = load_fsdp_model(
+        model = maybe_load_fsdp_model(
```
SolitaryThinker, line 406: we are using fsdp for inference as well

fastvideo/v1/models/loader/fsdp_load.py
SolitaryThinker, line 1: We are always using fsdp and regardless, changes to this should be in a separate PR if it is warranted at all?

fastvideo/v1/models/loader/utils.py
```diff
@@ -16,3 +18,30 @@ def set_default_torch_dtype(dtype: torch.dtype):
     torch.set_default_dtype(dtype)
     yield
     torch.set_default_dtype(old_dtype)
+
+
+def get_param_names_mapping(
```
SolitaryThinker, line 23: thanks for moving this

fastvideo/v1/pipelines/composed_pipeline_base.py
```diff
@@ -96,7 +97,7 @@ def __init__(self,
 
         if not fastvideo_args.training_mode:
             logger.info("Creating pipeline stages...")
-            self.create_pipeline_stages(fastvideo_args)
+            self.create_inference_stages(fastvideo_args)
```
SolitaryThinker: Could we leave the name as `create_pipeline_stages` for now? I'm not sure how much of the current training code can actually be refactored/abstracted into stages yet. And I feel if we do, it will be better to just unify it under `create_pipeline_stages`. Happy to hear your thoughts if you have taken a look at the current training_pipeline.py

SolitaryThinker (commented):
Looks great, will do another pass tomorrow. Thanks for the effort!

fastvideo/v1/models/loader/component_loader.py
Edenzzzz: Not sure why fsdp should be used when TP is available--it will make inference super slow? It will also make merging lora weights harder

fastvideo/v1/pipelines/composed_pipeline_base.py
Edenzzzz: sure

fastvideo/v1/models/loader/component_loader.py
SolitaryThinker: For inference, both FSDP and SP is used only for DiT. TP is only used for encoders. 

We are using fsdp2 and I feel the DTensor API shouldn't cause too much issues with merging lora? Is it not possible to do `full_tensor` -> merge lora -> `distributed_tensor`? Let me know if this isn't the case.

SolitaryThinker: TP and FSDP is never used on the same model

Edenzzzz: It just doesn't quite make sense to use FSDP for inference--is the memory really that constrained? An all-gather per forward per layer is very expensive.

Edenzzzz: The DiT shouldn't take that much space, I think using SP is mostly enough. If you use SP and FSDP they will contend for network bandwidth

Edenzzzz: image

SolitaryThinker: For GPUs without NVLink not using FSDP makes sense, I've seen pretty high overheads when using multi-GPU without NVLink. 

I've been considering support DP directly for inference, since from a cost perspective (even with nvlink) , DP beats FSDP+ SP assuming everything fits.

SolitaryThinker: In any case, I feel this change should be in a separate PR regardless and we can add this as a TODO item? This PR is already quite large and if you only support single GPU lora initially thats okay. You can open another PR to support FSDP+lora?

Edenzzzz: Yeah let's not use FSDP for this PR. FSDP is generally not suitable for inference. DP can be used for online serving like in SGLang

Edenzzzz:
DP vs SP (from Snowflow Arctic)
image

fastvideo/v1/pipelines/lora_pipeline.py
```diff
@@ -0,0 +1,118 @@
...
+            return
+
+        # Merge the new adapter
+        adapted_count = 0
+        for name, layer in self.lora_layers.items():
+            name = ".".join(name.split(".")[1:]) # remove the transformer prefix
+            lora_A_name = name + ".lora_A"
+            lora_B_name = name + ".lora_B"
+            if lora_A_name in self.lora_adapters[adapter_nickname]\
+                and lora_B_name in self.lora_adapters[adapter_nickname]:
```
Edenzzzz: @SolitaryThinker The weight names are confusing here--the hf model weights uses the same convention as ours (attn1, attn2), but [the lora weights from here](https://huggingface.co/benjamin-paine/steamboat-willie-1.3b) use a [different set of names](https://github.com/Wan-Video/Wan2.1/blob/ec902046f6f8b264cff94bd8b2cb1d44c10090e3/wan/modules/model.py#L260) from the original Wan repo...
We must then define a new `_lora_param_names_mapping`, but I don't know if other LoRA weights will follow this format

Edenzzzz: image
image

Edenzzzz:
cc @SolitaryThinker I think this is ready to merge, you can do a final review.
Will make more PRs for TODO items above
image
image

SolitaryThinker:
okay seems like the lora code is causing some failures in ssim test

Edenzzzz:
Could be `non_blocking`. Will fix before I merge

fastvideo/v1/layers/lora/linear.py
```diff
@@ -0,0 +1,270 @@
...
+        self.unmerge_count = 0
+        # indicates adapter weights don't contain this layer
+        # (which shouldn't normally happen, but we want to separate it from the case of erroneous merging)
+        self.disable_lora: bool = False
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.base_layer.forward(x)
+
+    def slice_lora_a_weights(self, A: torch.Tensor,
+                             tp_rank: int) -> torch.Tensor:
```
jzhang38: tp_rank is not used?

jzhang38: OK I got it it is overloaded later in subclass.

jzhang38:
I took a pass and lgtm

Edenzzzz:
The SSIM tests were failing due to OOM. This PR now supports FSDP and cpu offload for LoRA, which are enabled by default. The latency seems negligible due to prefetch.

PR #452, opened 2025-05-30
Fix distributed datasets issue

JerryZhou54:


fastvideo/v1/dataset/parquet_datasets.py
```diff
@@ -54,6 +55,7 @@ def __init__(self,
             # This will be useful when resume training
             if os.path.exists(self.plan_output_dir):
                 print(f"Using existing plan from {self.plan_output_dir}")
+                dist.barrier()
```
Edenzzzz, line 58: I think this will cause rank 0 to hang infinitely

SolitaryThinker:
should be included in https://github.com/hao-ai-lab/FastVideo/pull/461. Please reopen if I'm wrong

PR #459, opened 2025-06-01
[Training] Support Multi-Node training with FSDP + SP

SolitaryThinker:
Tested overfitting on t2v for the following setups
(world_size, sp_size, dp_size, dp_shard_degree)
(8, 8, 1, 8)
(8, 4, 2, 4)
(16, 8, 2, 8)

### DIFF
--- a/fastvideo/v1/dataset/parquet_datasets.py
+++ b/fastvideo/v1/dataset/parquet_datasets.py
@@ -82,16 +82,12 @@ class ParquetVideoTextDataset(Dataset):
             group_ranks_list: List[Any] = list(
                 set(tuple(r) for r in group_ranks))
             num_sp_groups = len(group_ranks_list)
-            print(f"num_sp_groups: {num_sp_groups}")
-            print(f"group_ranks_list: {group_ranks_list}")
             plan = defaultdict(list)
             for idx, metadata in enumerate(metadatas):
                 sp_group_idx = idx % num_sp_groups
                 for global_rank in group_ranks_list[sp_group_idx]:
                     plan[global_rank].append(metadata)
 
-            print(f"plan: {plan}")
-            # import pdb; pdb.set_trace()
             with open(self.plan_output_dir, "w") as f:
                 json.dump(plan, f)
         dist.barrier()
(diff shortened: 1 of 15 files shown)


fastvideo/v1/dataset/parquet_datasets.py
```diff
@@ -121,9 +127,9 @@ def __getitem__(self, idx):
         cumulative = 0
         for i in range(parquet_file.num_row_groups):
             num_rows = parquet_file.metadata.row_group(i).num_rows
-            if cumulative + num_rows > idx:
+            if cumulative + num_rows > row_idx:
```
jd-nuva, line 130: Not directly related to this PR but torchdataset has `__getitems__` that naturally works with a list of indices and much more performant for training setting too beyond bs=1, i just tried the same code today and __getitems__ is 5~10x faster on my small dataset.

For multi-resolution this might require picking the resolution buckets in higher level component like dataset / sampler tho, so `__getitems__` itself is only getting list of indices of the same resolution that respects randomness / weighted sampling.

fastvideo/v1/distributed/parallel_state.py
```diff
@@ -794,9 +794,18 @@ def get_sp_group() -> GroupCoordinator:
     return _SP
 
 
+_DP: Optional[GroupCoordinator] = None
+
+
+def get_dp_group() -> GroupCoordinator:
+    assert _DP is not None, ("data parallel group is not initialized")
+    return _DP
+
```
jd-nuva, line 803: Quick question -- do we support any (dp, tp, sp) combination for both training and validation in given device mesh with this PR? My previous fork had to keep coming back here or in component_loader.py that manually changes device mesh shape based on my setup.

SolitaryThinker: TP is currently not used in training or validation. And we actually set the mesh shape for DP in component_loader in this PR

fastvideo/v1/models/loader/fsdp_load.py
```diff
@@ -109,19 +110,25 @@ def load_fsdp_model(
...
+        # (Replicate(), Shard(dim=0))
+        mesh_shape=(data_parallel_size, data_parallel_shards),
+        mesh_dim_names=("dp", "sp"),
     )
     shard_model(model,
                 cpu_offload=cpu_offload,
                 reshard_after_forward=True,
                 mp_policy=mp_policy,
-                dp_mesh=device_mesh["dp"])
+                mesh=device_mesh)
```
SolitaryThinker, line 131: See here @jd-nuva

fastvideo/v1/dataset/parquet_datasets.py
SolitaryThinker: got it thank you for this pointer. Currently our dataset code is pretty rough and we plan on improving this and we'll definitely take a look at this

fastvideo/v1/distributed/parallel_state.py
jd-nuva: ah interesting, so it's only num_gpus = dp * sp ?

But i thought the fact that it uses FSDP loader already added some TP degrees iiuc

SolitaryThinker:
going to fix or disable transformer test later, as currently the transformer test no longer go through the same codepath as inference

PR #466, opened 2025-06-03
[do not merge] adding gradient checkpointing

Zihang-He:


fastvideo/v1/training/training_pipeline.py
```diff
@@ -52,6 +53,16 @@ def initialize_training_pipeline(self, training_args: TrainingArgs):
         self.transformer = self.get_module("transformer")
         assert self.transformer is not None
 
+        # if training_args.gradient_checkpointing:
+        if True:
```
Edenzzzz, line 57: `if training_args.gradient_checkpointing`

Edenzzzz (changes requested):

SolitaryThinker:
For some context, this PR is currently not working with mixed preicision and is for reference @jzhang38

philippe-eecs:
FWIW when I wrap wanvideo block with torch.utils.checkpoint.checkpoint it gives me this context error:

```
[rank0]:     assert _forward_context is not None, ( 
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: AssertionError: Forward context is not set. Please use `set_forward_context` to set the forward context.
```

I solved the error with a weird wrap. Doesn't seem to slow down the model at all, but now the model is able to run without OOMs with larger batch sizes. (I tried with 4xH100s with SP 4, TP 4, FSDP over 4 GPUs)

```
            # if teacache is enabled, we need to cache the original hidden states
            if enable_teacache:
                original_hidden_states = hidden_states.clone()

            if self.gradient_checkpointing and torch.is_grad_enabled():
                def _checkpoint_block(block, *tensors):
                    ctx = get_forward_context()
                    if ctx is None:
                        return block(*tensors)  # fallback
                    context_kwargs = ctx.as_kwargs()
                    def fn(*args):
                        with set_forward_context(**context_kwargs):
                            return block(*args)
                    return torch.utils.checkpoint.checkpoint(fn, *tensors, use_reentrant=False)

                for block in self.blocks:
                    hidden_states = _checkpoint_block(block,
                        hidden_states,
                        encoder_hidden_states,
                        timestep_proj,
                        freqs_cis
                    )
            else:
                for block in self.blocks:
                    hidden_states = block(hidden_states, encoder_hidden_states,
                                          timestep_proj, freqs_cis)
 ```

SolitaryThinker:
cc @Zihang-He could you take a look?

SolitaryThinker:
added in #584

PR #470, opened 2025-06-03
[Feature] Load weights from distributed

Edenzzzz:
The read speed of nvme SSDs is only a few GB/s, far slower than the 800GB/s on NVLink. This PR supports loading weights on one rank and broadcasting them to the node. 
After this PR, text encoders will be offloaded (w/ layer-wise prefetch) to cpu by default using FSDP, because encoder tests will simply OOM on A40 without this.
## TODO
- [x] Attach performance numbers 
- [ ] Use scatter instead of broadcast (in a future PR)
- [ ] Support multithreaded loading (https://github.com/sgl-project/sglang/pull/7277)


jzhang38:
How is the performance so far?

Edenzzzz:
Need to offload encoder for CIs to pass first

fastvideo/v1/tests/encoders/test_t5_encoder.py
```diff
@@ -43,9 +43,8 @@ def test_t5_encoder():
     loader = TextEncoderLoader()
     model2 = loader.load(TEXT_ENCODER_PATH, "", args)
 
-    # Convert to float16 and move to device
-    # model2 = model2.to(precision)
-    model2 = model2.to(device)
```
Edenzzzz, line 50: @SolitaryThinker Seems like the t5 test still gets OOM (and even consumes more memory) after using cpu offload. Any ideas? Seems that FSDP doesn't move param back to CPU in post-forward of each layer

Edenzzzz: Actually, this is because fsdp [disentangles the shared parameters and creates duplicates](https://github.com/pytorch/pytorch/blob/f59c76b5494f83abbb2ca169e13201439396c1aa/torch/distributed/fsdp/_fully_shard/_fsdp_param.py#L371)

Edenzzzz:
The `RuntimeError: CUDA error: invalid argument` in encoder test is probably a A40-specific bug. It usually happens when multiple streams run concurrently, or when device run out of memory. It passes on A100.
Since we still want to support text encoder offload, we'll have to disable it and manually offload for the CI, but enable by default otherwise.

Edenzzzz:
Results on H100 and 14B model (`Wan-AI/Wan2.1-T2V-14B-Diffusers`)
| Method                | 2 GPUs | 4 GPUs |
|-----------------------|--------|--------|
| From disk             | 16 s   | 21 s   |
| Broadcast (sync)      | 14 s   | 15 s   |
| Broadcast (async)     | 11 s   | 14 s   |

Note that when using RAID 1 (such as on our B200), reading from disk can be faster when #GPUs <= 4
cc @jzhang38

Edenzzzz (commented):
Let's merge this to avoid more conflicts from `get_local_torch_device`

BrianChen1129:
why the transformer test failed?

Edenzzzz:
https://github.com/hao-ai-lab/FastVideo/pull/557#issuecomment-3013949442 it failed a while ago

BrianChen1129:
> [#557 (comment)](https://github.com/hao-ai-lab/FastVideo/pull/557#issuecomment-3013949442) it failed a while ago

okay, then just merge

PR #473, opened 2025-06-04
[Preprocess] I2V dataset

BrianChen1129:
Upload test dataset [here](https://huggingface.co/datasets/FastVideo/mini-i2v-dataset)
1. crush-smol_raw/ contains raw crush-smol data and the generated JSON file using fastvideo/vi/dataset/json_prepare.py
2. crush-smol_preprocessed contains/ preproccessed parquet file for i2v training exp


jzhang38:
Do we really need to add a 500-line file for I2V? I think I2V preprocess should resuse code from T2V preprocess as much as possible ?

SolitaryThinker:
we could conditionally modify the existing t2v preprocess pipeline. Or better yet, split it into stages and add the i2v clip processing in a new stage

SolitaryThinker:
THis will definitely take some time though and perhaps we refactor later?

BrianChen1129:
The key issue is how to load the clip module if we only want to add new code to the existing t2v preprocess pipeline. In the current i2v pipeline, I set MODEL_PATH="Wan-AI/Wan2.1-I2V-14B-480P-Diffusers" to load all modules, which I think is the simplest way. Maybe we can optimize this later.

fastvideo/v1/pipelines/preprocess_pipeline_i2v.py
zhisbug, line 1: I briefly compared this file to `preprocess_pipeline_t2v.py` and it seems many code are overlapped?

BrianChen1129: Yeah, maybe @JerryZhou54  could refactor it later?

zhisbug: maybe you and @JerryZhou54 just improve the code now and then merge

BrianChen1129: Ok, I'll do it now.

BrianChen1129:
Updated, added preprocess_pipeline_base.py to reuse the base code.

SolitaryThinker (approved):

PR #478, opened 2025-06-05
[Feature] Adding VSA inference

BrianChen1129:


.gitignore
```diff
@@ -59,3 +59,6 @@ docs/source/inference/examples/
 
 # Static images
 !docs/source/_static/images/**/*.png
+
+# ThunderKittens submodule
+csrc_vsa/sliding_tile_attention/tk/
```
jzhang38: Remove this line?

csrc/sliding_tile_attention/st_attn.cpp
```diff
@@ -7,18 +7,44 @@
...
+    torch::Tensor q, torch::Tensor k, torch::Tensor v, torch::Tensor o, torch::Tensor l_vec, torch::Tensor og, torch::Tensor k2q_block_sparse_index, torch::Tensor k2q_block_sparse_num
+);
+#endif
+
+#ifdef TK_COMPILE_ST_ATTN
 
-#ifdef TK_COMPILE_ATTN
 extern torch::Tensor sta_forward(
-    torch::Tensor q, torch::Tensor k, torch::Tensor v, torch::Tensor o, int kernel_t_size, int kernel_w_size, int kernel_h_size, int text_length, bool process_text, bool has_text, int kernel_aspect_ratio_flag
+    torch::Tensor q, torch::Tensor k, torch::Tensor v, torch::Tensor o, int kernel_t_size, int kernel_w_size, int kernel_h_size, int text_length, bool process_text, bool has_text
```
jzhang38: why removing kernel_aspect_ratio_flag?

csrc/sliding_tile_attention/st_attn/selective_tile_attention.py
```diff
@@ -0,0 +1,47 @@
...
+    B, H, Q, K = cdf.shape
+    cdfthreshd_ts = cdfthreshd.view(1, H, 1, 1)
+    cdfthreshd_ts = cdfthreshd_ts.expand(B, -1, Q, 1).contiguous()
+    num_to_select = torch.searchsorted(cdf, cdfthreshd_ts, right=True).squeeze(-1)
+    final_map = torch.zeros_like(pooled_score, dtype=torch.bool)
+    final_map[~k_block_sim_mask] = 1
+    final_map[~q_block_sim_mask] = 1
+    return final_map
+    
+def selective_tile_attention(q, k, v, simthreshd1=0.1, cdfthreshd=0.9, return_sparsity=False):
```
jzhang38: This function is never used?

fastvideo/v1/attention/backends/sliding_tile_attn.py
```diff
@@ -228,6 +228,7 @@ def forward(
             raise ValueError("forward_batch cannot be None")
         # pattern:'.double_blocks.0.attn.impl' or '.single_blocks.0.attn.impl'
         layer_idx = int(self.prefix.split('.')[-3])
+
```
jzhang38: extra empty line?

fastvideo/v1/attention/backends/video_sparse_attn.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class VideoSparseAttentionBackend(AttentionBackend):
+
+    accept_output_buffer: bool = True
+
+    @staticmethod
+    def get_supported_head_sizes() -> List[int]:
+        # TODO(will-refactor): check this
+        return [32, 64, 96, 128, 160, 192, 224, 256]
```
jzhang38: Right now we only support 64 and 128?

fastvideo/v1/attention/backends/video_sparse_attn.py
```diff
@@ -0,0 +1,187 @@
...
+                         t=self.img_latent_shape_int[0] // self.sp_size,
+                         h=self.img_latent_shape_int[1],
+                         w=self.img_latent_shape_int[2])
+
+    def preprocess_qkv(
+        self,
+        qkv: torch.Tensor,
+        attn_metadata: VideoSparseAttentionMetadata,
+    ) -> torch.Tensor:
+        self.img_latent_shape_int = attn_metadata.img_latent_shape
```
jzhang38: consider renaming img_latent_shape to somehting like dit_seq_shape_3D? 
img_latent_shape is confusing.

fastvideo/v1/attention/backends/video_sparse_attn.py
```diff
@@ -0,0 +1,187 @@
...
+        causal: bool,
+        softmax_scale: float,
+        num_kv_heads: Optional[int] = None,
+        prefix: str = "",
+        **extra_impl_args,
+    ) -> None:
+        self.prefix = prefix
+        sp_group = get_sp_group()
+        self.sp_size = sp_group.world_size
+        self.STA_base_tile_size = [4, 4, 4]
```
jzhang38: VSA base tile size?

fastvideo/v1/attention/backends/video_sparse_attn_patterns/mask.py
```diff
@@ -0,0 +1,201 @@
+import torch
```
jzhang38: This file should be put under csrc or video_sparse_attn.py?

fastvideo/v1/attention/backends/video_sparse_attn_patterns/sparse_attns.py
```diff
@@ -0,0 +1,346 @@
+try:
```
jzhang38: Please clean up this file.

fastvideo/v1/attention/layer.py
```diff
@@ -99,7 +100,7 @@ def forward(
         qkv = sequence_model_parallel_all_to_all_4D(qkv,
                                                     scatter_dim=2,
                                                     gather_dim=1)
-        # Apply backend-specific preprocess_qkv
```
jzhang38: Why removing this?

fastvideo/v1/attention/layer.py
```diff
@@ -135,6 +135,70 @@ def forward(
...
+            replicated_q (Optional[torch.Tensor]): Replicated query tensor, typically for text tokens
+            replicated_k (Optional[torch.Tensor]): Replicated key tensor
+            replicated_v (Optional[torch.Tensor]): Replicated value tensor
+            
+        Returns:
+            Tuple[torch.Tensor, Optional[torch.Tensor]]: A tuple containing:
+                - o (torch.Tensor): Output tensor after attention for the main sequence
+                - replicated_o (Optional[torch.Tensor]): Output tensor for replicated tokens, if provided
+        """
+        # Check input shapes
```
jzhang38, line 170: maybe add assertion of replicated_q to be none here as we only supports Wan self attention for now (no text tokens)

fastvideo/v1/forward_context.py
```diff
@@ -70,6 +71,13 @@ def set_forward_context(current_timestep,
     _forward_context = ForwardContext(current_timestep=current_timestep,
                                       attn_metadata=attn_metadata,
                                       forward_batch=forward_batch)
+    if attn_metadata is None:
```
jzhang38: Why do we need this, you already put this under desnoising stage around line 200.

fastvideo/v1/models/dits/wanvideo.py
```diff
@@ -389,16 +540,19 @@ def __init__(self, config: WanVideoConfig, hf_config: dict[str,
         )
 
         # 3. Transformer blocks
+        import os
+        attn_type = os.environ.get("FASTVIDEO_ATTENTION_BACKEND")
```
jzhang38: I think we should use a flag under WanVideoArchConfig to control this.

fastvideo/v1/models/loader/fsdp_load.py
```diff
@@ -256,5 +262,30 @@ def load_model_from_full_model_state_dict(
             if cpu_offload:
                 sharded_tensor = sharded_tensor.cpu()
         sharded_sd[target_param_name] = nn.Parameter(sharded_tensor)
+
+    unused_keys = set(meta_sd.keys()) - used_keys
+    if unused_keys:
+        logger.warning("New keynames in meta_sd: %s", unused_keys)
+    for new_param_name in unused_keys:
+        meta_sharded_param = meta_sd.get(new_param_name)
+        if not hasattr(meta_sharded_param, "device_mesh"):
```
jzhang38, line 278: have an if statement, only initialize param to zero if the param has str "gate_compress" in its name. Else throw an error.

.gitignore
```diff
@@ -59,3 +59,4 @@ docs/source/inference/examples/
 
 # Static images
 !docs/source/_static/images/**/*.png
+
```
jzhang38: extra space.

csrc/sliding_tile_attention/bench/test_block_sparse.py
```diff
@@ -0,0 +1,266 @@
+import torch
```
jzhang38: put this file under test/?

csrc/sliding_tile_attention/config.py
```diff
@@ -1,15 +1,25 @@
...
+        'source_files': {
+            'h100': 'st_attn/full_attn.cu'
+        }
     }
 }
 
 ### WHICH KERNELS DO WE WANT TO BUILD?
 # (oftentimes during development work you don't need to redefine them all.)
-kernels = ['attn']
+kernels = ['block_sparse']
```
jzhang38: Let's not change this in main because it will break  sliding tile attention in main.

csrc/sliding_tile_attention/st_attn/__init__.py
```diff
@@ -22,14 +37,14 @@ def sliding_tile_attention(q_all, k_all, v_all, window_size, text_length, has_te
...
     else:
-        if img_latent_shape == '36x48x48': # Stepvideo 204x768x68
+        if dit_seq_shape == '36x48x48': # Stepvideo 204x768x68
             assert q_all.shape[2] == 82944
-        elif img_latent_shape == '18x48x80': # Wan 69x768x1280
+        elif dit_seq_shape == '18x48x80': # Wan 69x768x1280
             assert q_all.shape[2] == 69120
         else:
-            raise ValueError(f"Unsupported {img_latent_shape}, current shape is {q_all.shape}, only support '36x48x48' for Stepvideo and '18x48x80' for Wan")
+            raise ValueError(f"Unsupported {dit_seq_shape}, current shape is {q_all.shape}, only support '36x48x48' for Stepvideo and '18x48x80' for Wan")
```
jzhang38, line 33: add hunyuan in the warning message as well.

csrc/sliding_tile_attention/st_attn/__init__.py
```diff
@@ -44,3 +59,240 @@ def sliding_tile_attention(q_all, k_all, v_all, window_size, text_length, has_te
...
+def block_sparse_attention_fwd(q, k, v, q2k_block_sparse_index, q2k_block_sparse_num):
+    """
+    block_sparse_mask: [bs, h, num_q_blocks, num_kv_blocks]. 
+        [*, *, i, j] = 1 means the i-th q block should attend to the j-th kv block.
+    """
+    # assert all elements in q2k_block_sparse_num can be devisible by 2
+    o, lse = block_sparse_fwd(q, k, v, q2k_block_sparse_index, q2k_block_sparse_num)
+    return o, lse
+
+def block_sparse_attention_fwd_16x16(q, k, v, block_sparse_map):
```
jzhang38: we do not need 16x16?

csrc/sliding_tile_attention/st_attn/__init__.py
```diff
@@ -1,12 +1,27 @@
 import math
```
jzhang38, line 1: actually let's have split this init into different files for vsa and sta.

fastvideo/v1/fastvideo_args.py
```diff
@@ -422,20 +423,20 @@ def check_fastvideo_args(self) -> None:
                 "Currently enabling vae_sp requires enabling vae_tiling, please set --vae-tiling to True."
             )
 
-        if len(self.text_encoder_configs) != len(self.text_encoder_precisions):
-            raise ValueError(
-                f"Length of text encoder configs ({len(self.text_encoder_configs)}) must be equal to length of text encoder precisions ({len(self.text_encoder_precisions)})"
-            )
+        # if len(self.text_encoder_configs) != len(self.text_encoder_precisions):
```
jzhang38: do not comment out this

scripts/inference/v1_inference_hunyuan_STA.sh
```diff
@@ -18,6 +18,7 @@ fastvideo generate \
     --num-inference-steps 50 \
     --guidance-scale 1 \
     --embedded-cfg-scale 6 \
+    --STA-mode STA_inference \
```
jzhang38: Do not add this flag to our existing inference example.

csrc/attn/vsa/block_sparse_h100.cu
```diff
@@ -0,0 +1,1347 @@
...
+                                        torch::TensorOptions().dtype(torch::kFloat).device(q.device()).memory_format(at::MemoryFormat::Contiguous));
+        
+
+    bf16*  o_ptr = reinterpret_cast<bf16*>(o.data_ptr<c10::BFloat16>());
+    bf16*  d_o   = reinterpret_cast<bf16*>(o_ptr);
+
+    float* l_ptr = reinterpret_cast<float*>(l_vec.data_ptr<float>());
+    float* d_l   = reinterpret_cast<float*>(l_ptr);
+
+    cudaDeviceSynchronize();
```
Edenzzzz, line 943: Is this required for correctness? Device sync is expensive

fastvideo/v1/attention/layer.py
```diff
@@ -135,6 +135,73 @@ def forward(
...
+
+    def forward(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        replicated_q: Optional[torch.Tensor] = None,
+        replicated_k: Optional[torch.Tensor] = None,
+        replicated_v: Optional[torch.Tensor] = None,
+        gate_compress: Optional[torch.Tensor] = None,
```
SolitaryThinker, line 150: instead of inheriting `DistributedAttention`, can we just add kwargs to DistributedAttention and check for required arguments in attn implementation? This means we also add kwargs to `self.impl.forward(q, k, v, gate_compress, ctx_attn_metadata)`. I feel inheriting for changes like these begin to quickly defeat the purpose of these abstractions?

fastvideo/v1/configs/sample/base.py
```diff
@@ -38,6 +38,7 @@ class SamplingParam:
     num_inference_steps: int = 50
     guidance_scale: float = 1.0
     guidance_rescale: float = 0.0
+    VSA_sparsity: float = 0.0
```
SolitaryThinker, line 41: `VSA_sparsity` -> `vsa_sparsity` ?

SolitaryThinker: I see we also use STA_mode. I'm open to either format, but currently we don't have other caps in our args

fastvideo/v1/models/dits/wanvideo.py
```diff
@@ -353,6 +356,155 @@ def forward(
         return hidden_states
 
 
+class WanTransformerBlock_VSA(nn.Module):
```
SolitaryThinker, line 359: If we can keep the same interface for DistributedAttention, we should be able to eliminate the need for a different VSA transformer block right? Just pass in whether VSA is enabled into init and define the `gate_compress` layer accordingly. Are there other differences in this block to standard wan2.1?

fastvideo/v1/models/loader/fsdp_load.py
SolitaryThinker: I'm slightly concerned about adding these kinds of hardcoded edge cases. But this is okay for now. If there are other techniques that require more perhaps we refactor this at that point? Otherwise there's shouldn't be a need to extend this.

fastvideo/v1/configs/sample/base.py
BrianChen1129: I think VSA and STA are method names, so we can keep them in caps？

fastvideo/v1/attention/layer.py
BrianChen1129: I think we still need DistributedAttention_VSA since there are quite a lot of changes inside forward().

fastvideo/v1/models/dits/wanvideo.py
BrianChen1129: I think it would be better to have a dedicated VSA block—otherwise, we’d need to add multiple if-else statements, which would make the code less clean.

fastvideo/v1/attention/layer.py
SolitaryThinker: Would it make sense to introduce new methods for this then? For example something like `prepare_for_all2all()` that implementations can override similar to `preprocess_qkv()`, and other backends can leave as no-ops?

BrianChen1129: I think the key design concern is to avoid forcing other base methods to adapt to `gate_compress`—we only want these arguments to show up on VSA-related methods. Since our goal with these abstractions is to enable reuse across different models, and VSA introduces a new attention pattern, it seems reasonable to define a separate attention method that inherits from `DistributedAttention`?

SolitaryThinker: okay sounds good then. Perhaps if/when there's other patterns we can reexamine how we can unify them

Issue #474, opened 2025-06-04
[Bug] torch 2.8 is required for 5090 to work


warycat:
### Environment

pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu128

### Describe the bug

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
fastvideo 0.1.0 requires torch==2.6.0, but you have torch 2.8.0.dev20250603+cu128 which is incompatible.
xformers 0.0.29.post2 requires torch==2.6.0, but you have torch 2.8.0.dev20250603+cu128 which is incompatible.
vllm 0.8.3 requires torch==2.6.0, but you have torch 2.8.0.dev20250603+cu128 which is incompatible.
vllm 0.8.3 requires torchaudio==2.6.0, but you have torchaudio 2.8.0.dev20250603+cu128 which is incompatible.
vllm 0.8.3 requires torchvision==0.21.0, but you have torchvision 0.23.0.dev20250603+cu128 which is incompatible.

### Reproduction
```
from fastvideo import VideoGenerator


def main():
    # Create a video generator with a pre-trained model
    generator = VideoGenerator.from_pretrained(
        "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
        num_gpus=1,  # Adjust based on your hardware
    )

    # Define a prompt for your video
    prompt = "A curious raccoon peers through a vibrant field of yellow sunflowers, its eyes wide with interest."

    # Generate the video
    video = generator.generate_video(
        prompt,
        return_frames=True,  # Also return frames from this call (defaults to False)
        output_path="my_videos/",  # Controls where videos are saved
        save_video=True,
    )


if __name__ == "__main__":
    main()
```

BrianChen1129:
Did you mean your environment only supports torch 2.8？

warycat:
I need cuda >= 12.8 for 5090. The supported pytorch is torch >= 2.7.

PR #483, opened 2025-06-06
[chore] Bump torch to 2.7.1 to support Blackwell

Edenzzzz:
Closes https://github.com/hao-ai-lab/FastVideo/issues/474. Wait for CIs to pass


SolitaryThinker:
The preinstalled flash-attn in the docker image is breaking ci as it is installed with the old torch 2.6.0 version. perhaps we can first open a PR that builds a docker image with torch 2.7 then use that in this PR. Then if it passes CI we should be able to merge. vllm recently upgraded to 2.7 as well

Edenzzzz:
sure can you do that and fix the permission issue so that everyone can pull?

kevin314:
It looks like torch 2.7 currently only officially supports CUDA 11.8, 12.6, and 12.8. FastVideo inference has a tendency to hang during NCCL initialization outside of 12.4. Some further investigation might be needed

Edenzzzz:
> It looks like torch 2.7 currently only officially supports CUDA 11.8, 12.6, and 12.8. FastVideo inference has a tendency to hang during NCCL initialization outside of 12.4. Some further investigation might be needed

This seems rare, can you try with torch 2.7?

kevin314:
I tried with torch 2.7 / cuda 12.8 and ran into this problem

Edenzzzz:
I can run examples basic.py on b200. Please share your command and error

SolitaryThinker:
I wonder if this issue is specific with runpod, could you try on modal @kevin314 ? The hanging issue occurs quite consistently on runpod

kevin314:
My issue was likely something else, I've since been able to use FastVideo on torch 2.7.1 on both RunPod and Modal, see #524 for updated Docker image

Edenzzzz:
@kevin314 I think this is ready to merge, but for some reason SSIM tests frequently times out in many PRs.  It runs on B200

SolitaryThinker (approved):

Edenzzzz:
the vae tests are failling, but it's not due to this PR (it only triggers the test)
cc @kevin314 @SolitaryThinker

kevin314:
I think we may need to re-calculate some of the reference latents/videos to address the flaky tests, it's possible that seemingly negligible differences have started to add up over several months

Edenzzzz:
> I think we may need to re-calculate some of the reference latents/videos to address the flaky tests, it's possible that seemingly negligible differences have started to add up over several months

I think it's only because of some recent PRs and docker changes. We should check the image build timeline against PRs

PR #487, opened 2025-06-07
[misc] Improve distributed related env variables and setup

jzhang38:
1. remove fastvideo_args.device and fastvideo_args.device_str. Use get_torch_device() instead.
2. remove unused functions in weight_utils.py
3. prioritize using maybe_init_distributed_environment_and_model_parallel
4. move check_fastvideo_args inside post_init


fastvideo/data_preprocess/v1_preprocess.py
SolitaryThinker, line 1: lets we just move this and related V1 preprocessing files (except those under `scripts`) to under `fastvideo/v1`?

fastvideo/data_preprocess/v1_preprocess.py
```diff
@@ -37,12 +29,9 @@ def main(args):
     pipeline_config_args = shallow_asdict(pipeline_config)
     pipeline_config_args.update(kwargs)
     fastvideo_args = FastVideoArgs(model_path=args.model_path,
-                                   num_gpus=world_size,
-                                   device_str="cuda",
+                                   num_gpus=get_world_size(),
```
SolitaryThinker, line 32: perhaps we should just remove num_gpus altogether, including in inference? If we always hardcode `num_gpus=get_world_size()` users would need to control it using `CUDA_VISIBLE_DEVICES`. When I added num_gpus I was only considering inference and also assumed `tp_size == sp_size`

SolitaryThinker:
Thanks!

JerryZhou54 (approved):
lgtm

fastvideo/data_preprocess/v1_preprocess.py
Edenzzzz: I think num_gpus is easier to control, if not specified we can use world size

PR #490, opened 2025-06-08
[bugfix] [Training] use diffusers fp32layernorm for wan2.1

SolitaryThinker:
Both wan2.1's official repo and diffuser's implementation explicitly force compute param dtype to float32

Example: https://github.com/huggingface/diffusers/blob/73a9d5856f2d7ae3637c484d83cd697284ad3962/src/diffusers/models/transformers/transformer_wan.py#L238

fp32layernorm taken from:
https://github.com/huggingface/diffusers/blob/73a9d5856f2d7ae3637c484d83cd697284ad3962/src/diffusers/models/normalization.py#L85

Finetuning results on Finetrainer's smol-crush dataset:

https://wandb.ai/wlsaidhi/finetrainers-wan/runs/vg27btrk?nw=nwuserwlsaidhi


fastvideo/v1/layers/layernorm.py
```diff
@@ -95,6 +96,22 @@ def forward(self, residual: torch.Tensor, x: torch.Tensor,
         return residual + x * gate
 
 
+# adapted from Diffusers: https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/normalization.py
+# NOTE(will): Needed to match behavior of diffusers and wan2.1 even while using
+# FSDP's MixedPrecisionPolicy
+class FP32LayerNorm(nn.LayerNorm):
```
jzhang38, line 102: should we subclass CustomOp?

fastvideo/v1/layers/layernorm.py
```diff
@@ -144,9 +167,11 @@ def forward(self, residual: torch.Tensor, x: torch.Tensor,
         # Apply residual connection with gating
         residual_output = residual + x * gate
         # Apply normalization
-        normalized = self.norm(residual_output)
+        normalized = self.norm(residual_output.float()).to(
```
jzhang38: This is not necessary as you already call float() and to(origin_dtype) in Float32LayerNorm?

fastvideo/v1/layers/layernorm.py
```diff
@@ -144,9 +167,11 @@ def forward(self, residual: torch.Tensor, x: torch.Tensor,
...
+            residual_output.dtype)
         # Apply scale and shift
-        modulated = normalized * (1.0 + scale) + shift
+        modulated = (normalized * (1.0 + scale) + shift).to(
+            residual_output.dtype)
```
jzhang38: Why do we need to(residual_output.dtype)  here?

fastvideo/v1/layers/layernorm.py
SolitaryThinker: not needed for now since we don't have cuda kernels for these

fastvideo/v1/layers/layernorm.py
SolitaryThinker: fixed

fastvideo/v1/layers/layernorm.py
SolitaryThinker: fixed

fastvideo/v1/layers/layernorm.py
Edenzzzz: [NVIDIA Apex ](https://github.com/NVIDIA/apex) has fast layernorm (though takes a long time to compile) and LigerKernel also has layernorm

SolitaryThinker: got it, lets leave this to a later ticket

fastvideo/v1/layers/layernorm.py
```diff
@@ -121,10 +139,15 @@ def __init__(
...
                                 dtype=dtype)
         elif norm_type == "layer":
-            self.norm = nn.LayerNorm(hidden_size,
-                                     elementwise_affine=elementwise_affine,
-                                     eps=eps,
-                                     dtype=dtype)
+            if compute_dtype == torch.float32:
+                self.norm = FP32LayerNorm(hidden_size,
+                                          elementwise_affine=elementwise_affine,
+                                          eps=eps)
```
JerryZhou54, line 145: Don't we need to pass in "dtype=dtype" here in order to set the potential weight and bias to dtype?

fastvideo/v1/layers/layernorm.py
```diff
@@ -163,23 +186,33 @@ def __init__(
...
                                 eps=eps)
         elif norm_type == "layer":
-            self.norm = nn.LayerNorm(hidden_size,
-                                     elementwise_affine=elementwise_affine,
-                                     eps=eps,
-                                     dtype=dtype)
+            if self.compute_dtype == torch.float32:
+                self.norm = FP32LayerNorm(hidden_size,
+                                          elementwise_affine=elementwise_affine,
+                                          eps=eps)
```
JerryZhou54, line 202: Same Q here

jzhang38 (approved):

fastvideo/v1/layers/layernorm.py
jzhang38: no we do not. This three are seperate thing:
1. fp32 weight, fp32 compute
2. bf16 weight, bf16 compute
3. bf16 weight, fp32 compute

I think we only need to care about 1 and 3.

JerryZhou54 (approved):

jd-nuva (commented):
great catch! you might also want to call .float() on hidden_states and ff_output to match HF implementation too?

https://github.com/huggingface/diffusers/blob/73a9d5856f2d7ae3637c484d83cd697284ad3962/src/diffusers/models/transformers/transformer_wan.py#L286-L303

Previously wanvideo.py has assertion on `# assert orig_dtype != torch.float32` that got commented out it seems, then the HF parity should be float32 assertion instead?

jd-nuva (commented):
my local version looks like this for referece, the gist of it is 

"compute attention in bfloat16, normalization and residual in float32"

```python
    def forward(
        self,
        hidden_states: torch.Tensor,
        encoder_hidden_states: torch.Tensor,
        temb: torch.Tensor,
        freqs_cis: Tuple[torch.Tensor, torch.Tensor],
    ) -> torch.Tensor:
        """
        We do all attention computation in bfloat16, but normalization and
        residual computation in float32
        """
        if hidden_states.dim() == 4:
            hidden_states = hidden_states.squeeze(1)

        orig_dtype = hidden_states.dtype
        assert orig_dtype != torch.float32
        e = self.scale_shift_table + temb.float()
        shift_msa, scale_msa, gate_msa, c_shift_msa, c_scale_msa, c_gate_msa = e.chunk(
            6, dim=1
        )
        assert shift_msa.dtype == torch.float32

        # 1. Self-attention
        norm_hidden_states = (
            self.norm1(hidden_states.float()) * (1 + scale_msa) + shift_msa
        ).to(orig_dtype)
        query, _ = self.to_q(norm_hidden_states)
        key, _ = self.to_k(norm_hidden_states)
        value, _ = self.to_v(norm_hidden_states)

        if self.norm_q is not None:
            query = self.norm_q.forward_native(query)
        if self.norm_k is not None:
            key = self.norm_k.forward_native(key)

        # QKV has [B, seq, num_heads * hidden_dim] here
        query = query.squeeze(1).unflatten(2, (self.num_attention_heads, -1))
        key = key.squeeze(1).unflatten(2, (self.num_attention_heads, -1))
        value = value.squeeze(1).unflatten(2, (self.num_attention_heads, -1))

        # We QKV reshaped to be [bs, seq_len, num_heads, head_size]
        # Apply rotary embeddings
        cos, sin = freqs_cis
        query, key = (
            # first arg expects [num_tokens, num_heads, head_size]
            _apply_rotary_emb(query, cos, sin, is_neox_style=False),
            _apply_rotary_emb(key, cos, sin, is_neox_style=False),
        )

        attn_output, _ = self.attn1(query, key, value)
        attn_output = attn_output.flatten(2)  # [B, seq, num_head * hidden_dim]
        attn_output, _ = self.to_out(attn_output)  # [B, seq, num_head * hidden_dim]
        attn_output = attn_output.squeeze(1)

        null_shift = null_scale = torch.tensor([0], device=hidden_states.device)
        # The first argument `hidden_states.float()` ensures the residual connection happens in float32.
        # self_attn_residual_norm will return float32 tensors because its internal ops are float32.
        norm_hidden_states, hidden_states = self.self_attn_residual_norm(
            hidden_states.float(), attn_output, gate_msa, null_shift, null_scale
        )
        norm_hidden_states = norm_hidden_states.to(orig_dtype)

        # 2. Cross-attention
        attn_output = self.attn2(
            norm_hidden_states, context=encoder_hidden_states, context_lens=None
        )
        norm_hidden_states, hidden_states = self.cross_attn_residual_norm(
            hidden_states.float(), attn_output, 1, c_shift_msa, c_scale_msa
        )
        norm_hidden_states = norm_hidden_states.to(orig_dtype)

        # 3. Feed-forward
        ff_output = self.ffn(norm_hidden_states)
        hidden_states = self.mlp_residual(
            hidden_states.float(), ff_output.float(), c_gate_msa
        )

        return hidden_states.to(orig_dtype)
```

SolitaryThinker:
good catch @jd-nuva, I'm going to open another PR and address these. Thanks!

PR #491, opened 2025-06-09
[misc] rename dp_size to hdsp_replicate_dim

jzhang38:
dp_size and dp_shards actually **has nothing to do with data parallel**.

Consider the case: num gpu = 16, dp_size=2, dp_shards=8, sp_size=4. 

We are actually using **sequence_parallel =4 and data_parallel = 4**. 

dp_size and dp_shards only control the **weight sharding** plan in hdsp/fsdp.


SolitaryThinker (approved):
makes sense, thanks

jzhang38:
@SolitaryThinker Done

Edenzzzz (approved):
LGTM

SolitaryThinker (approved):

PR #492, opened 2025-06-09
[Feat][Dataloader] 1/n Refactor parquet map-style dataloader

jzhang38:


fastvideo/v1/training/training_pipeline.py
```diff
@@ -92,22 +92,15 @@ def initialize_training_pipeline(self, training_args: TrainingArgs):
...
-            prefetch_factor=2,
-            shuffle=False,
-            pin_memory=True,
-            pin_memory_device=f"cuda:{torch.cuda.current_device()}",
-            drop_last=True)
+            training_args.train_batch_size,
+            num_data_workers=training_args.dataloader_num_workers,
+            drop_last=True,
+            text_padding_length=
+            512,  # TODO(peiyuan): set this according to text length of each model.
```
SolitaryThinker: Can use: `training_args.text_encoder_configs[0].arch_config.text_len`
Speaking of which, would this dataloader handle multiple text encoders or how difficult would it to extend and support? Don't need to support it in this PR or refactor, but feel it would be good to keep in mind.

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,492 @@
...
+    def __len__(self):
+        return len(self.sp_group_local_indices) // self.batch_size
+
+
+def get_parquet_files_and_length(path: str):
+    lengths = []
+    file_names = []
+    for root, _, files in os.walk(path):
+        for file in sorted(files):
+            if file.endswith('.parquet'):
```
SolitaryThinker, line 88: Should we raise some kind of warning or error if there's no parquet files at all in the directory?

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,492 @@
...
+        text_padding_length: int = 512,
+    ):
+        super().__init__()
+        self.path = path
+        self.cfg_rate = cfg_rate
+        if cfg_rate > 0.0:
+            raise ValueError(
+                "cfg_rate > 0.0 is not supported for now because it will trygger bug when num_data_workers > 0"
+            )
+        # self.rng = torch.Generator().manual_seed(seed)
```
SolitaryThinker: remove

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,492 @@
...
+        for i, row in enumerate(rows):
+            # Get tensors from row
+            data = self._get_torch_tensors_from_row_dict(row)
+            latents, emb = data["vae_latent"], data["text_embedding"]
+
+            # if torch.rand(1, generator=self.rng) < self.cfg_rate:
+            #     # all zero
+            #     emb = torch.zeros(self.text_padding_length, emb.shape[1], dtype=emb.dtype, device=emb.device)
+            #     mask = torch.zeros(self.text_padding_length, dtype=emb.dtype, device=emb.device)
+            # else:
```
SolitaryThinker: remove if not needed

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,492 @@
...
+        batch_size,
+        cfg_rate=cfg_rate,
+        drop_last=drop_last,
+        text_padding_length=text_padding_length,
+        seed=seed)
+
+    loader = StatefulDataLoader(
+        dataset,
+        batch_sampler=dataset.sampler,
+        # prefetch_factor=4,
```
SolitaryThinker: is this needed?

SolitaryThinker (changes requested):

fastvideo/v1/training/training_pipeline.py
jzhang38: I think it would be easy to support multiple text encoder. I think that is just an extral column in the parquet file.

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,486 @@
...
+        logger.info("Total samples: %d", total_samples)
+        logger.info("Average speed: %.2f samples/sec", samples_per_sec)
+        logger.info("Time per batch: %.2f ms", elapsed / total_batches * 1000)
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    finally:
+        cleanup_dist_env_and_memory()
```
Edenzzzz: It may be better to decouple usage example from the class def?

fastvideo/v1/distributed/parallel_state.py
```diff
@@ -923,6 +923,8 @@ def maybe_init_distributed_environment_and_model_parallel(
     rank = int(os.environ.get("RANK", 0))
 
     torch.cuda.set_device(local_rank)
+    # set cuda visible devices
+    os.environ["CUDA_VISIBLE_DEVICES"] = str(local_rank)
```
Edenzzzz: Is this required?

fastvideo/v1/fastvideo_args.py
```diff
@@ -605,7 +605,8 @@ def from_cli_args(cls, args: argparse.Namespace) -> "TrainingArgs":
             # Use getattr with default value from the dataclass for potentially missing attributes
             else:
                 default_value = getattr(cls, attr, None)
-                kwargs[attr] = getattr(args, attr, default_value)
+                if getattr(args, attr, default_value) is not None:
+                    kwargs[attr] = getattr(args, attr, default_value)
```
Edenzzzz: try 
```
value = getattr(args, attr, default_value)
if value is not None:
    kwargs[attr] = value
```

fastvideo/v1/dataset/parquet_dataset_map_style.py
```diff
@@ -0,0 +1,486 @@
...
+        seed=seed)
+
+    loader = StatefulDataLoader(
+        dataset,
+        batch_sampler=dataset.sampler,
+        collate_fn=passthrough,
+        num_workers=num_data_workers,
+        pin_memory=True,
+        persistent_workers=num_data_workers > 0,
+    )
```
Edenzzzz, line 320: If latent size is small, we can use `pin_memory_device` to move directly to cuda

Edenzzzz (changes requested):

fastvideo/v1/distributed/parallel_state.py
jzhang38: Not required but good to have?

Edenzzzz: After calling any `torch.cuda` cuda is initialized, so this will have no effect

fastvideo/v1/dataset/parquet_dataset_map_style.py
jzhang38: https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader
If not given, the current [accelerator](https://docs.pytorch.org/docs/stable/torch.html#accelerators) will be the default. This argument is discouraged and subject to deprecated.
I guess it is not necessary

Edenzzzz (approved):

Edenzzzz:
LGTM, need to add training functionality tests later

PR #494, opened 2025-06-10
Fix pre-commit CI

Edenzzzz:
Fixes https://github.com/hao-ai-lab/FastVideo/pull/493
@JerryZhou54 Please make sure the CIs pass next time


BrianChen1129:
This commit will cause a new bug in inference. Did you test it?

PR #495, opened 2025-06-10
[bugfix][Cli Inference] Resolve runtime errors when running fastvideo generate

JerryZhou54:


BrianChen1129 (approved):

fastvideo/v1/fastvideo_args.py
```diff
@@ -387,9 +387,8 @@ def from_cli_args(cls, args: argparse.Namespace) -> "FastVideoArgs":
             # Use getattr with default value from the dataclass for potentially missing attributes
             else:
                 default_value = getattr(cls, attr, None)
-                attr = getattr(args, attr, default_value)  # type: ignore
-                if attr is not None:
-                    kwargs[attr] = attr
+                if getattr(args, attr, default_value) is not None:
+                    kwargs[attr] = getattr(args, attr, default_value)
```
Edenzzzz, line 391: Aren't these logically equivalent?

Edenzzzz:
Some of my opinions:
1. It's discouraged but acceptable to merge a PR directly without review when 1) the fix is small but needs to be quickly merged 2) the person is confident about the change 3) the CIs pass. The SGLang core developers do this a lot because they have more PRs than reviewers, but since we are at an early stage, it's fine if we want to avoid this. I saw a few training PRs from @SolitaryThinker merged by himself so wasn't aware we wanted to disallow this.
2. The CIs are not reliable. SSIMs tests often fail on their own (https://github.com/hao-ai-lab/FastVideo/actions/runs/15552269454/job/43785040653?pr=495), and encoder/transformer tests sometimes hit OOM(https://github.com/hao-ai-lab/FastVideo/actions/runs/15538454651/job/43778178282?pr=470).  We want to rely on CIs, not human eyes. 
3. If a PR passes all CIs but still causes an error, please directly provide the reproducible command and error under/linking to the PR, so that we can validate. When submitting a new fix, please provide context on why the old way fails and this works. Also, ideally we should add a new test for that functionality in the same or a follow-up PR.

What do you guys think?

fastvideo/v1/fastvideo_args.py
BrianChen1129: But it indeed introduces an error when running bash scripts/inference/v1_inference_wan.sh. Next time, we may need to give such reproducible commands/scripts when submitting a bug-fixing PR, then others can also validate.

Edenzzzz: Then maybe let's add that to CI?

BrianChen1129:
Thanks for the suggestion! For 1, to maintain code quality and reliability, I think we should require at least one review before merging any changes that may affect inference or training.
For 2 and 3. Currently, passing the Cls alone doesn't guarantee full correctness across all features. We need to design additional tests that better cover the current inference and training pipelines. And we'd better provide reproducible commands/scripts when submitting a bug-fixing PR, so others can offer help.

fastvideo/v1/fastvideo_args.py
BrianChen1129: Yeah, we should add that.

Edenzzzz:
My intention was to not call `getattr` twice, and yes there was a typo.
it should be like this
```
                value = getattr(args, attr, default_value)
                if value is not None:
                    kwargs[attr] = value
```

PR #500, opened 2025-06-10
[bugfix] fix validation pipeline initialization

SolitaryThinker:
directly call validation pipeline's  `__init__` in wan_training_pipeline since we already have TrainingArgs available


SolitaryThinker:
fix in https://github.com/hao-ai-lab/FastVideo/pull/492 already

PR #501, opened 2025-06-10
[CI] Add inference test

Edenzzzz:
Future training tests can be added like this

### DIFF
--- a/fastvideo/v1/tests/inference/test_basic_generation.py
+++ b/fastvideo/v1/tests/inference/test_basic_generation.py
@@ -1,0 +1,46 @@
+import pytest
+from fastvideo import VideoGenerator
+import os
+import glob
+
+@pytest.mark.parametrize("use_fsdp_inference", [True, False])
+@pytest.mark.parametrize("use_cpu_offload", [True, False])
+@pytest.mark.parametrize("num_gpus", [1, 2])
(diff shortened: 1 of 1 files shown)


fastvideo/v1/tests/inference/test_basic_generation.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.parametrize("use_cpu_offload", [True, False])
+@pytest.mark.parametrize("num_gpus", [1, 2])
+def test_multiple_generations(use_fsdp_inference: bool, use_cpu_offload: bool, num_gpus: int):
+    """Test that multiple video generations don't throw any errors"""
+    output_path = "./test_output"
+    if num_gpus == 1 and use_fsdp_inference:
+        pytest.skip("FSDP inference is not needed with 1 GPU")
+    try:
+        generator = VideoGenerator.from_pretrained(
+            "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
```
Edenzzzz: Should we test more models?

BrianChen1129: I think 1 model is enough? Since they share the same base pipeline.

SolitaryThinker:
Doesn't SSIM already cover these functionalities?

Edenzzzz:
> Doesn't SSIM already cover these functionalities?

Yeah I see this is actually a duplicate

PR #502, opened 2025-06-10
[Refactor] Move attention under layers

Edenzzzz:
This structure makes more sense


SolitaryThinker:
I'm open to this, although I prefer current organization. We follow vllm's structure while sglang's attention/ is under layers/

Edenzzzz:
It's not a major issue, but there are many folders under `fastvideo/v1`, we might want to organize better

PR #504, opened 2025-06-11
[CI] [Training] Initial e2e small training test

SolitaryThinker:
Adds a small e2e finetune run on wan2.1 1.3B that runs on 4xA40s. Verifies that the loss and other metrics is within a threshold of a previous, verified run with same parameters on same GPU.


fastvideo/v1/training/training_pipeline.py
```diff
@@ -177,114 +179,118 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
...
+
+                # Run validation inference
+                with torch.no_grad(), torch.autocast("cuda",
+                                                     dtype=torch.bfloat16):
+                    output_batch = self.validation_pipeline.forward(
+                        batch, training_args)
+                    samples = output_batch.output
+
+                # Re-enable gradients for training
+                transformer.requires_grad_(True)
```
Edenzzzz: Maybe no need to call this if we decorate `validation_pipeline.forward` with `@torch.no_grad`?

SolitaryThinker: Okay, once the first CI run passes, I can remove this and reun CI :)

fastvideo/v1/tests/training/test_training_loss.py
```diff
@@ -0,0 +1,142 @@
...
+    wandb_dir = f"wandb/latest-run"
+    summary_file = f"{wandb_dir}/files/wandb-summary.json"
+
+    reference_wandb_summary = json.load(open(reference_wandb_summary_file))
+    wandb_summary = json.load(open(summary_file))
+
+    fields_and_thresholds = {
+        'avg_step_time': 0.5,
+        'grad_norm': 0.1,
+        'step_time': 0.1,
```
Edenzzzz: `step_time` can change as we optimize more? It might make it more troublesome to maintain

SolitaryThinker: I feel if we expect a significant change this test should be explicitly updated. We can add other tests with any optional optimizations turned on. I will increase this threshold to 1 second for now

fastvideo/v1/training/training_pipeline.py
SolitaryThinker: removed `requires_grad_` and moved the rest to the bottom. Let's see if CI passes

fastvideo/v1/tests/training/test_training_loss.py
Edenzzzz: Yeah sure, maybe something like 0.5 sec

Edenzzzz (approved):

Edenzzzz:
@SolitaryThinker please add some filters for the training test so that not every change triggers CI

SolitaryThinker:
Other than docs, we want to currently always run ssim and e2e training test

SolitaryThinker:
wandb run on this commit for future reference https://wandb.ai/wlsaidhi/wan_finetune_m2/runs/3aohue5m?nw=nwuserwlsaidhi

fastvideo/v1/tests/training/test_training_loss.py
```diff
@@ -0,0 +1,142 @@
...
+    parser = TrainingArgs.add_cli_args(parser)
+    parser = FastVideoArgs.add_cli_args(parser)
+    
+    # Set the arguments as they are in finetune_v1_test.sh
+    args = parser.parse_args([
+        "--model_path", "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
+        "--inference_mode", "False",
+        "--pretrained_model_name_or_path", "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
+        "--cache_dir", "/home/.cache",
+        "--data_path", "data/crush-smol_parq/combined_parquet_dataset",
```
Edenzzzz, line 40: Could you write a README for data preprocessing?

PR #508, opened 2025-06-12
[CI] Restrict training CI to v1

Edenzzzz:
cc @SolitaryThinker

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -12,14 +12,6 @@ on:
     paths:
       - "fastvideo/**/*.py"
       - ".github/workflows/pr-test.yml"
-    paths-ignore:
-      - 'docs/**'
-      - '*.md'
-      - '*.rst'
-      - 'LICENSE'
-      - 'README.md'
-      - 'CHANGELOG.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -84,6 +85,15 @@ jobs:
...
               - 'fastvideo/v1/attention/**'
+            training-test:
+              - '**'
+              - '!docs/**'
+              - '!*.md'
+              - '!*.rst'
+              - '!LICENSE'
+              - '!README.md'
+              - '!CHANGELOG.md'
+              - '!CONTRIBUTING.md'
```
kevin314: The workflow filter at the top should ensure the workflow only runs on .py file changes, which I believe makes any job-level exclusions for non-Python files redundant
```
on:
  push:
    branches: [main]
    paths:
      - "fastvideo/**/*.py"
      - ".github/workflows/pr-test.yml"
```

Edenzzzz: ok then let's include only v1 files?

kevin314: that works

PR #512, opened 2025-06-13
[CI] Add current PR test workflow to Buildkite/Modal

kevin314:
TODO

- [ ] Add block step for maintainers to approve

image

Resulting Buildkite pipeline looks very similar to what we have in Github Actions currently, a single PR test pipeline with separate jobs for each test. See [https://buildkite.com/fvtest/fastvideo/builds/20#_](https://buildkite.com/fvtest/fastvideo/builds/20#_) for a sample build.

It looks like Modal does not have A40s, so we may need to recalculate the hardcoded latents being used for encoder and vae tests on the next best suited GPU (L40S?) And possibly re-generate some SSIM reference videos.


Edenzzzz:
If we are upgrading CIs, we should also add a standard docker environment that works for torch 2.7 and STA compilation (and misc. stuff like zsh and vim). The current one is very hard to use

SolitaryThinker:
can we move the modal directory to `fastvideo/v1/tests/modal/`?

Edenzzzz (changes requested):
Could you write a docker file that supports torch 2.7.1, flash-attn, and STA & VSA installation?

SolitaryThinker (approved):

PR #513, opened 2025-06-13
[Feature][Training]vsa for t2v training ready

BrianChen1129:


csrc/attn/vsa/__init__.py
```diff
@@ -154,6 +154,7 @@ def block_sparse_attention_fwd(q, k, v, q2k_block_sparse_index, q2k_block_sparse
     return o, lse
 
 def block_sparse_attention_backward(q, k, v, o, l_vec, grad_output, k2q_block_sparse_index, k2q_block_sparse_num):
+    grad_output = grad_output.contiguous()
```
jzhang38, line 157: Is this necessary?

fastvideo/v1/dataset/parquet_dataset_iterable_style.py
SolitaryThinker, line 1: Revert these?

fastvideo/v1/fastvideo_args.py
```diff
@@ -422,6 +433,13 @@ class TrainingArgs(FastVideoArgs):
     # master_weight_type
     master_weight_type: str = ""
 
+    # For fast checking in LoRA pipeline
+    training_mode: bool = True
```
SolitaryThinker: Remove this. It is a bug as it overrides the property `training_mode` in FastVideoArgs

fastvideo/v1/training/wan_training_pipeline.py
```diff
@@ -148,14 +164,18 @@ def train_one_step(
                         [1000.0],
                         device=noisy_model_input.device,
                         dtype=torch.bfloat16)
+                attn_metadata = VideoSparseAttentionMetadata(
+                    current_timestep=timesteps,
+                    dit_seq_shape=dit_seq_shape,
+                    VSA_sparsity=current_vsa_sparsity)
+
```
SolitaryThinker: Can we build this only if using VSA?

fastvideo/v1/dataset/parquet_dataset_iterable_style.py
SolitaryThinker, line 1: Could we revert changes in this file if it still passes linter? Feel these kinds of style changes make it hard to rebase/merge. We can almost update these in a later PR

fastvideo/v1/dataset/parquet_dataset_map_style.py
SolitaryThinker, line 1: same as above

fastvideo/v1/training/wan_training_pipeline.py
```diff
@@ -295,6 +321,8 @@ def forward(
                 self.training_args.logit_mean,
                 self.training_args.logit_std,
                 self.training_args.mode_scale,
+                self.training_args.pipeline_config.dit_config.patch_size,
+                current_vsa_sparsity,
```
SolitaryThinker, line 337: Perhaps the CI will test this, but we should only do these if VSA is enabled

SolitaryThinker (approved):
Overall lgtm. We just need to be careful that non-vsa still works and vsa related operations are not performed if not using VSA

csrc/attn/vsa/__init__.py
Edenzzzz: this should have no effect

fastvideo/v1/training/wan_training_pipeline.py
```diff
@@ -148,8 +165,17 @@ def train_one_step(
...
                         device=noisy_model_input.device,
                         dtype=torch.bfloat16)
+
+                if envs.FASTVIDEO_ATTENTION_BACKEND == "VIDEO_SPARSE_ATTN":
+                    attn_metadata = VideoSparseAttentionMetadata(
+                        current_timestep=timesteps,
+                        dit_seq_shape=dit_seq_shape,
+                        VSA_sparsity=current_vsa_sparsity)
+                else:
+                    attn_metadata = None
```
Edenzzzz, line 175: the `vsa_available` check should be applied here?

Edenzzzz: the `vsa_available` check should be applied here?

csrc/attn/vsa/__init__.py
BrianChen1129: The weird thing is without this line it will report og is not contiguous issue

fastvideo/v1/attention/layer.py
```diff
@@ -183,8 +183,7 @@ def forward(
                                                      scatter_dim=2,
                                                      gather_dim=1)
 
-        qkvg = self.impl.preprocess_qkv(
-            qkvg, ctx_attn_metadata)  # (yongqi) pass latent shape here?
+        qkvg = self.impl.preprocess_qkv(qkvg, ctx_attn_metadata)
```
Edenzzzz: better rename as `attn_impl`?

Edenzzzz (changes requested):

fastvideo/v1/attention/layer.py
BrianChen1129: make sense. I'll change it

csrc/attn/vsa/__init__.py
Edenzzzz: that's weird, leave a comment and investigate later?

Edenzzzz: Potentially because the attention output during forward is not contiguous

BrianChen1129: yeah, we can look deep into it later

fastvideo/v1/training/wan_training_pipeline.py
BrianChen1129: in fact we do not need the vsa_available in this file, since if `envs.FASTVIDEO_ATTENTION_BACKEND == "VIDEO_SPARSE_ATTN"` it already shows vsa is available

csrc/attn/README.md
```diff
@@ -6,6 +6,15 @@
 ## Installation
 We test our code on Pytorch 2.5.0 and CUDA>=12.4. Currently we only have implementation on H100.
 First, install C++20 for ThunderKittens:
+```bash
+sudo apt update
+sudo apt install gcc-11 g++-11
```
Edenzzzz, line 11: This is still not working for me using the current Dockerfile. Could you share your local working environment, or a docker image?
image

Edenzzzz: Actually, I think the issue is that current docker's ubuntu version is old. I installed the Ubuntu Toolchain PPA  and it works

fastvideo/v1/attention/layer.py
BrianChen1129: it's a good point, but for now, we'd better rename it later since we're also adding other new features.

fastvideo/v1/training/wan_training_pipeline.py
Edenzzzz: I think that part is problematic. 
The returned attention backend doesn't override the env variable, while it should 
https://github.com/hao-ai-lab/FastVideo/blob/3eec1281cfe3cb20984bed909dfcb8ef9d1c54cf/fastvideo/v1/attention/selector.py#L118

Edenzzzz: It will return `VIDEO_SPARSE_ATTN` no matter whether it's compiled?

BrianChen1129: It will return FLASH_ATTN if VSA is not found. You can refer to here: https://github.com/hao-ai-lab/FastVideo/blob/3eec1281cfe3cb20984bed909dfcb8ef9d1c54cf/fastvideo/v1/platforms/cuda.py#L109

fastvideo/v1/attention/backends/video_sparse_attn.py
```diff
@@ -5,7 +5,11 @@
 
 import torch
 from einops import rearrange
-from vsa import video_sparse_attn
+
+try:
+    from vsa import video_sparse_attn
+except ImportError:
+    video_sparse_attn = None
```
Edenzzzz, line 12: Raise an `NotImplementedError` in forward if import fails (though theoretically, it should never throw it since `get_attn_backend_cls` would return something else)

fastvideo/v1/training/wan_training_pipeline.py
```diff
@@ -273,11 +294,19 @@ def forward(
...
+        vsa_sparsity = self.training_args.VSA_sparsity
+        vsa_decay_rate = self.training_args.VSA_decay_rate
+        vsa_decay_interval_steps = self.training_args.VSA_decay_interval_steps
         for step in range(self.init_steps + 1,
                           self.training_args.max_train_steps + 1):
             start_time = time.perf_counter()
 
+            current_decay_times = min(step // vsa_decay_interval_steps,
+                                      vsa_sparsity // vsa_decay_rate)
+            current_vsa_sparsity = current_decay_times * vsa_decay_rate
```
Edenzzzz: Do not calculate these if backend is not VSA

Edenzzzz (changes requested):

fastvideo/v1/attention/backends/video_sparse_attn.py
BrianChen1129: This is used to prevent 'VSA not found' error from breaking vanilla training when VSA is not installed.

fastvideo/v1/training/wan_training_pipeline.py
Edenzzzz: actually, can we move these inside `train_one_step`? The main pipeline should be kept clean and general.

fastvideo/v1/attention/backends/video_sparse_attn.py
Edenzzzz: Yeah I mean add an error in `forward`

fastvideo/v1/training/wan_training_pipeline.py
Edenzzzz: but it should overwrite `envs.FASTVIDEO_ATTENTION_BACKEND`

fastvideo/v1/training/wan_training_pipeline.py
BrianChen1129: Now we can put it outside `train_one_step` since we need to pass it to wandb.log and it needs to pass step into `train_one_step`.

Edenzzzz: ok sure

fastvideo/v1/attention/backends/video_sparse_attn.py
BrianChen1129: Why do we need to add error in the forward? If it can go into forard() vsa is surely existing.

Edenzzzz: we should make that explicit because looking at the file alone forward would still run, with `None`. People won't dig that much and this can seem semantically confusing.

fastvideo/v1/training/wan_training_pipeline.py
BrianChen1129: I see, `envs.FASTVIDEO_ATTENTION_BACKEND` is not used to ensure we're using the backend, let me change it.

BrianChen1129:
> LGTM. Let's refer to SGLang's attn backend & import handling in the future https://github.com/sgl-project/sglang/blob/50876abc47f1be4739f837e477bcf7506fafcff3/python/sglang/lang/backend/litellm.py#L30

Thanks for reviewing and sharing this!

PR #517, opened 2025-06-14
[Kernel] Remove all syncs from STA & VSA kernels

Edenzzzz:
None of the `cudaStreamSynchronize` (blocks the CPU) or the `cudaDeviceSynchronize` (blocks all streams including comm) is useful, but they are very expensive. Precisions are identical before and after removing.

# STA
`python csrc/attn/benchmarks/bench_sta.py`
### Before PR
(FastVideo)
### After PR
iverage time for forward pass (ms) 195 96

# VSA
`python csrc/attn/benchmarks/bench_vsa.py `
### Before PR
image
### After PR
Pasted Graphic 2


Edenzzzz:
cc @BrianChen1129 @jzhang38 @zhisbug

csrc/attn/st_attn/st_attn_h100.cu
```diff
@@ -5,8 +5,15 @@
 #include <iostream>
 #include <stdio.h>
 
-#define CLAMP(value, min, max) ((value) < (min) ? (min) : ((value) > (max) ? (max) : (value)))
-#define ABS(x) ((x) < 0 ? -(x) : (x))
+// #define CLAMP(value, min, max) ((value) < (min) ? (min) : ((value) > (max) ? (max) : (value)))
```
Edenzzzz, line 9: `#define` cannot trigger compiler optimizations is not type-safe. See https://wiki.sei.cmu.edu/confluence/display/c/PRE00-C.%2BPrefer%2Binline%2Bor%2Bstatic%2Bfunctions%2Bto%2Bfunction-like%2Bmacros

BrianChen1129:
Have you tried inference videos for both STA and VSA?

Edenzzzz:
> Have you tried inference videos for both STA and VSA?

The current inference scripts seem to be still from v0. I will try and I don't think it will be any different, as the precision tests pass.

BrianChen1129:
> > Have you tried inference videos for both STA and VSA?
> 
> The current inference scripts seem to be still from v0. I will try and I don't think it will be any different, as the precision tests pass.

The STA inference script can be found [here](https://github.com/hao-ai-lab/FastVideo/blob/main/scripts/inference/v1_inference_wan_STA.sh). VSA training script is [here](https://github.com/hao-ai-lab/FastVideo/blob/main/scripts/finetune/finetune_v1_VSA.sh). And I do think it should pass the e2e test to ensure no other issues.

Edenzzzz:
Here's the output on H100 
`bash scripts/inference/v1_inference_wan_STA.sh`
### Before PR

https://github.com/user-attachments/assets/453b6332-f909-455b-b059-fc31eab347c1

### After PR
https://github.com/user-attachments/assets/720e91ca-55e2-4291-9bfe-4735e77ecc2d

They are identical. I've seen lots of cuda kernels, and none of them use syncs like this.
cc @BrianChen1129

csrc/attn/vsa/block_sparse_h100.cu
```diff
@@ -940,7 +940,7 @@ block_sparse_attention_forward(torch::Tensor q, torch::Tensor k, torch::Tensor v
     float* l_ptr = reinterpret_cast<float*>(l_vec.data_ptr<float>());
     float* d_l   = reinterpret_cast<float*>(l_ptr);
 
-    cudaDeviceSynchronize();
```
zhisbug, line 943: I am not sure if we can remove these synchronizations?

Edenzzzz: Everything in the same stream will execute sequentially; there's no need for any synchronization

Edenzzzz: Thunderkittens uses it to synchronously report errors on the cpu side or for timing (like `CUDA_LAUNCH_BLOCKING=1` in torch), but they are a bit careless in making them all over the place

Edenzzzz:
> Have you tried inference videos for both STA and VSA?

Do you have a precision test for VSA? e.g. against Flexattn

jzhang38:
Precision test for VSA: 

https://github.com/hao-ai-lab/FastVideo/blob/main/csrc/attn/test/test_block_sparse.py

Edenzzzz:
@jzhang38 The precision for VSA is also identical.

Edenzzzz:
Confirmed with Dan Fu that these can be removed, and added a device guard for corner cases
cc @jzhang38 @zhisbug 
https://github.com/HazyResearch/ThunderKittens/pull/129#issuecomment-2981992662

SolitaryThinker:
could rebase with main and see if CI pass?

Edenzzzz:
I'm restricting STA & VSA tests to kernel changes only, because H100 is expensive. Will add another common inference test

SolitaryThinker:
We should keep vsa and sta enabled as there are other code that effects them. These tests also are very short.

Edenzzzz:
> We should keep vsa and sta enabled as there are other code that effects them. These tests also are very short.

Added them back and added precision tests. Note that `workflow_dispatch` can have at most 10 variables, so I removed `custom_image` and used hardcoded string. Should probably try a different approach later.

Edenzzzz:
Let's merge this and I will open another PR to split the pr test yaml and add Lora tests? @SolitaryThinker

jzhang38 (approved):

jzhang38:
lgtm

PR #525, opened 2025-06-16
[Bugfix][Inference]Fix envs.attn_backend

BrianChen1129:


fastvideo/v1/fastvideo_args.py
```diff
@@ -63,7 +63,7 @@ class FastVideoArgs:
 
     # STA (Sliding Tile Attention) parameters
     mask_strategy_file_path: Optional[str] = None
-    STA_mode: Optional[str] = None
+    STA_mode: str = "STA_inference"
```
SolitaryThinker: Could we change STA_mode to use enums

BrianChen1129: yeah, let me change it

fastvideo/v1/platforms/cuda.py
```diff
@@ -155,8 +151,6 @@ def get_attn_backend_cls(cls,
                     VideoSparseAttentionBackend)
                 logger.info("Using Video Sparse Attention backend.")
 
-                # Overwrite with the actual backend
-                envs.FASTVIDEO_ATTENTION_BACKEND = "VIDEO_SPARSE_ATTN"
```
Edenzzzz, line 159: You still need to disable VSA when import fails, and throw a warning

Edenzzzz (changes requested):

fastvideo/v1/platforms/cuda.py
Edenzzzz: Actually just throw an error if the users wants STA/VSA and it's not available

Edenzzzz (approved):

Edenzzzz (changes requested):
Wait for CIs to pass

Edenzzzz (approved):

PR #526, opened 2025-06-17
[CI][bugfix] Use new 3.12 docker image

SolitaryThinker:


Edenzzzz:
Test this on B200 first?

kevin314:
> This image is still not working on Blackwell because it uses the cuda 12.6 version of torch 2.7. Try `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu128 --force-reinstall` in the docker file?

should we try to resolve this with FastVideo's dependency handling (like in pyproject.toml) versus in the docker image?

Edenzzzz:
Yes you can try adding `--index-url https://download.pytorch.org/whl/cu128` to it 
image

Edenzzzz:
The SSIM tests failed, and also we need to upgrade pytorch

PR #540, opened 2025-06-21
[Feature][Training]Add diffusers format cp saving

BrianChen1129:


SolitaryThinker:
is it possible to generate the reverse mapping from the forward? This way we only need to define these mappings once per model and will also lower the chance of errors. If needed we could slightly change how the mapping is specified

SolitaryThinker:
does resume_from_checkpoint still work with this PR?

BrianChen1129:
> does resume_from_checkpoint still work with this PR?

Yeah, resume_from_checkpoint will use distributed checkpoint. I just changed the format of inference checkpoint.

PR #546, opened 2025-06-24
[CI] Add lora test

Edenzzzz:


fastvideo/v1/models/loader/fsdp_load.py
```diff
@@ -80,12 +79,13 @@ def maybe_load_fsdp_model(
                                      output_dtype,
                                      cast_forward_inputs=False)
 
-    set_mixed_precision_policy(master_dtype=default_dtype,
```
Edenzzzz, line 84: Don't find master_dtype useful, and it also doesn't align with torch

Edenzzzz:
I've rebased but don't know why it's not running buildkite

SolitaryThinker:
It looks like you didn't add the reference videos for lora similarity test. `*.mp4` is in `.gitignore` so you'll need to do `git add -f`

image

.pre-commit-config.yaml
```diff
@@ -56,20 +56,14 @@ repos:
   rev: v1.7.7
   hooks:
   - id: actionlint
-- repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v1.15.0
-  hooks:
-  - id: mypy
-    args: [--python-version, '3.10', --follow-imports, "skip", ]
-    additional_dependencies: [types-cachetools, types-setuptools, types-PyYAML, types-requests]
```
SolitaryThinker: Let's not completely remove mypy for now. I've already reduced the strictness slightly.

Edenzzzz: sure

SolitaryThinker:
seems like there's still some errors in lora inference test. Also the SSIM test shouldn't time out (should finish in ~24 minutes or so)

Edenzzzz:
There's still a bug with FSDP that doesn't allow re-register params for mutli-lora. Reaching out for a solution
https://github.com/pytorch/pytorch/issues/157209

Edenzzzz:
Added a note that currently must reload model to use multi-lora

Edenzzzz:
@SolitaryThinker All tests passed. Let's merge this?

fastvideo/utils/collect_env.py
```diff
@@ -80,7 +80,6 @@
 DEFAULT_PIP_PATTERNS = {
     "torch",
     "numpy",
-    "mypy",
```
SolitaryThinker, line 83: Please revert this, then feel free to directly merge.

PR #552, opened 2025-06-25
[CI] Fix CI checks

Edenzzzz:
Now draft PRs won't trigger anything, and correctly restrict precision tests to the paths


Edenzzzz:
Still need to use a public branch

PR #553, opened 2025-06-25
[CI] Fix CI checks

Edenzzzz:
Now draft PRs won't trigger anything, and correctly restrict precision tests to the paths

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -158,7 +158,7 @@ jobs:
   encoder-test:
     needs: change-filter
     if: >-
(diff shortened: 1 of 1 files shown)


Edenzzzz:
cc @SolitaryThinker

SolitaryThinker:
Doesn't the condition in `change-filter` https://github.com/hao-ai-lab/FastVideo/blob/15fddf81f34268683eff735e0a1d6d48d2f514be/.github/workflows/pr-test.yml#L86 already prevent ci tests from being run on draft PRs as all the tests depends on `change-filter`

Edenzzzz:
ok then let's remove all of them?

SolitaryThinker:
yes

PR #555, opened 2025-06-26
[Training] Fixes SP for training; Improve Datasets and schema 

SolitaryThinker:
Changes in this PR:
- fix Training with SP
- pass parquet schema to into datasets to make sure we only extract correct tensor and metadata fields 
- Add full example scripts for crush-smol dataset. T2v is runnable and working (includes i2v scripts which currently are not runnable)
- Remove prompt embed mask from schema fields
- further improvements to training pipeline structure


BrianChen1129 (approved):

BrianChen1129:
VAE test failed?

SolitaryThinker:
we probably need to do a git bisect for vae test, it failed a long time ago and we didn't catch it I think. Our CI filter didn't run the test.

fastvideo/v1/training/training_pipeline.py
```diff
@@ -521,25 +604,24 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
         sampling_param = SamplingParam.from_pretrained(training_args.model_path)
 
         # Set deterministic seed for validation
-        validation_seed = training_args.seed if training_args.seed is not None else 42
-        torch.manual_seed(validation_seed)
-        torch.cuda.manual_seed_all(validation_seed)
-
-        logger.info("Using validation seed: %s", validation_seed)
+        set_random_seed(self.seed)
+        logger.info("Using validation seed: %s", self.seed)
```
Edenzzzz, line 609: Do we have to do this in every validation step? I think if we set the seed when starting training, the whole sequence later will be deterministic?

SolitaryThinker: good point

PR #557, opened 2025-06-27
[CI] fix vae and ssim tests

SolitaryThinker:
Turns on the different in hunyuan vae was due to the upgrade from torch 2.6 to 2.7.1 in the following commit:

d71a4ebffc2034922fc379568b6a6aa722f3744c

cc @Edenzzzz


Edenzzzz:
@SolitaryThinker Looks like the transformer tests are still failing. Please make sure all CIs pass before merging

PR #566, opened 2025-06-28
Fix clip config

Edenzzzz:
There seems to be some bugs associated with pytest to cause one worker to always lag behind, and therefore [hangs when loading weights from distributed.](https://github.com/hao-ai-lab/FastVideo/actions/runs/15940892905/job/44968804148) Everything runs fine with `python examples/inference/basic/basic.py`. You can reproduce by inserting a `dist.breakpoint` anywhere in the weight loading code.
With cpu offload only one GPU is sufficient.

### DIFF
--- a/fastvideo/v1/configs/models/encoders/clip.py
+++ b/fastvideo/v1/configs/models/encoders/clip.py
@@ -62,6 +62,13 @@ class CLIPVisionArchConfig(ImageEncoderArchConfig):
     attention_dropout: float = 0.0
     initializer_range: float = 0.02
     initializer_factor: float = 1.0
+    stacked_params_mapping: List[Tuple[str, str,
+                                       str]] = field(default_factory=lambda: [
+                                           # (param_name, shard_name, shard_id)
+                                           ("qkv_proj", "q_proj", "q"),
+                                           ("qkv_proj", "k_proj", "k"),
+                                           ("qkv_proj", "v_proj", "v"),
+                                       ])
 
 
 @dataclass


SolitaryThinker:
The reason we use 2 GPUs is also to verify that multi-gpu generation does not break. I feel tests are much more stable on modal+buildkite and since modal tests are ready (just merged) let's first see how stable it is before making changes or fixes?

Edenzzzz:
ok sure, but I think it's more of an issue with pytest. Or we can disable distributed loading with pytest

SolitaryThinker:
On modal it seems to run quite well? https://buildkite.com/fastvideo/pr-test/builds/305

SolitaryThinker:
need to rebase if you want to try buildkite

Edenzzzz:
We can skip this if it runs on Modal. Also, why tp_size must equal sp_size? With encoder offload I think we almost never need tp

SolitaryThinker:
Currently we don't have the distributed logic to broadcast the encoder's results to all SP ranks.

Edenzzzz:
> Currently we don't have the distributed logic to broadcast the encoder's results to all SP ranks.

Doesn't each rank load its own encoder?

SolitaryThinker:
You are right, without TP is the easiest to support. I was thinking of the case where TP>1 but TP != SP and that overcomplicates it. 
The current TP==SP also cause the SP size to be restricted unnecessarily by the num_attn_heads of a pipeline's encoder.

PR #571, opened 2025-06-29
[Revert] "[Feature] Load weights from distributed"

SolitaryThinker:
reverts #568 and #470


SolitaryThinker:
training test fails due to bad threshold, other tests pass in buildkite

PR #577, opened 2025-06-29
fully_shard usage on RMSNorm

weifengpy:
Hi from pytorch distributed, thanks for asking questions for fully_shard https://github.com/pytorch/pytorch/issues/157209

want to help how fully_shard should be applied if we decide to hack forward_native, or even better, use avoid hacking forward_native


weifengpy:
cc @Edenzzzz

Edenzzzz (approved):
Thanks!  `model.layer_norm(input)` works

PR #578, opened 2025-06-29
[CI] Fix pre-commit CI

Edenzzzz:
Formatting issue from https://github.com/hao-ai-lab/FastVideo/pull/577
cc @BrianChen1129 @SolitaryThinker 
Also now that buildkite is up, we should stop pushing to the public repo and run CIs for 3rd-party contributors.


SolitaryThinker:
you need to add `go` label for buildkite CI to run

SolitaryThinker (approved):

PR #584, opened 2025-06-30
[Feat] activation checkpointing

Eigensystem:
#466

### DIFF
--- a/fastvideo/v1/training/activation_checkpoint.py
+++ b/fastvideo/v1/training/activation_checkpoint.py
@@ -1,0 +1,91 @@
+import collections
+from enum import Enum
+from typing import Optional
+
+import torch
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    checkpoint_wrapper)
+
+TRANSFORMER_BLOCK_NAMES = [
+    "blocks",
+    "double_blocks",
+    "single_blocks",
+    "transformer_blocks",
+    "temporal_transformer_blocks",
+    "transformer_double_blocks",
+    "transformer_single_blocks",
+]
+
+
+class CheckpointType(str, Enum):
+    FULL = "full"
+    OPS = "ops"
+    BLOCK_SKIP = "block_skip"
+
+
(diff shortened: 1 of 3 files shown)


fastvideo/v1/training/training_pipeline.py
```diff
@@ -83,6 +85,10 @@ def initialize_training_pipeline(self, training_args: TrainingArgs):
         self.transformer.requires_grad_(True)
         self.transformer.train()
 
+        if training_args.gradient_checkpointing:
+            self.transformer = apply_activation_checkpointing(
```
SolitaryThinker, line 89: Could you make this configurable instead of defaulting to `full`?

fastvideo/v1/training/activation_checkpoint.py
```diff
@@ -0,0 +1,91 @@
...
+    "temporal_transformer_blocks",
+    "transformer_double_blocks",
+    "transformer_single_blocks",
+]
+
+
+class CheckpointType(str, Enum):
+    FULL = "full"
+    OPS = "ops"
+    BLOCK_SKIP = "block_skip"
```
SolitaryThinker, line 23: Are all three working?

SolitaryThinker (commented):
Thanks for the effort! How much memory savings do you observe?

fastvideo/v1/training/activation_checkpoint.py
Eigensystem: Yes all work.

Eigensystem:
> Thanks for the effort! How much memory savings do you observe?

Wan 1.3B, 4* A100, sp=4, fsdp=4


activation checkpoint | type | Memory | step time
-- | -- | -- | --
False | None | 25.989Gi/80.000Gi | 4.3s
True | block skip | 14.858Gi/80.000Gi | 5.6s
True | full | 14.858Gi/80.000Gi | 5.8s
True | ops | 25.117Gi/80.000Gi | 8.7s

PR #585, opened 2025-06-30
[Training] Use inference pipeline for training validation

SolitaryThinker:
**Before this PR**
Training pipeline was using a separate ValidationPipeline that did not include text encoders and image encoders. This meant that preprocessing code had to also preprocess these for the validation dataset. It introduced bugs and was hard to debug as it was different from inference pipeline.

**After this PR**
Instead of using a separate ValidationPipeline, we instead directly use the same inference pipeline for training's validation. This means that preprocessing will no longer deal with validation samples. 

Instead of  `--validation_preprocessed_path "$VALIDATION_DIR"`,
 we will directly pass   `--validation_dataset_file "$VALIDATION_DATASET"` into training pipeline


fastvideo/v1/pipelines/stages/denoising.py
```diff
@@ -299,9 +299,9 @@ def forward(
         if st_attn_available and self.attn_backend == SlidingTileAttentionBackend and fastvideo_args.STA_mode == STA_Mode.STA_SEARCHING:
             self.save_sta_search_results(batch)
 
-        if fastvideo_args.use_cpu_offload:
-            self.transformer.to('cpu')
-            torch.cuda.empty_cache()
+        # if fastvideo_args.use_cpu_offload:
+        #     self.transformer.to('cpu')
+        #     torch.cuda.empty_cache()
```
Edenzzzz: Can just remove this, FSDP handles it

fastvideo/v1/training/training_pipeline.py
```diff
@@ -617,45 +599,40 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
...
+                                                       num_inference_steps)
 
                 assert batch.prompt is not None and isinstance(
                     batch.prompt, str)
                 step_captions.append(batch.prompt)
 
                 # Run validation inference
-                with torch.no_grad(), torch.autocast("cuda",
-                                                     dtype=torch.bfloat16):
+                with torch.no_grad():
```
Edenzzzz: Remove this

fastvideo/v1/training/training_pipeline.py
```diff
@@ -716,8 +689,9 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
                     # Other sp_group leaders send their results to global rank 0
                     world_group.send_object(step_videos, dst=0)
                     world_group.send_object(step_captions, dst=0)
-
+        world_group.barrier()
```
Edenzzzz: Is this required?

Edenzzzz (changes requested):

Edenzzzz:
Otherwise LGTM

fastvideo/v1/training/training_pipeline.py
SolitaryThinker: yes, it will sometimes hang here without the barrier

SolitaryThinker:
going to wait for ci to pass again just to make sure the hanging fix in validation is consistent

SolitaryThinker:
I can never locally replicate the hanging that occurs on modal sometimes... It seems that I is hanging before some ranks reach training altogether so I'm going to currently assume that this is an issue specific to modal

Edenzzzz:
Hanging happens when you have mismatching collectives on different ranks, such as mismatching send & recv in pipeline parallel. Some ranks reaching a collective faster shouldn't be an issue

PR #586, opened 2025-06-30
naive video gen working on apple silicon

RandNMR73:
generated video: https://github.com/user-attachments/assets/42b1e1fe-e618-4c69-85f8-19aa1f9d1d0a
setup: Macbook M3 Air, 16 GB RAM
peak memory usage during inference (with SPDA attention backend):  15.84 GB
total time: 15 m 26 s (roughly 8 m of generation time)


fastvideo/v1/layers/layernorm.py
```diff
@@ -35,7 +36,15 @@ def __init__(
...
-        self.weight = torch.ones(hidden_size)
+        # Get the target device from the current platform
+        if current_platform.is_cuda_alike():
+            device = torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif current_platform.is_mps():
+            device = torch.device("mps")
+        else:
+            device = torch.device("cpu")
+            
+        self.weight = torch.ones(hidden_size, device=device, dtype=dtype)
```
SolitaryThinker, line 47: does using 'cpu' device not work for mps backend? Does torch treat it differently? I think vllm just use cpu for apple silicon

fastvideo/v1/distributed/parallel_state.py
```diff
@@ -186,21 +186,33 @@ def __init__(
...
+                )
+            else:
+                # For MPS and CPU, use the base device communicator
+                from fastvideo.v1.distributed.device_communicators.base_device_communicator import DeviceCommunicatorBase
+                self.device_communicator = DeviceCommunicatorBase(
+                    cpu_group=self.cpu_group,
+                    device=self.device,
+                    device_group=self.device_group,
+                    unique_name=self.unique_name,
+                )
```
SolitaryThinker: we should just add a cpu platform backend like the Cuda one. Same with communicator

fastvideo/v1/models/encoders/t5.py
```diff
@@ -326,7 +327,7 @@ def forward(
                 bs, 1, 1,
                 -1) if attention_mask.ndim == 2 else attention_mask.unsqueeze(1)
             attn_bias.masked_fill_(attention_mask == 0,
-                                   torch.finfo(q.dtype).min)
+                                   -1e4)
```
SolitaryThinker, line 330: I just realized this change is probably why fp16 worked on mac. And also why currently fp16 doesn't work on Cuda... Thanks lol

SolitaryThinker: cc @JerryZhou54

fastvideo/v1/layers/layernorm.py
SolitaryThinker: https://docs.vllm.ai/en/latest/getting_started/installation/cpu.html#apple-silicon_1

fastvideo/v1/models/loader/component_loader.py
```diff
@@ -273,17 +273,22 @@ def load_model(self,
...
-                            fsdp_shard_conditions=model._fsdp_shard_conditions,
-                            pin_cpu_memory=fastvideo_args.pin_cpu_memory)
+                # Disable FSDP for MPS as it's not compatible
+                from fastvideo.v1.platforms import current_platform
+                if current_platform.is_mps():
+                    logger.info("Disabling FSDP sharding for MPS platform as it's not compatible")
+                else:
+                    mesh = init_device_mesh(
+                        "cuda",
+                        mesh_shape=(1, dist.get_world_size()),
```
SolitaryThinker: instead of checking here, we should check for cpu/mps backend in FastVideoArgs and just force the relevant arguments to the correct setting.

https://github.com/hao-ai-lab/FastVideo/blob/ad16289871866dfe26fbfe2fa7d7456f5400079f/fastvideo/v1/fastvideo_args.py#L84-L85

fastvideo/v1/models/loader/fsdp_load.py
```diff
@@ -88,23 +88,38 @@ def maybe_load_fsdp_model(
...
+            mesh_shape=(hsdp_replicate_dim, hsdp_shard_dim),
+            mesh_dim_names=("replicate", "shard"),
+        )
+        shard_model(model,
+                    cpu_offload=cpu_offload,
+                    reshard_after_forward=True,
+                    mp_policy=mp_policy,
+                    mesh=device_mesh,
+                    fsdp_shard_conditions=model._fsdp_shard_conditions,
+                    pin_cpu_memory=pin_cpu_memory)
```
SolitaryThinker: same here as above

fastvideo/v1/distributed/parallel_state.py
SolitaryThinker: just realized you added a mps backend. I guess now the question is same as above

fastvideo/v1/layers/layernorm.py
RandNMR73: From the link you sent above, it looks like vllm doesn't actually use the apple silicon GPU since they don't support the mps backend (https://github.com/vllm-project/vllm/issues/1441)

PR #598, opened 2025-07-05
[bugfix] [training] fix deadlock in latent datasets and init error in multi-node training 

SolitaryThinker:
Cause of hanging was actually in a different place than validation dataset: `fastvideo/v1/dataset/parquet_dataset_iterable_style.py` (latent parquet datasets using for training data). 

In method `get_parquet_files_and_length`

**before this PR**
Each rank was first checking for existence of a cache, if found it will immediately return the cache's contents. If cache is not found, then rank 0 will create a cache and then block on barrier. After, it will recursively call the same method again with the assumption that cache has been created for all ranks. 

However there is actually a race condition here where rank0 may not see a cache and then end up creating a cache and writing it to disk before other ranks start checking for the cache. This meant that rank 0 is blocked on the barrier while other ranks that found the cache proceeded to return before calling barrier.

**after this PR**
Now the method is organized as follows:

rank 0 will always check for cache and create one if it does not exist. non rank 0 will immediately block on barrier. After rank0 is done, it will call barrier and at this point we know the cache exists. (either already does or rank 0 created it)

Then all ranks will read from cache and return contents

### DIFF
--- a/fastvideo/v1/training/training_pipeline.py
+++ b/fastvideo/v1/training/training_pipeline.py
@@ -29,7 +29,7 @@ from fastvideo.v1.dataset.dataloader.schema import (
 from fastvideo.v1.dataset.validation_dataset import ValidationDataset
 from fastvideo.v1.distributed import (cleanup_dist_env_and_memory,
                                       get_local_torch_device, get_sp_group,
-                                      get_world_group, get_world_rank)
+                                      get_world_group)
 from fastvideo.v1.fastvideo_args import FastVideoArgs, TrainingArgs
 from fastvideo.v1.forward_context import set_forward_context
 from fastvideo.v1.logger import init_logger
@@ -70,10 +70,7 @@ class TrainingPipeline(ComposedPipelineBase, ABC):
         self.validation_dataset_schema = pyarrow_schema_t2v_validation
 
     def initialize_training_pipeline(self, training_args: TrainingArgs):
-        rank = get_world_rank()
-        logger.info("Rank %s: Initializing training pipeline...",
-                    rank,
-                    local_main_process_only=False)
+        logger.info("Initializing training pipeline...")
         self.device = get_local_torch_device()
         self.training_args = training_args
         world_group = get_world_group()
@@ -607,7 +604,9 @@ class TrainingPipeline(ComposedPipelineBase, ABC):
 
         # Prepare validation prompts
         logger.info('rank: %s: fastvideo_args.validation_dataset_file: %s',
-                    self.global_rank, training_args.validation_dataset_file)
+                    self.global_rank,
+                    training_args.validation_dataset_file,
+                    local_main_process_only=False)
         validation_dataset = ValidationDataset(
             training_args.validation_dataset_file)
         validation_dataloader = DataLoader(validation_dataset,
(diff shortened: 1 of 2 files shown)


Edenzzzz:
Is this supposed to fix hangs like in here? https://buildkite.com/fastvideo/pr-test/builds/418#0197dca4-7353-413f-88c5-45ed1813ef56

SolitaryThinker:
yes, still working on the fix

Edenzzzz (commented):
So what was the main reason for the deadlock?

SolitaryThinker:
updated the PR description above with reason and fix

fastvideo/v1/training/training_pipeline.py
```diff
@@ -663,47 +663,56 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
...
+                local_captions = step_captions
+            else:
+                # Other ranks contribute empty data
+                local_videos = []
+                local_captions = []
+
+            # Gather data from all ranks
+            all_videos_gathered = world_group.all_gather_object(local_videos)
+            all_captions_gathered = world_group.all_gather_object(
+                local_captions)
```
Edenzzzz: If only rank 0 logs results, just call gather instead of all-gather

fastvideo/v1/training/training_pipeline.py
```diff
@@ -663,47 +663,56 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
                     frames.append((x * 255).numpy().astype(np.uint8))
                 step_videos.append(frames)
```
Edenzzzz, line 664: Instead of converting to numpy and calling all_gather_object using cpu group, why not gather the GPU tensor using  NVLINK, then convert to numpy?

Edenzzzz:
> updated the PR description above with reason and fix

Took a look, makes sense

fastvideo/v1/training/training_pipeline.py
SolitaryThinker: okay, I'll just revert this back to p2p communication and directly use send/recv tensors instead of send/recv object

SolitaryThinker: Although it is much easier to reason about correctness using a gather instead of p2p

Edenzzzz: Recommend using gather instead of p2p

SolitaryThinker: it's actually not obvious to me that using all_gather and then processing/saving video will be faster than just doing all_gather_object on cpu_group. As for larger validation datasets (say 64 videos) rank0 will have to serially process them before saving

Edenzzzz: nvlink is faster than PCIE

SolitaryThinker: I'm just going to revert everything for now. To truly be efficient we would need to infer tensor shapes as well as how many rows each rank computed for validation and I feel it adds unnecessary complexity with not much gain

SolitaryThinker: The current logic is also quite well tests

SolitaryThinker:
I'm just going to revert everything for now. To truly be efficient we would need to infer tensor shapes as well as how many rows each rank computed for validation and I feel it adds unnecessary complexity with not much gain

PR #599, opened 2025-07-07
[optimization] [training] Use gather in validation instead of p2p comm

SolitaryThinker:
Now we use `gather` for video tensors on nccl and `gather_object` for captions on cpu_group instead of p2p.


fastvideo/v1/training/training_pipeline.py
```diff
@@ -613,98 +653,134 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
...
-                    world_group.send_object(step_videos, dst=0)
-                    world_group.send_object(step_captions, dst=0)
 
-        # Re-enable gradients for training
-        training_args.inference_mode = False
-        transformer.train()
-        gc.collect()
-        torch.cuda.empty_cache()
+            assert batch.prompt is not None and isinstance(batch.prompt, str)
+            step_captions.append(batch.prompt)
```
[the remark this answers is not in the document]
SolitaryThinker: needed for each rank to know how many validation videos it processed. So that the correct empty tensor shape can be built

Edenzzzz:
We should only enable Gemini instead of Copilot

fastvideo/v1/training/training_pipeline.py
```diff
@@ -613,98 +654,135 @@ def _log_validation(self, transformer, training_args, global_step) -> None:
...
+        # Prepare data for gathering - only SP group leaders have valid
+        # data, other ranks have duplicate data and so we send empty data.
+        if self.rank_in_sp_group == 0:
+            # SP group leaders contribute their data
+            local_videos = torch.cat(step_video_tensors, dim=0)
+            local_captions = step_captions
+        else:
+            # Other ranks contribute empty data
+            local_videos = torch.zeros(final_video_shape, device=self.device)
+            local_captions = []
```
Edenzzzz, line 726: might be better to create a process group containing rank 0 in each sp group

SolitaryThinker: I agree, will add this. I did some testing on multi-node and this current implementation consistently hangs

SolitaryThinker:
closing as this PR doesn't work

PR #606, opened 2025-07-09
Remove all unnecessary torch.cuda.empty_cache

Edenzzzz:
`empty_cache` is super expensive because it scans all the blocks in torch's cache allocator to defragment small blocks. (See https://discuss.pytorch.org/t/gpu-ram-fragmentation-diagnostics/34073/6) When the allocation pattern is stable with no memory leaks, it's not helpful and slows things down a lot.
The better approach is to write code without memory leaks and use offload/reduce seqlen.

### DIFF
--- a/fastvideo/v1/distributed/parallel_state.py
+++ b/fastvideo/v1/distributed/parallel_state.py
@@ -23,7 +23,6 @@ If you only need to use the distributed environment without model parallelism,
  you can skip the model parallel initialization and destruction steps.
 """
 import contextlib
-import gc
 import os
 import pickle
 import weakref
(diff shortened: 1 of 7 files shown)


fastvideo/v1/distributed/parallel_state.py
```diff
@@ -1016,15 +1015,6 @@ def cleanup_dist_env_and_memory(shutdown_ray: bool = False):
...
         ray.shutdown()
-    gc.collect()
-    from fastvideo.v1.platforms import current_platform
-    if not current_platform.is_cpu():
-        torch.cuda.empty_cache()
-    try:
-        torch._C._host_emptyCache()
-    except AttributeError:
-        logger.warning(
-            "torch._C._host_emptyCache() only available in Pytorch >=2.5")
```
Edenzzzz, line 1027: All memory is released automatically on program exit

SolitaryThinker (approved):

PR #616, opened 2025-07-14
[Refactor][Preprocessing] introduce workflow

Eigensystem:
Still have the gpu memory leak issue


uv.lock
SolitaryThinker, line 1: delete this please

fastvideo/v1/workflow/workflow_base.py
```diff
@@ -1,28 +1,72 @@
...
-from fastvideo.v1.fastvideo_args import FastVideoArgs
+from fastvideo.v1.fastvideo_args import ExecutionMode, FastVideoArgs
+from fastvideo.v1.logger import init_logger
 from fastvideo.v1.pipelines import ComposedPipelineBase, build_pipeline
+from fastvideo.v1.pipelines.pipeline_registry import PipelineType
+
+logger = init_logger(__name__)
 
 
 class WorkflowBase(ABC):
```
SolitaryThinker, line 12: Could you add some doc strings describing the reason for a workflow class? Thanks

fastvideo/v1/workflow/preprocess/preprocess_workflow_t2v.py
```diff
@@ -0,0 +1,71 @@
...
+class PreprocessWorkflowT2V(PreprocessWorkflow):
+    training_dataloader: "DataLoader"
+    validation_dataloader: "DataLoader"
+    preprocess_pipeline: "ComposedPipelineBase"
+    processed_dataset_saver: "ParquetDatasetSaver"
+    video_forward_batch_builder: "VideoForwardBatchBuilder"
+
+    def register_components(self) -> None:
+        assert self.fastvideo_args.preprocess_config is not None
+        super().register_components()
```
SolitaryThinker, line 29: Generally I want to avoid relying on a `super()` call to make sure a pipeline/workflow has correct behavior, especially for the components used by a workflow. For example in i2v I think it is okay to redefine all the components. What do you think? I'm open to either way

Eigensystem: Yes I agree.

Eigensystem:
move to #638

PR #627, opened 2025-07-22
Fix lora train steps 

Edenzzzz:


BrianChen1129:
Could you also change ```--num_latent_t``` to match ```--num_frames```? ```num_frames```should be equal to ```4 * (num_latent_t-1) + 1```. I recommend set 20 and 77 respectively as ```num_latent_t``` should also be divided by sp_size

Edenzzzz:
> Could you also change `--num_latent_t` to match `--num_frames`? `num_frames`should be equal to `4 * (num_latent_t-1) + 1`. I recommend set 20 and 77 respectively as `num_latent_t` should also be divided by sp_size

Done

BrianChen1129 (approved):

PR #630, opened 2025-07-22
[Bugfix] Fix LoRA trainable params and training ckpt loading

Edenzzzz:
Previously Validation pipeline init mistakenly re-enabled grads for all params

### DIFF
--- a/fastvideo/v1/pipelines/composed_pipeline_base.py
+++ b/fastvideo/v1/pipelines/composed_pipeline_base.py
@@ -71,14 +71,17 @@ class ComposedPipelineBase(ABC):
         # Load modules directly in initialization
         logger.info("Loading pipeline modules...")
         self.modules = self.load_modules(fastvideo_args, loaded_modules)
+
+    def set_trainable(self) -> None:
         # Only train DiT
-        for name, module in self.modules.items():
-            if not isinstance(module, torch.nn.Module):
-                continue
-            if name == "transformer":
-                module.requires_grad_(True)
-            else:
-                module.requires_grad_(False)
+        if getattr(self.fastvideo_args, "training_mode", False):
+            for name, module in self.modules.items():
+                if not isinstance(module, torch.nn.Module):
+                    continue
+                if name == "transformer":
+                    module.requires_grad_(True)
+                else:
+                    module.requires_grad_(False)
 
     def post_init(self) -> None:
         assert self.fastvideo_args is not None, "fastvideo_args must be set"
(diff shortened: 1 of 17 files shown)


Edenzzzz:
@BrianChen1129 Do we wanna set `num_latent_t` to 20 for `examples/training/finetune/wan_t2v_1_3b/crush_smol/finetune_t2v.sh` too?

BrianChen1129:
> @BrianChen1129 Do we wanna set `num_latent_t` to 20 for `examples/training/finetune/wan_t2v_1_3b/crush_smol/finetune_t2v.sh` too?

Yeah, we can set all ```--num_latent_t``` to 20 and add comment "--num_latent_t should be divided by sp_size

Edenzzzz:
> Yeah, we can set all `--num_latent_t` to 20 and add comment "--num_latent_t should be divided by sp_size

What's finetrainer's value for this?

BrianChen1129:
> > Yeah, we can set all `--num_latent_t` to 20 and add comment "--num_latent_t should be divided by sp_size
> 
> What's finetrainer's value for this?

It's default value is 21 for Wan, we can set to 21 and add comments for user

Edenzzzz:
@BrianChen1129 Let's merge this first? I will submit another PR for lora checkpoint loading. The last commit passed all CIs

fastvideo/v1/pipelines/composed_pipeline_base.py
```diff
@@ -71,14 +71,17 @@ def __init__(self,
...
+    def set_trainable(self) -> None:
         # Only train DiT
-        for name, module in self.modules.items():
-            if not isinstance(module, torch.nn.Module):
-                continue
-            if name == "transformer":
-                module.requires_grad_(True)
-            else:
-                module.requires_grad_(False)
+        if getattr(self.fastvideo_args, "training_mode", False):
```
SolitaryThinker, line 77: why not just directly use `self.fastvideo_args.training_mode`? training_mode is defined as a property and so will always exist

Edenzzzz:
The tests can actually pass, but SSIM times out and LoRA tests sometimes hit extremely volatile runtimes in a build

SolitaryThinker:
I'll do a git bisect on the ssim issue, but the lora runtime seems a bit strange. Perhaps this is caused by storage speed. For now maybe we can just raise the threshold?

Edenzzzz:
@BrianChen1129 See lora ckpt loading example here `examples/inference/lora/wan_lora_inference_from_ckpt.py`

PR #638, opened 2025-07-24
[1/3][Preprocess] refactor preprocessing configs

Eigensystem:


fastvideo/v1/fastvideo_args.py
```diff
@@ -35,6 +79,12 @@ class FastVideoArgs:
     # Model and path configuration (for convenience)
     model_path: str
 
+    # Running mode
+    mode: ExecutionMode = ExecutionMode.INFERENCE
```
SolitaryThinker: we should replace all the current usages of `inference_mode` and `training_mode` with this enum. Perhaps change `inference_mode` to a property as well? This way we directly set which mode through the enum

Eigensystem: I think maybe we can use the `inference_mode` for now, and then standardize on the mode after we port the training to workflow to avoid duplicating changes?

Eigensystem: Because in the current design, `mode` is a variable specified by the user through cli args to indicate the current working mode, while inference_mode changes during runtime based on whether the current pipeline is used for training or inference (validation). When the training pipeline is migrated to a workflow-based approach, different pipelines will be used separately for training and inference without needing to change the user-specified `mode`. Therefore, I think it's best to keep inference_mode unchanged for now.

SolitaryThinker (approved):

PR #664, opened 2025-07-29
[Feat] Improve SP to support any resolution.

jzhang38:
Before this PR, we shard latents in their temporal (number of latent frames) dimension before feed them to DiT. This requires the number of latent frames to be multiple of SP size.
After this PR, we shard tensors inside DiT (after pachification). So long the dit_sequence_length is multiple if SP size. It should work.


zhanglonghao1992:
Does it work now？

PR #688, opened 2025-08-02
[Feat] Add Wan2.2 14B MoE

JerryZhou54:


fastvideo/pipelines/composed_pipeline_base.py
```diff
@@ -271,12 +278,9 @@ def load_modules(
                 logger.info("Using module %s already provided", module_name)
                 modules[module_name] = loaded_modules[module_name]
                 continue
-            if 'transformer' in module_name:
-                loading_module_name = module_name.split("_")[-1]
```
JerryZhou54, line 275: What's the reason for adding this piece of code? It will cause errors when loading "transformer_2"

BrianChen1129: This is used for distill to load different transformer_xxx

BrianChen1129: @SolitaryThinker Could you help reorganize this part? (We could also add the logic to load different models for different transformers, but that's not urgent.)

SolitaryThinker (approved):

PR #692, opened 2025-08-03
[Fix] training pipeline pin_cpu_memory issue

Eigensystem:
Solved `RuntimeError: CUDA error: invalid argument` issue when running training pipeline

### DIFF
--- a/fastvideo/training/wan_distillation_pipeline.py
+++ b/fastvideo/training/wan_distillation_pipeline.py
@@ -41,8 +41,6 @@ class WanDistillationPipeline(DistillationPipeline):
         args_copy = deepcopy(training_args)
 
         args_copy.inference_mode = True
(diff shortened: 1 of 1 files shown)


BrianChen1129:
Could you also change the distillation_pipelines?

Edenzzzz (changes requested):
Add an explanation for this in `fastvideo_args.py`, and report to pytorch?

SolitaryThinker:
What's the issue? does pin memory need to be disabled for trainingpipelines?

Edenzzzz:
Some times pytorch hits cuda error for no obvious reasons

SolitaryThinker:
context: 

> The issue has been reported and solved by Wenxuan. I think it will be solved after we upgrade to the next torch version. https://github.com/pytorch/pytorch/pull/157147

PR #735, opened 2025-08-12
Fix vsa backward gQ

jzhang38:
Fix race condition when reducing gQ gradient on HBM

### DIFF
--- a/csrc/attn/vsa/block_sparse_h100.cu
+++ b/csrc/attn/vsa/block_sparse_h100.cu
@@ -578,6 +578,7 @@ void bwd_attend_ker(const __grid_constant__ bwd_globals<D> g) {
         if (threadIdx.x / 32 == 0) {
             coord<qg_tile> tile_idx = {blockIdx.z, blockIdx.y, store_qg_block_index, 0};
             tma::store_add_async(g.qg, qg_smem, tile_idx);
+            tma::store_async_wait();
         }
     }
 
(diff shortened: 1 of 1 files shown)


BrianChen1129 (approved):

csrc/attn/vsa/block_sparse_h100.cu
```diff
@@ -578,6 +578,7 @@ void bwd_attend_ker(const __grid_constant__ bwd_globals<D> g) {
         if (threadIdx.x / 32 == 0) {
             coord<qg_tile> tile_idx = {blockIdx.z, blockIdx.y, store_qg_block_index, 0};
             tma::store_add_async(g.qg, qg_smem, tile_idx);
+            tma::store_async_wait();
```
Edenzzzz, line 581: Seems a race condition with qk store, but weird that [this line](https://github.com/hao-ai-lab/FastVideo/blob/6ceb4942a0a2bc172a2aded6d3891cc3bf9e9636/csrc/attn/vsa/block_sparse_h100.cu#L571) should've done the job

jzhang38: Yeah this is weird.  Somehow adding extra wait significantly reduce the max error in attn/tests/test_vsa.py.

Can you try to reproduce?

I am going to merge this. We can  revert if you find something else.

PR #736, opened 2025-08-12
[Fix] fix seed in dmd denoising loop

jzhang38:
1. fix seed.
2. remove unnecessary wait in vsa kernel
3. update vsa kernel test from abs max to relative max.


SolitaryThinker:
could you fix pre-commit errors?

SolitaryThinker (approved):

PR #744, opened 2025-08-18
[bugfix] [dmd] Align backward simulation with dmd2 sample back

nappengman:
- Remove redundant iterations
- Set noisy_input to pure noise when target_timestep_idx_int == 0

### DIFF
--- a/fastvideo/training/distillation_pipeline.py
+++ b/fastvideo/training/distillation_pipeline.py
@@ -241,9 +241,6 @@ class DistillationPipeline(TrainingPipeline):
                                             len(self.denoising_step_list), [1],
                                             device=self.device,
                                             dtype=torch.long)
-        # # In https://github.com/tianweiy/DMD2/blob/main/main/sd_unified_model.py, all rank share a same timestep
-        # from accelerate.utils import broadcast
-        # target_timestep_idx = broadcast(target_timestep_idx, from_process=0)
         target_timestep_idx_int = target_timestep_idx.item()
         target_timestep = self.denoising_step_list[target_timestep_idx]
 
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
Hi, thank you for the PR! 

You are completely right that despite the previous PR, there's is still a bug here. 
In our experience, having each rank use a different seed and thus timestep when using DP results in better videos. And when each rank has a different timestep, if each rank performs different number of steps for backward simulation, it results in deadlock in FastVideo (even when SP==1). This is why currently we always perform N-1 steps and index into the correct latent to use.

With the above in mind, I feel we just need to fix/change the following line (take a look at the comment)

fastvideo/training/distillation_pipeline.py
```diff
@@ -258,58 +261,43 @@ def _generator_multi_step_simulation_forward(
...
-        # Step 3: Use the simulated noisy input for the final training step
-        # For timestep index 0, this is pure noise
-        # For timestep index k > 0, this is the result after k denoising steps + noise at target level
-        if noise_latent_index >= 0:
-            assert noise_latent_index < len(
-                self.denoising_step_list
-            ) - 1, "noise_latent_index is out of bounds"
-            noisy_input = noise_latents[noise_latent_index]
-        else:
-            noisy_input = current_noise_latents
```
SolitaryThinker, line 312: `noisy_input` here should be set to a `clone()` of the initial noise from before the above loop. What do you think?

nappengman: Got it, thanks for pointing that out! It makes sense now — setting noisy_input as a clone() of the initial noise can just fix this problem.

PR #794, opened 2025-09-10
[preprocessing] [self-forcing] [2/n] Improve preprocessing and add ode trajectory dataset schema

SolitaryThinker:
Refactors common logic for handling parquet files and schemas into a common file parquet_io.py

- [ ] add tests to CI
- [ ] make sure flushing remaining rows are handled properly


fastvideo/workflow/preprocess/components.py
```diff
@@ -259,147 +245,32 @@ def _process_non_padded_embeddings(
...
+        if self._writer is None:
+            return
+        _ = self._writer.flush(write_remainder=write_remainder)
+        # Reset processed sample count modulo samples_per_file
+        remainder = self.num_processed_samples % self.samples_per_file
+        self.num_processed_samples = 0 if write_remainder else remainder
 
-        del combined_table
-        gc.collect()
+    def flush_last(self, output_dir: str):
```
Eigensystem: call this func in `__del__` method?

SolitaryThinker: I refactored this slightly and now clean_up calls flush_tables with remainder=True. Also __del__ now calls clean_up

PR #808, opened 2025-09-19
[self-forcing] [5/n] Add Self-Forcing distillation pipeline

JerryZhou54:


SolitaryThinker:
I cannot figure out why pre-commit is failing....

PR #811, opened 2025-09-22
[self-forcing] [6/n] Add Ode Init training

SolitaryThinker:


fastvideo/training/ode_causal_pipeline.py
```diff
@@ -0,0 +1,407 @@
...
+                    timestep,
+                    local_main_process_only=False)
+
+        # Prepare inputs for transformer
+        latent_vis_dict["noisy_input"] = noisy_input.permute(
+            0, 2, 1, 3, 4).detach().clone().cpu()
+        latent_vis_dict["x0"] = target_latent.permute(0, 2, 1, 3,
+                                                      4).detach().clone().cpu()
+
+        model_dtype = next(self.transformer.parameters()).dtype
```
JerryZhou54: This should be the actual compute dtype, not master weight dtype

PR #820, opened 2025-09-28
[Feature]Update count trainable param for FSDP2

BrianChen1129:


fastvideo/training/training_utils.py
```diff
@@ -1298,8 +1298,14 @@ def get_scheduler(
                          last_epoch=last_epoch)
 
 
+def _local_numel(p: torch.Tensor) -> int:
+    if hasattr(p, "to_local"):
+        return p.to_local().numel()
```
Edenzzzz, line 1303: Why do we need this? See below comments by gemini

PR #841, opened 2025-10-13
[feat] unified trainer logging

Ohm-Rishabh:


SolitaryThinker:
Overall this PR looks good. Could you please take a look at the failing training tests in CI?
specifically this error: 
```
E       FileNotFoundError: [Errno 2] No such file or directory: 'wandb/latest-run/files/wandb-summary.json'
  |  

```

It is most likely caused by changes to wandb's logging dir, but could you take a look. After that we can merge this PR. Thanks!

PR #853, opened 2025-10-23
Improve FSDP loading with size-based filtering

Ohm-Rishabh:


Edenzzzz (commented):
Could you show some loading and training speed diff?
Also I think you need 
https://github.com/pytorch/pytorch/blob/c7eee495259a5ce2f2f5e8830bcec3b6eca84b31/test/distributed/_composable/fsdp/test_fully_shard_ignore_params.py#L305
to mix FSDP and non-FSDP params, which may add overheads

fastvideo/models/loader/fsdp_load.py
```diff
@@ -215,20 +218,42 @@ def shard_model(
...
+        # Size-based filtering mode
+        min_params = int(os.environ.get("FASTVIDEO_FSDP2_MIN_PARAMS", "10000000"))
+        logger.info("Using size-based filtering with threshold: %.2fM", min_params / 1e6)
+        
+        for n, m in reversed(list(model.named_modules())):
+            if any([shard_condition(n, m) for shard_condition in fsdp_shard_conditions]):
+                # Count all parameters
+                param_count = sum(p.numel() for p in m.parameters(recurse=True))
+
+                logger.info( "Inspecting module: name = %s, type = %s, param_count = %.2fM", n, type(m).__name__, param_count / 1e6 )
```
Edenzzzz: no need to print verbose info

fastvideo/models/loader/fsdp_load.py
```diff
@@ -215,20 +218,42 @@ def shard_model(
...
+                               n, param_count / 1e6, min_params / 1e6)
+                    continue
+                
+                # Shard this module
+                logger.info("Sharding module %s (%.2fM params)", n, param_count / 1e6)
+                fully_shard(m, **fsdp_kwargs)
+                num_layers_sharded += 1
+    else:
+        # Original logic: shard all modules matching conditions
+        logger.info("Using original logic: shard all modules matching conditions")
```
Edenzzzz: no need to print this?

PR #870, opened 2025-11-11
[bugfix] [lora] [CI] Fix LoRA alpha scaling factor & Fix LoRA Inference CI

shaoxiongduan:
When doing inference, if the rank and the alpha are different, the strength of which the lora is applied is wrong because merge_lora_weights in BaseLayerWithLoRA does not scale the whole thing by alpha/rank.

Fixed by:
- Add lora_alpha parameter to BaseLayerWithLoRA.set_lora_weights()
- Store lora_rank and lora_alpha as instance attributes
- Apply alpha/rank scaling when merging LoRA weights
- Extract alpha values from LoRA state dict in pipeline
- Pass alpha values to layer.set_lora_weights() when setting up adapters

Also fixed file naming issues with the LoRA inference CI.

### DIFF
--- a/fastvideo/tests/inference/lora/test_lora_inference_similarity.py
+++ b/fastvideo/tests/inference/lora/test_lora_inference_similarity.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import json
 import os
+import re
 
 import pytest
 
@@ -76,6 +77,13 @@ LORA_CONFIGS = [
 MODEL_TO_PARAMS = {
     "Wan-AI/Wan2.1-T2V-1.3B-Diffusers": WAN_LORA_PARAMS,
 }
+
+def _sanitize_filename_component(name: str) -> str:
+    """Sanitize filename to remove invalid characters (same logic as VideoGenerator)"""
+    sanitized = re.sub(r'[\\/:*?"<>|]', '', name)
+    sanitized = sanitized.strip().strip('.')
+    sanitized = re.sub(r'\s+', ' ', sanitized)
+    return sanitized or "video"
 
 @pytest.mark.parametrize("model_id", list(MODEL_TO_PARAMS.keys()))
 def test_merge_lora_weights(model_id):
@@ -156,14 +164,16 @@ def test_lora_inference_similarity(ATTENTION_BACKEND, model_id):
         generation_kwargs["negative_prompt"] = lora_config["negative_prompt"]
 
         generator.set_lora_adapter(lora_nickname=lora_nickname, lora_path=lora_path)
+        # Sanitize the filename before adding .mp4 extension to match VideoGenerator's behavior
         output_video_name = f"{lora_path.split('/')[-1]}_{prompt[:50]}"
-        generation_kwargs["output_path"] = output_dir
-        generation_kwargs["output_video_name"] = output_video_name
+        output_video_name = _sanitize_filename_component(output_video_name)
+        generated_video_path = os.path.join(output_dir, f"{output_video_name}.mp4")
+        generation_kwargs["output_path"] = generated_video_path
 
         generator.generate_video(prompt, **generation_kwargs)
 
         assert os.path.exists(
-            output_dir), f"Output video was not generated at {output_dir}"
+            generated_video_path), f"Output video was not generated at {generated_video_path}"
 
         reference_folder = os.path.join(script_dir, 'L40S_reference_videos', model_id.split('/')[-1], ATTENTION_BACKEND)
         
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
Hi, could you run pre-commit?
```# Linting, formatting and static type checking
pre-commit install --hook-type pre-commit --hook-type commit-msg

# You can manually run pre-commit with
pre-commit run --all-files

**Comment by shaoxiongduan** [comment]
> Hi, could you run pre-commit?
> 
> ```
> pre-commit install --hook-type pre-commit --hook-type commit-msg
> 
> # You can manually run pre-commit with
> pre-commit run --all-files
> ```

ok! I ran the pre-commit and fixed the ruff issues.

File: fastvideo/layers/lora/linear.py [hunk 1]
```diff
@@ -134,8 +141,16 @@ def merge_lora_weights(self) -> None:
...
-                     @ self.slice_lora_a_weights(self.lora_A).to(data))
+            
+            # Calculate alpha/rank scaling factor
+            if self.lora_alpha is not None and self.lora_rank is not None and self.lora_rank != 0:
+                alpha_scale = self.lora_alpha / self.lora_rank
+            else:
+                alpha_scale = 1.0
+            
+            # Apply LoRA with alpha scaling
+            data += alpha_scale * (self.slice_lora_b_weights(self.lora_B).to(data)
```
Edenzzzz: Please do not multiply with scale if it's one, as it unecessarily launches a kernel.

Edenzzzz (commented):
Thanks for the fix! I was not aware this because no weight in our tests contains the `lora_alpha`-- could you add one to https://github.com/hao-ai-lab/FastVideo/blob/62b95d82878fbf8e05e740fa769acb99de227938/fastvideo/tests/inference/lora/test_lora_inference_similarity.py#L40

fastvideo/layers/lora/linear.py
shaoxiongduan: thanks for pointing this out! I have fixed this

fastvideo/pipelines/lora_pipeline.py
```diff
@@ -183,9 +183,22 @@ def set_lora_adapter(self,
...
             to_merge_params: defaultdict[Hashable,
                                          dict[Any, Any]] = defaultdict(dict)
             for name, weight in lora_state_dict.items():
+                if "lora_alpha" in name:
+                    # Extract alpha value
+                    layer_name = name.replace("diffusion_model.",
+                                              "").replace(".lora_alpha", "")
+                    layer_name, _, _ = lora_param_names_mapping_fn(layer_name)
+                    target_name, _, _ = param_names_mapping_fn(layer_name)
+                    alpha_values[target_name] = weight.item() if weight.numel(
```
Edenzzzz: `target_name` is the translated layer name, please do `self.alpha_values[lora_nickname] = weight.item()` without any mapping

fastvideo/pipelines/lora_pipeline.py
```diff
@@ -183,9 +183,22 @@ def set_lora_adapter(self,
             lora_param_names_mapping_fn = get_param_names_mapping(
                 self.modules["transformer"].lora_param_names_mapping)
 
+            # Extract alpha values and weights in a single pass
+            alpha_values = {}
```
Edenzzzz: could you replace the `lora_alpha` attribute in the pipeline with `alpha_values`?

fastvideo/pipelines/lora_pipeline.py
```diff
@@ -227,9 +240,15 @@ def set_lora_adapter(self,
             lora_B_name = name + ".lora_B"
             if lora_A_name in self.lora_adapters[lora_nickname]\
                 and lora_B_name in self.lora_adapters[lora_nickname]:
+                # Get alpha value for this layer (defaults to None if not present)
+                lora_A = self.lora_adapters[lora_nickname][lora_A_name]
+                lora_B = self.lora_adapters[lora_nickname][lora_B_name]
+                alpha = alpha_values.get(name) if adapter_updated else None
```
Edenzzzz: Can different layers have different alpha values? If so please make `self.alpha_values` a 2D dict

shaoxiongduan:
This change is mainly for a in-progress ticket porting over longcat-video (they used an alpha value that is two times smaller than their rank), I couldn't find any available lora for Wan-AI/Wan2.1-T2V-1.3B-Diffusers that stored the things like this so I currently can't write a test for this :(

fastvideo/pipelines/lora_pipeline.py
shaoxiongduan: I made it so that it stores the alpha values directly in self.lora_adapters.

fastvideo/pipelines/lora_pipeline.py
shaoxiongduan: for this do you mean replacing the self.lora_alpha with self.alpha_values? lora_alpha is used for the training part, I kept it as-is to not make that many changes to training. For the alpha_values part, I'm now storing them in self.lora_adapters. I figured this is more clear and would avoid having two alpha value related instances in self.

fastvideo/pipelines/lora_pipeline.py
shaoxiongduan: yes they are now stored in self.lora_adapters, so it automatically accounts for different alpha values for different layers.

Edenzzzz: sure then it makes sense

fastvideo/layers/lora/linear.py
```diff
@@ -134,8 +141,20 @@ def merge_lora_weights(self) -> None:
             current_device = self.base_layer.weight.data.device
             data = self.base_layer.weight.data.to(
                 get_local_torch_device()).full_tensor()
-            data += (self.slice_lora_b_weights(self.lora_B).to(data)
-                     @ self.slice_lora_a_weights(self.lora_A).to(data))
+
+            # Calculate alpha/rank scaling factor
+            if self.lora_alpha is not None and self.lora_rank is not None and self.lora_rank != 0:
+                alpha_scale = self.lora_alpha / self.lora_rank
```
Edenzzzz: Either follow gemini's approach or do 
```
lora_delta = ....
 if self.lora_alpha and self.lora_rank and self.lora_alpha != self.lora_rank:
       lora_delta *= alpha_scale
data += lora_delta
```

SolitaryThinker:
I've increased the lora inference timeout and going to rerun to see what happens

fastvideo/layers/lora/linear.py
shaoxiongduan: I have changed this. The CI for the lora inference should now pass since https://github.com/hao-ai-lab/FastVideo/pull/881 has been fixed.

Edenzzzz: Please merge main and ensure it passes

shaoxiongduan: ok I merged and also fixed issues regarding the CI testing code (there were filename mismatches). Now the lora inference ci is passing.

PR #871, opened 2025-11-11
[feat] training mfu calculation scripts

Ohm-Rishabh:
I've pushed in the new change. Current performance numbers: https://docs.google.com/document/d/112cSQrSHrgWU0UDJAcRGuFbJB4EFu8QCSB4Q6FTY9y0/edit?usp=sharing

### DIFF
--- a/fastvideo/tests/training/Vanilla/mfu_calculation.py
+++ b/fastvideo/tests/training/Vanilla/mfu_calculation.py
@@ -77,12 +77,13 @@ def run_worker():
         "--validation_steps", "200",
         "--validation_sampling_steps", "50",
         "--validation_guidance_scale", "6.0",
+        #"--enable_torch_compile",
         #"--log_validation",
         "--num_gpus", NUM_GPUS_PER_NODE,
         "--sp_size", NUM_GPUS_PER_NODE,
         "--tp_size", "1",
-        "--hsdp_replicate_dim", "1",
-        "--hsdp_shard_dim", NUM_GPUS_PER_NODE
+        "--hsdp_replicate_dim", NUM_GPUS_PER_NODE,
+        "--hsdp_shard_dim", "1"
     ])
     # Call the main training function
     pipeline = WanTrainingPipeline.from_pretrained(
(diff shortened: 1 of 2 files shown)


fastvideo/tests/training/Vanilla/mfu_calculation.py
```diff
@@ -0,0 +1,195 @@
...
+        num_layers = 24
+
+        
+        # Calculate achieved FLOPs (forward + backward)
+        # Per layer, per batch, forward pass:
+        # - Linear ops (QKV, out_proj, MLP): 12 * hidden_dim^2 * seq_len
+        # - Attention ops (QK^T, Attn*V): 2 * seq_len^2 * hidden_dim
+        # Training multiplier: 3x (1 forward + 2 backward)
+        linear_flops = 12 * hidden_dim * hidden_dim * seq_len
+        attention_flops = 2 * seq_len * seq_len * hidden_dim
```
Edenzzzz: non-causal is 4x, 2 for multply-add and 2 for qk + vo

Ohm-Rishabh: thanks for pointing it out!

zhisbug:
if that's true, then i think this is very bad. Could you double check? And if we confirm things are correct, could you start figuring out where the system overhead is.

One way is to use a 1.3B language model and run training on megatron/deepspeed on the same hardware and see what the MFU is..

Edenzzzz:
You can run torch profiler first to see where bubbles arise in GPUs. For example, there could be high python overheads due to input validation/preprocessing/post-processing pipeline, or try async `save_checkpoint` (overlap with next forward pass).

fastvideo/training/training_pipeline.py
```diff
@@ -656,6 +656,17 @@ def train(self) -> None:
                     "grad_norm": grad_norm,
                     "vsa_sparsity": current_vsa_sparsity,
                 }
+                # Log actual batch shape for MFU calculation
+                if (hasattr(training_batch, 'raw_latent_shape')
+                        and training_batch.raw_latent_shape is not None):
```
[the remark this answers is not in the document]
SolitaryThinker: don't use getattr

SolitaryThinker (approved):

PR #925, opened 2025-12-09
[bugfix] [VSA] Fix block_size computation in backward kernel

Chuge0335:
fix(sparse-attn): correct block_size masking in backward kernel

The backward kernel used an incorrect block_size when applying the mask, causing padding positions to be treated as valid tokens. As a result, p = tl.math.exp2(qk - m) was computed on invalid entries, leading to Inf and NaN values during gradient accumulation (especially in dQ).

This patch fixes the block_size computation for split blocks so that the mask correctly excludes padded regions in all cases.

### DIFF
--- a/csrc/attn/video_sparse_attn/vsa/block_sparse_attn_triton.py
+++ b/csrc/attn/video_sparse_attn/vsa/block_sparse_attn_triton.py
@@ -247,11 +247,12 @@ def _attn_bwd_dq(dq, q, K, V,  #
 
     kv_blocks = tl.load(q2k_num  + meta_base)                 # int32
     kv_ptr    = q2k_index + meta_base * max_kv_blks           # ptr to list
-    block_size = tl.load(variable_block_sizes + q_blk)
     
     
     for blk_idx in range(kv_blocks*2):
-        block_sparse_offset = (tl.load(kv_ptr + blk_idx//2).to(tl.int32)*2 + blk_idx%2) *step_n * stride_tok
+        kv_idx = tl.load(kv_ptr + blk_idx//2).to(tl.int32)
+        block_size = tl.load(variable_block_sizes + kv_idx) -  (blk_idx % 2) * step_n
+        block_sparse_offset = (kv_idx*2 + blk_idx%2) * step_n * stride_tok
         kT = tl.load(kT_ptrs + block_sparse_offset)
         vT = tl.load(vT_ptrs + block_sparse_offset)
         qk = tl.dot(q, kT)


alexzms:
Thanks for the fix! This is a great catch and the fix looks correct to me.  

The original bug was very hard to see because the existing `test_vsa.py` didn’t really stress the difference between q/k block sizes – with `variable_block_sizes` sampled in \[32, 64\], the variance was small so the dq issue barely showed up in the metrics. After changing the test to sample block sizes in \[16, 64\], the discrepancy in dq becomes clearly visible with the old code, and disappears with your patch. Based on both the updated tests and the derivation, I’m confident the new dq implementation is the correct one.

<details>
<summary><strong>Click to expand: Test Output Verification (Before vs. After Fix)</strong></summary>

### Before Fix (High Error in gQ)
```text
Error Analysis for h=16, d=128, mode=backward
======================================================================================================================================================
Config                Blocks    K     gQ Avg        Rel gQ Max    gK Avg        Rel gK Max    gV Avg        Rel gV Max    gO Avg        Rel gO Max  
------------------------------------------------------------------------------------------------------------------------------------------------------
Small sequence        16        2     3.606109e-03 1.043750e+01 2.622788e-04 2.792969e-01 2.795203e-04 1.308594e-01 1.739330e-04 5.786133e-02
Medium sequence       32        4     5.612697e-03 1.456250e+01 2.179964e-04 1.679688e-01 2.284541e-04 8.056641e-02 1.360651e-04 7.666016e-02
Large sequence        53        6     4.896506e-03 1.462500e+01 1.822947e-04 1.523438e-01 1.890964e-04 9.814453e-02 1.131065e-04 4.736328e-02
------------------------------------------------------------------------------------------------------------------------------------------------------
````

### After Fix (Normal Error)

```text
Error Analysis for h=16, d=128, mode=backward
======================================================================================================================================================
Config                Blocks    K     gQ Avg        Rel gQ Max    gK Avg        Rel gK Max    gV Avg        Rel gV Max    gO Avg        Rel gO Max  
------------------------------------------------------------------------------------------------------------------------------------------------------
Small sequence        16        2     3.730352e-04 1.210938e-01 2.652338e-04 1.376953e-01 2.838454e-04 1.279297e-01 1.794014e-04 5.590820e-02
Medium sequence       32        4     2.848884e-04 2.412109e-01 2.170908e-04 1.679688e-01 2.275568e-04 8.056641e-02 1.363263e-04 7.666016e-02
Large sequence        53        6     2.339672e-04 1.972656e-01 1.813933e-04 2.041016e-01 1.879096e-04 9.863281e-02 1.123956e-04 4.760742e-02
------------------------------------------------------------------------------------------------------------------------------------------------------
```

SolitaryThinker:
great! thanks for looking into this @alexzms

SolitaryThinker (approved):
Thanks for the fix!

Issue #926, opened 2025-12-09
[Bug] There is a bug in lora distillation training


zhanglonghao1992:
### Describe the bug

In the train_one_step() func of FastVideo/fastvideo/training/distillation_pipeline.py, I add:
```
for name, param in self.transformer.named_parameters():    
      if param.grad is not None:  
          print(f'req grad: {name}'). 
          print('---------------')
```

and got:

```
req grad: blocks.29._checkpoint_wrapped_module.to_q.lora_A
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_q.lora_B
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_q.base_layer.weight
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_q.base_layer.bias
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_k.lora_A
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_k.lora_B
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_k.base_layer.weight
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_k.base_layer.bias
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_v.lora_A
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_v.lora_B
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_v.base_layer.weight
---------------
req grad: blocks.29._checkpoint_wrapped_module.to_v.base_layer.bias
```

which means that parameters other than lora also have gradients.

### Reproduction

FastVideo/examples/distill/Wan2.2-TI2V-5B-Diffusers/crush_smol/distill_dmd_VSA_t2v_5B_lora.sh

### Environment

None

Edenzzzz:
The problem comes from these two PRs(https://github.com/hao-ai-lab/FastVideo/pull/818, https://github.com/hao-ai-lab/FastVideo/pull/808), which ignored LoRA when enabling gradients. 
[...]

PR #933, opened 2025-12-13
[bugfix] [lora] [distillation] Fix lora distillation bug

SolitaryThinker:
Each inheriting training pipeline can override `trainable_transformer_names: list[str]` so that trainable lora layers are created automatically.

fixes https://github.com/hao-ai-lab/FastVideo/issues/926


SolitaryThinker:
@Edenzzzz  I've confirmed both dmd and dmd+vsa converges with lora, the CI tests are really messed up right now and will fix in a follow up PR.

PR #947, opened 2025-12-22
[docs] small fixes

RandNMR73:


docs/requirements-mkdocs.txt
SolitaryThinker, line 1: the github action requires this to be under `FastVideo/`

PR #950, opened 2025-12-22
feat: magi-sparse-attn

mulinmeng:


SolitaryThinker:
Hi @mulinmeng, thanks for the PR! Could you share more info on the status of this PR and future plans? Do you also plan on adding support for the MAGI model?

PR #974, opened 2025-12-30
[feat] Introduce Cosmos 2.5 Text2World pipeline

KyleShao1016:
- Added new Reason1 text encoder and its configuration.
- Implemented Cosmos 2.5 video generation pipeline with necessary stages.
- Updated model and pipeline registries to include new components.
- Introduced Cosmos 2.5 specific latent preparation and denoising stages.
- Added utility for converting PyTorch checkpoints to safetensors format.

### DIFF
--- a/examples/inference/basic/basic_cosmos2_5_t2w.py
+++ b/examples/inference/basic/basic_cosmos2_5_t2w.py
@@ -1,0 +1,42 @@
+from fastvideo import VideoGenerator
+
+
+def main():
+    # Point this to your local diffusers model dir (or replace with a HF model ID).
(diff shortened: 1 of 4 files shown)


fastvideo/pipelines/stages/latent_preparation.py
```diff
@@ -427,6 +428,210 @@ def retrieve_latents(
 
         return batch
 
+
+class Cosmos25LatentPreparationStage(CosmosLatentPreparationStage):
```
kevin314, line 432: Should we start moving these model specific stage implementations into their own separate files? Same for denoising and text encoding, along with Cosmos 1.0

PR #966, opened 2025-12-27
Layer offloading

Ohm-Rishabh:
This feature implements layer-wise offloading where we asynchronously offload layers and prefetch the next layer as well using the CPU-GPU communication channel.

without layerwise offloading. when we swap transformers during denoising (high to low noise) for wan2.2 we create an overhead of ~25s.
end-to-end run time: 457.98s

with this feature now that we are streaming the layers, the overhead is neglibable at ~2s at swap boundary. 
end-to-end run time: 424.46s

ran all experiments for wan2.2 for height=720, width=1280, num_frames=81, sp=4, fsdp disabled

PR #996, opened 2026-01-07
[feat!] Disable FSDP inference by default

XOR-op:
Originally, fastvideo leaves `use_fsdp_inference` on by default. However, FSDP as a VRAM saving approach can bring performance overhead. In the latest commit with 1GPU, FSDP slows down by 30% for DiT denoising due to the synchronous H2D weight transfer.

This PR 1) disable FSDP in inference examples explicitly with further instruction; 2) change the default value of `use_fsdp_inference` from `True` to `False` as a  BREAKING CHANGE; 3) add more documentations on this option.

This breaking change will only affect inference scripts, since in training mode, this flag will be ignored.

### DIFF
--- a/docs/inference/inference_quick_start.md
+++ b/docs/inference/inference_quick_start.md
@@ -107,7 +107,7 @@ If you encounter CUDA out of memory errors:
 - Try a smaller model or use distilled versions
 - Use `num_gpus` > 1 if multiple GPUs are available
 - Try enabling FSDP inference with `use_fsdp_inference=True` (may slow down generation)
-- Try enableling DiT layerwise offload with `dit_layerwise_offload=True` (now only a few models support this, but may introduce less overhead than FSDP)
+- Try enabling DiT layerwise offload with `dit_layerwise_offload=True` (now only a few models support this, but may introduce less overhead than FSDP)
 
 ### Slow Generation
 


Edenzzzz:
FSDP H2D [is async](https://github.com/pytorch/pytorch/blob/1fe009cc533d0bdfd94b0394e33d120545663499/torch/distributed/fsdp/_fully_shard/_fsdp_param.py#L725), but you may write an explicit transfer controller to reduce the dispatch overhead as in https://github.com/sgl-project/sglang/pull/15511.

SolitaryThinker:
@Ohm-Rishabh already ported that sglang PR to fastvideo here https://github.com/hao-ai-lab/FastVideo/pull/966

This PR should just be cleaning up the default settings and our example scripts. Since similar to sglang, `use_fsdp_inference` and this layerwise offloading are mutually exclusive

XOR-op:
@Edenzzzz In the current implementation, the H2D is on the same CUDA stream of the main one and not prefetched, which exhibits the synchronous behavior even if is invoked as async. That could explain why I hit performance issue.

Edenzzzz:
> @Edenzzzz In the current implementation, the H2D is on the same CUDA stream of the main one and not prefetched, which exhibits the synchronous behavior even if is invoked as async. That could explain why I hit performance issue.

If so that's really poor implementation and worth raising an issue, if you have any profiling results showing they are on the same stream

XOR-op:
> If so that's really poor implementation and worth raising an issue, if you have any profiling results showing they are on the same stream

This is when `use_fsdp_inference` enabled:
image

This is when `use_fsdp_inference` dsiabled and `dit_layerwise_offload` enabled.
image

Note that `dit_layerwise_offload` is only added for Wan models now. It's feasible to add support for other models, but should be in another PR. Also that's only for CPU offloading and not all-gather prefetch. I agree that we should raise an issue when we can build some minimal reproducible code.

XOR-op:
Close this PR for now, since some tests encounter OOM issue. Another PR will be opened which will enable `--dit-layerwise-offload` by default when it is supported for all models. Layerwise offloading has far less overhead compared to torch's FSDP CPU offloading.

PR #1006, opened 2026-01-13
[feat] Hooks API and layerwise offloading for all DiTs

XOR-op:
### Background
Current implementation of layerwise offloading requires manual modification to model implementation, which is prone to errors when we add more models. Instead, inspired by [diffusers](https://github.com/huggingface/diffusers/tree/main/src/diffusers/hooks), we believe a hooks API can be more modular and less intrusive. Besides, we can further incorporate other techniques more easily in the future (e.g. DiT cache).

### Design
Hooks API is implemented by intercepting `nn.Module.forward` and run custom hooks before and after the original hooks. `ModuleHookManager` is a new attribute added to `nn.Module` and manage the installed hooks. Developers should extend `ForwardHook` to implement new hooks.

### Scope
Currently, hooks API is mainly for internal use, and not designed for end users. The primary reason is when we have multiple hooks, they may conflict and the orders should be considered beforehand. Instead, we should provide boolean flags for users coordinate the applying order ourselves. Users may add their own hooks at the risk of breaking other hooks. We may provide a better API for end users if we have more understanding for how users may use it.

### Call For Feedback
I'd like to collect comments and feedbacks on this design. Feel free to leave any question or comment below. Also cc @Ohm-Rishabh

### DIFF
--- a/fastvideo/tests/inference/lora/test_lora_inference_similarity.py
+++ b/fastvideo/tests/inference/lora/test_lora_inference_similarity.py
@@ -93,8 +93,11 @@ def test_merge_lora_weights(model_id):
 
     lora_nickname = lora_config["lora_nickname"]
     lora_path = lora_config["lora_path"]
+    # When layerwise offload is enabled, placeholder tensors cannot be compared directly.
     args = FastVideoArgs.from_kwargs(
         model_path=model_id,
+        dit_layerwise_offload=False,
+        use_fsdp_inference=True,
         dit_cpu_offload=True,
         dit_precision="bf16",
     )


fastvideo/hooks/layerwise_offload.py
```diff
@@ -0,0 +1,122 @@
...
+        self,
+        async_copy_stream: torch.cuda.Stream,
+        next_state: "LayerwiseOffloadState | None" = None,
+    ) -> None:
+        self.async_copy_stream = async_copy_stream
+        self.next_state = next_state
+        self.gpu_named_parameters: dict[str, torch.Tensor] = {}
+        self.cpu_named_parameters: dict[str, torch.Tensor] = {}
+        self.module_ref: nn.Module = None  # type: ignore
+        self.device: torch.device = torch.device("cuda")
```
SolitaryThinker: This will cause issues if on non-cuda device. Is it possible to allow this to handle other devices or at least no-op, with a warning?

XOR-op: Fixed by adding a warning. Layerwise offloading is useless for MPS, and I don't have AMD devices to implement or test, so CUDA is the only supoprted platform for now.

SolitaryThinker (approved):

PR #1030, opened 2026-01-21
[feat] Add GLM-Image inference support

ShreejithSG:
Ported the GLM-Image architecture to FastVideo, enabling high-quality image generation.

Key features and fixes:
- Implemented GlmImageTransformer2DModel in fastvideo/models/dits/glm_image.py.
- Fixed architectural alignment with SGLang (timesteps, prior embeddings, CFG batching).
- Created modular GlmImagePipeline with before_denoising, denoising, and decoding stages!

<h2>Test Plan</h2>
<p>All tests are local (GPU + full <code>zai-org/GLM-Image</code> weights required; they skip cleanly in CI). Run on <strong>GB200, transformers 5.x, diffusers 0.37.x</strong>.</p>
<p><strong>Component parity</strong> (FastVideo component vs official HF <code>transformers</code>/<code>diffusers</code>):</p>
<pre><code class="language-bash">pytest tests/local_tests/glm_image/test_transformer_parity.py -v -s
pytest tests/local_tests/glm_image/test_transformer_strict_load.py -v -s
pytest tests/local_tests/glm_image/test_vae_parity.py -v -s
pytest tests/local_tests/glm_image/test_t5_parity.py -v -s
pytest tests/local_tests/glm_image/test_ar_parity.py -v -s
</code></pre>
<p><strong>Pipeline parity</strong> — full-pipeline pixel parity is ill-posed (stochastic AR + independent latent RNG), so the test injects matched <code>(prompt_embeds, prior_token_ids, latents)</code> into both the diffusers pipeline and the real <code>GlmImageDenoisingStage</code>, then compares denoised latents (cosine + relative MAE) and the decoded image (uint8 MAE):</p>
<pre><code class="language-bash">pytest tests/local_tests/glm_image/test_pipeline_parity.py -v -s       # T2I
pytest tests/local_tests/glm_image/test_edit_pipeline_parity.py -v -s  # I2I / edit
</code></pre>
<p><strong>End-to-end examples</strong> (smoke, real generation):</p>
<pre><code class="language-bash">python examples/inference/basic/basic_glm_image.py   # T2I
python examples/inference/basic/edit_glm_image.py    # I2I / edit
</code></pre>
<h2>Test Results</h2>
<p><strong>Component parity — all non-skip PASS:</strong></p>

Component | Result
-- | --
transformer (DiT) | fp32 cosine 1.000000 (100% within 5e-3); bf16 cosine 0.999969 (rounding over 30 layers vs a different SDPA kernel; RoPE proven exactly equivalent)
transformer strict-load | strict=True state-dict load, 0 unmatched params (loader raises on any miss)
vae (AutoencoderKL) | decode + encode + config, 3/3
text_encoder (T5) + ByT5 tokenizer | tokenizer + glyph extract + embeds, 3/3
vision_language_encoder (GLM-4-9B AR) | surface + to/eval + greedy determinism, 3/3


fastvideo/models/dits/glm_image.py
```diff
@@ -0,0 +1,839 @@
...
+This module implements the diffusion transformer (DiT) decoder for GLM-Image.
+The model takes tokens from the autoregressive vision-language encoder and
+denoises them into high-fidelity images.
+"""
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from diffusers.models.attention import FeedForward
```
SolitaryThinker: Can you use fastvideo's MLP layer instead?

fastvideo/pipelines/basic/glm_image/glm_image_pipeline.py
```diff
@@ -0,0 +1,112 @@
...
+        
+        self.add_stage(
+            stage_name="decoding_stage",
+            stage=GlmImageDecodingStage(
+                vae=self.get_module("vae"),
+                pipeline=self,
+            ),
+        )
+    
+    def __call__(
```
SolitaryThinker: you shouldn't need to override this here?

scripts/glm_image_inference.py
SolitaryThinker, line 1: instead of this example, could you create an example similar to the files in this folder? https://github.com/hao-ai-lab/FastVideo/tree/main/examples/inference/basic

SolitaryThinker:
could you run pre-commit and also rebase to resolve conflicts? thanks

ShreejithSG:
@SolitaryThinker I've rebased and fixed pre-commit issues!

fastvideo/configs/models/base.py
```diff
@@ -51,9 +51,7 @@ def update_model_arch(self, source_model_dict: dict[str, Any]) -> None:
         for key, value in source_model_dict.items():
             if key in valid_fields:
                 setattr(arch_config, key, value)
-            else:
-                raise AttributeError(
-                    f"{type(arch_config).__name__} has no field '{key}'")
+            # Skip unknown fields - HF configs often have extra fields we don't need
```
loaydatrain: why was this removed?

fastvideo/attention/utils/flash_attn_no_pad.py
```diff
@@ -69,12 +76,32 @@ def flash_attn_no_pad_v3(qkv,
...
+            f"Unexpected unpad_input output length: {len(unpad_output_k)}")
+
+    # Handle version differences for value
+    unpad_output_v = unpad_input(rearrange(value, "b s h d -> b s (h d)"),
+                                 key_padding_mask)
+    if len(unpad_output_v) >= 1:
+        value_unpad = unpad_output_v[0]
+    else:
+        raise ValueError(
+            f"Unexpected unpad_input output length: {len(unpad_output_v)}")
```
loaydatrain: from what i see in https://github.com/Dao-AILab/flash-attention/blob/57cef6c2e772602154fc8cefcef80c93ecc4422a/flash_attn/bert_padding.py#L98
unpad_input always returns 5 values? 

either way the if statements can be simplified

fastvideo/pipelines/stages/decoding.py
```diff
@@ -242,6 +242,13 @@ def forward(
                 decoded_frames = self.decode(cur_latent, fastvideo_args)
                 batch.trajectory_decoded.append(decoded_frames.cpu().float())
 
+        # Apply post-decoding hook if configured
+        # This allows pipelines to apply custom transformations after VAE decoding
+        if hasattr(fastvideo_args.pipeline_config, 'post_decoding') and \
+           fastvideo_args.pipeline_config.post_decoding is not None:
+            logger.debug("Applying post_decoding hook")
+            frames = fastvideo_args.pipeline_config.post_decoding(frames)
```
loaydatrain: i think you can use getattr?
either way should we add a default for post_decoding to fastvideo_args? @SolitaryThinker

fastvideo/pipelines/composed_pipeline_base.py
```diff
@@ -329,8 +329,10 @@ def load_modules(
         logger.info("Loading required modules: %s", required_modules)
 
         modules = {}
-        for module_name, (transformers_or_diffusers,
-                          architecture) in model_index.items():
+        for module_name, value in model_index.items():
+            if module_name.startswith("_"):
+                continue
```
loaydatrain, line 416: i think youre adding this to remove the "_name_or_path" entry from the model index, instead you can just pop the entry: https://github.com/hao-ai-lab/FastVideo/blob/main/fastvideo/pipelines/composed_pipeline_base.py#L289-L290

fastvideo/attention/layer.py
```diff
@@ -325,5 +327,12 @@ def forward(
             q = _apply_rotary_emb(q, cos, sin, is_neox_style=False)
             k = _apply_rotary_emb(k, cos, sin, is_neox_style=False)
 
+        if attention_mask is not None:
+            if ctx_attn_metadata is None:
+                ctx_attn_metadata = SDPAMetadata(current_timestep=0,
```
loaydatrain: this is pretty problematic, its better to pass the attn mask as an explicit input and deal with it from there. are you using sdpa because flashattn doesnt support an attn mask?

loaydatrain (changes requested):

loaydatrain:
would the current CI support SSIM tests for this? @SolitaryThinker

examples/inference/basic/basic_glm_image.py
SolitaryThinker, line 1: could you follow other example's structure and API under examples/inference/basic/ for this script?

examples/inference/basic/basic_glm_image.py
```diff
@@ -0,0 +1,75 @@
...
+from fastvideo.pipelines.basic.glm_image import GlmImagePipeline
+from fastvideo.pipelines.pipeline_batch_info import ForwardBatch
+
+OUTPUT_PATH = "image_output"
+
+
+
+def main():
+    # Load the GLM-Image pipeline
+    pipe = GlmImagePipeline.from_pretrained(
```
SolitaryThinker: Create a VideoGenerator Instead

fastvideo/attention/utils/flash_attn_no_pad.py
```diff
@@ -29,8 +29,15 @@ def flash_attn_no_pad(qkv,
...
-        x, key_padding_mask)
+
+    # Handle version differences in flash_attn unpad_input
+    unpad_output = unpad_input(x, key_padding_mask)
+    if len(unpad_output) == 5:
+        x_unpad, indices, cu_seqlens, max_s, used_seqlens_in_batch = unpad_output
+    else:
+        # Older versions return 4 values or different order?
+        # Typically: x_unpad, indices, cu_seqlens, max_s
+        x_unpad, indices, cu_seqlens, max_s = unpad_output[:4]
```
SolitaryThinker: no need to handle these, can assume latest FA version

fastvideo/attention/layer.py
SolitaryThinker: passing attn mask though forward context should be the way to go instead of modifying the arguments for forward. This is one of the main purposes of forward context

fastvideo/attention/layer.py
```diff
@@ -324,6 +325,13 @@ def forward(
             cos, sin = freqs_cis
             q = _apply_rotary_emb(q, cos, sin, is_neox_style=False)
             k = _apply_rotary_emb(k, cos, sin, is_neox_style=False)
+        
+        if attention_mask is not None:
+            if ctx_attn_metadata is None:
+                from fastvideo.attention.backends.sdpa import SDPAMetadata
+                ctx_attn_metadata = SDPAMetadata(current_timestep=0, attn_mask=attention_mask)
+            else:
+                ctx_attn_metadata.attn_mask = attention_mask
```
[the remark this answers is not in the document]
SolitaryThinker: please fix this

fastvideo/models/loader/component_loader.py
```diff
@@ -479,6 +481,85 @@ def load(self, model_path: str, fastvideo_args: FastVideoArgs):
...
+            )
+            logger.info("Loaded processor: %s", processor.__class__.__name__)
+        except Exception as e:
+            logger.warning("Failed to load processor: %s", e)
+            processor = None
+        
+        # Load the model - try GlmImageForConditionalGeneration directly for GLM-Image
+        # Always use device_map="auto" for AR model since it's needed for inference
+        try:
+            from transformers import GlmImageForConditionalGeneration
```
SolitaryThinker: we shouldn't be using pipelines or models from transformers library. These models should be defined in fastvideo similar to the DIT

fastvideo/pipelines/stages/decoding.py
SolitaryThinker: yes it probably makes sense, just havent needed to before

examples/inference/basic/basic_glm_image.py
ShreejithSG: Switched to VideoGenerator API

fastvideo/attention/utils/flash_attn_no_pad.py
ShreejithSG: I have simplified `unpad_input` handling by removing version-specific if statements.

fastvideo/configs/models/base.py
ShreejithSG: Restored

fastvideo/pipelines/stages/decoding.py
ShreejithSG: Now uses: `post_decoding_hook = getattr(fastvideo_args.pipeline_config, 'post_decoding', None)`

fastvideo/pipelines/composed_pipeline_base.py
ShreejithSG: Added explicit `model_index.pop("_name_or_path", None)` and removed the `startswith("_")` check.

Now follows the same pattern as other metadata keys (`_class_name`, `_diffusers_version`) by explicitly popping `_name_or_path` before iterating over `model_index.items()`

examples/inference/basic/basic_glm_image.py
ShreejithSG: Done!

fastvideo/attention/layer.py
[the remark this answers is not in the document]
ShreejithSG: fixed

fastvideo/attention/layer.py
ShreejithSG: I’ve switched it to use set_forward_context() when constructing the metadata and removed the forward arg mutation

fastvideo/pipelines/stages/glm_image_decoding.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+import torch
+from fastvideo.fastvideo_args import FastVideoArgs
+from fastvideo.pipelines.pipeline_batch_info import ForwardBatch
+from fastvideo.pipelines.stages.base import PipelineStage
+from fastvideo.pipelines.stages.validators import VerificationResult
+
+
+class GlmImageDecodingStage(PipelineStage):
```
loaydatrain: is it possible to use the decoding stage already present? or is there some blocker because we're dealing with images here
reference: https://github.com/hao-ai-lab/FastVideo/blob/main/fastvideo/pipelines/stages/decoding.py

fastvideo/attention/layer.py
```diff
@@ -325,5 +327,26 @@ def forward(
...
+        if attention_mask is not None:
+            current_timestep = forward_context.current_timestep
+            forward_batch = forward_context.forward_batch
+            if ctx_attn_metadata is None:
+                # Create new metadata and update forward context
+                new_attn_metadata = SDPAMetadata(
+                    current_timestep=current_timestep,
+                    attn_mask=attention_mask
+                )
+                with set_forward_context(
```
loaydatrain: as discussed on call, can you add the metadata changes in inits or before the transformer blocks

fastvideo/pipelines/stages/glm_image_decoding.py
ShreejithSG: Refactored GlmImageDecodingStage to inherit from DecodingStage (of FastVideo) instead of PipelineStage

fastvideo/attention/layer.py
ShreejithSG: the mask is now set in forward context before calling the transformer

SolitaryThinker (commented):
**Hi @ShreejithSG** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

## Summary
New-model port adding GLM-Image (T2I + an I2I/edit path): DiT, arch/VAE/pipeline configs, a 4-stage composed pipeline, registry wiring, an SSIM test, two examples, and a large local parity suite. **Open PR, currently CONFLICTING/DIRTY.** Reviewed via the `model`-scope rubric (param-mapping coverage, registry dispatch, pipeline composition, SSIM presence). Good news first: across the 29 commits you've already fixed the earlier "critical" flags — CFG pos/neg concat (`before_denoising.py:205`), the VLM-loader returning a single `nn.Module` instead of a tuple, the `VideoGenerator`-based examples, and the VAE `latents_mean`/`latents_std` access. The remaining blockers-to-merge are process/integration, not GLM correctness.

## Verdict
**COMMENT** — no confirmed code BLOCKER, but four MAJORs gate merge.

## Major concerns
- **Needs rebase.** `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`; mergify has asked twice. Rebase onto current `main` and re-push — this review is against the current (stale) diff, so line cites may drift.
- **Repo-wide `transformers` bump to a release candidate.** `pyproject.toml` / `pyproject_other.toml` bump `transformers==4.57.3` → `==5.0.0rc3`. That's a **major-version jump affecting every model**, pinned to an **rc** even though `transformers==5.0.0` final is now on PyPI. CI shows `microscope-transformer-tests` and `microscope-unit-tests` **failing** while `encoder`/`vae`/`kernel` pass — the signature of a transformers API break hitting the most-coupled suites. The bump is genuinely needed (GLM's AR encoder `GlmImageForConditionalGeneration` first ships in `transformers>=5.0.0rc0`), so please: pin to stable `5.0.0` (or justify the rc), fix the cross-model fallout so `fastcheck` is green, and call out the bump + its blast radius in the PR description. (I can't read the external buildkite log to prove root cause — flagging as MAJOR, not BLOCKER.)
- **SSIM test can't run in CI.** `fastvideo/tests/ssim/test_glm_image_similarity.py` is wired correctly with the shared helpers, but skips unless local weights + `transformers>=5.0.0rc0` + CUDA are present, and there are **no seeded references** on `FastVideo/ssim-reference-videos` for `zai-org__GLM-Image`. For a new model the SSIM run is the regression gate — please seed references via the `seed-ssim-references` skill, and add a second entry to `TEST_PROMPTS` (`:45`) since a single-prompt SSIM is one hash from a false green. (loaydatrain raised the same CI question earlier — still open.)
- **No Test Plan / Test Results in the body.** The description is three bullets; the template's "verified SSIM regression tests pass" box is unaddressed and no parity output is shown despite the extensive local suite (`test_pipeline_parity.py`, `test_transformer_parity.py`, `test_transformer_strict_load.py`, …). Please paste the local parity + strict-load output (and the SSIM result once references are seeded).

## Minor
- `fastvideo/models/dits/glm_image.py:600` — `_compile_conditions` is set from `arch_config._fsdp_shard_conditions`. Every other DiT reads the dedicated `_compile_conditions` field. Behaviorally identical today (base `__post_init__` copies fsdp→compile when empty, and both are `[is_blocks]`), so no runtime bug, but it hardcodes the coupling and will diverge silently if a distinct compile condition is ever set. Prefer `arch_config._compile_conditions`.
- `before_denoising.py:162` — AR `generate(do_sample=True)` is hardcoded and the `ar_*` sampling knobs were dropped from `GlmImageConfig`; determinism rides on the seed being re-set each call (it is). Fine for a first port — just noting it's no longer configurable.
- `tests/local_tests/glm_image/README.md` — references a `PORT_STATUS.md` not in the PR and embeds machine-specific absolute paths (`/home/hal-kaiqin/...`).

Mapping/registry/pipeline composition all check out: `param_names_mapping` targets match `MLP.fc_in/fc_out`, the loader **raises** on unmatched params (so a mapping gap fails loudly, not silently), and `test_transformer_strict_load.py` proves completeness — though that proof is gated behind local weights and doesn't run in CI. Nice, clean port overall; the path forward is "land it green," not a redesign.

SolitaryThinker (commented):
**Hi @ShreejithSG** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

## Summary
Differential re-review at head `5f075364` (I last commented on `469d2a8e`). **Good news: 3 of the 4 MAJORs from that review are resolved** — rebased (MERGEABLE now), `transformers` is on stable `>=5.0.0` with a justifying comment, and the `_compile_conditions`/README-path MINORs are fixed. The GLM port itself is clean: pipeline composes the 5 stages via the registry (`WorkloadType.T2I`), `param_names_mapping` load-safety is structural (loader raises on unmatched params + `test_transformer_strict_load.py` asserts completeness with `strict=True`), and no merge-conflict markers leaked. Remaining gates are process/coverage, not GLM correctness.

## Verdict
**COMMENT** — no model-scope BLOCKER. Two MAJORs still gate merge, plus a couple of carries.

## Major concerns
- **CI red — `fastcheck` fails on `microscope-unit-tests`** (build #3524, a real ~8 min failure). Note the failing suite *flipped* since the last rebase: `dreamverse-app-tests` now passes, but unit-tests now fails. That suite runs `pytest fastvideo/tests/{api,contract,dataset,workflow,entrypoints,train}` — and it was **green** at the prior head; the only GLM change since is a 4-line I2I-only edit that doesn't run there. Most likely cause: `transformers>=5.0.0` is an **open floor** (no upper bound) while `accelerate==1.0.1` is exact-pinned, so CI now resolves to transformers 5.10.x (cf. the "fix fir transformer 5.10.2" commit) and breaks a non-GLM unit test. I can't read the Modal log to name the test — please check it and either cap the floor (e.g. `>=5.0.0,<5.11`) or fix the fallout so `fastcheck` is green. (general/CI scope, not GLM.)
- **No CI parity coverage for the new model + thin body.** The SSIM test (`fastvideo/tests/ssim/test_glm_image_similarity.py`) is wired correctly (`min_acceptable_ssim=0.98`, shared harness, now with a full-quality variant) but still can't run in CI — no references seeded on `FastVideo/ssim-reference-videos` for `zai-org__GLM-Image`, and it's still **single-prompt** (the two param dicts are quality variants of one prompt). Please seed references via the `seed-ssim-references` skill and add a 2nd prompt. Separately, the PR body is still three bullets with no Test Plan/Results and the "verified SSIM regression tests pass" box unaddressed, despite strong local parity numbers in the README — please paste the local parity + strict-load output into the body.

## Minor
- `examples/inference/basic/basic_glm_image.py` — @SolitaryThinker's request on this head to follow the other examples' structure/API (peers like `basic_flux2.py` use a docstring + `argparse` + the `fastvideo.api` config objects) is still unaddressed.
- `tests/local_tests/glm_image/README.md:16,56,72,81,129,141` — stale: still says `transformers==5.0.0rc3` (shipped pin is `>=5.0.0`) and lists I2I as "planned follow-up" though the PR ships the full I2I/edit path. Also: the `diffusers` floor bumped `0.33.1 → 0.38.0` — fold that into the dependency note with the transformers bump.
- `fastvideo/pipelines/basic/glm_image/stages/denoising.py:32` — `getattr(batch, "guidance_scale", 1.5)`: `ForwardBatch.guidance_scale` always exists, so the `1.5` fallback is dead and would silently disagree with the field default if it fired. Read `batch.guidance_scale` directly.
- `fastvideo/pipelines/basic/glm_image/stages/before_denoising.py:162` — AR `generate(do_sample=True)` is hardcoded (the `ar_*` knobs aren't on `GlmImageConfig`); determinism rides on the per-call seed reset. Fine for a first port — just confirm `do_sample=True` is the intended final choice.
- No `tests/local_tests/pipelines/` smoke test for the new pipeline (cf. `test_ltx2_pipeline_smoke.py`, as Copilot noted) — a construct-and-short-forward test would guard stage wiring.

The I2I `pooler_output` getattr in the tip commit is a reasonable defensive adaptation to the transformers-5.10.x return-type change, and it's gated to the I2I path so the T2I/SSIM path is unaffected. Nice, clean port — the path forward is "land it green," not a redesign.

H1yori233:
Thanks @SolitaryThinker this round addresses everything actionable, and pushes back on two items I think are misattributed. Rebased onto current `main` (two merges, no conflicts beyond the `transformers` line). PR body now has a full **Test Plan / Test Results** section.

### Fixed
- **`denoising.py:32`** — now reads `batch.guidance_scale` directly.
- **`basic_glm_image.py`** — restructured to the peer styles.
- **`edit_glm_image.py`** — converted to the same typed-API structure for consistency.
- **`README.md`** — refactored; removed the stale `transformers==5.0.0rc3` pin references (now `>=5.0.0`), the `I2I (planned follow-up)` line (I2I ships), folded in the `diffusers>=0.38.0` floor.

### `before_denoising.py:162` `do_sample=True` — confirmed intended
This is byte-for-byte the official diffusers reference (`pipeline_glm_image.py:439-442`), which also hardcodes `generate(..., do_sample=True)` with a `torch.manual_seed(...)` reset immediately before `generate()` and **no** `temperature`/`top_p`/`top_k` (uses the model's `generation_config`). Our port matches exactly — the seed reset at the top of `forward()` makes it deterministic, and dropping the `ar_*` knobs is intentional parity with the reference, not an omission.

### CI red — not introduced by this PR
The only dependency delta vs `main` is the `transformers` floor (`>=4.57.3` → `>=5.0.0`); `accelerate` and `diffusers` are byte-identical to `main`.

### SSIM — references are seeded; single-prompt matches every sibling
- **References exist.** `FastVideo/ssim-reference-videos` already has: https://huggingface.co/datasets/FastVideo/ssim-reference-videos/tree/main/reference_videos/default/L40S_reference_videos/zai-org__GLM-Image/TORCH_SDPA. So the gate is live on L40S.
- **Single-prompt is the convention.** All 14 sibling SSIM tests use exactly one prompt per workload. The two GLM param dicts are a low-res / full-quality pair of the *same* prompt, consistent with peers. Adding a 2nd prompt would make GLM the only multi-prompt model and require a new per-device reference seeded for it.

image
image

SolitaryThinker:
Rebased onto current main (`fd270f15` -> `ee623198c`) to clear the merge conflict — no content changes. Union-resolved mechanically: `fastvideo/configs/models/dits/__init__.py` (`__all__`, main's ordering + `GlmImageDiTConfig`) and `pyproject.toml`/`pyproject_other.toml` (kept main's `tokenizers>=0.20.1,<0.23` pin + this PR's `transformers>=5.0.0` floor). `fastvideo/registry.py`, `fastvideo/models/registry.py`, `configs/models/vaes/__init__.py`, and `component_loader.py` auto-merged. Two intermediate commits became empty (absorbed by later cleanups) and were dropped; intermediate `fastvideo/attention/layer.py` edits that the branch itself later reverted were resolved to main's version throughout. No adaptation commits were needed. Range-diff verified: net delta vs main is identical to the pre-rebase delta.

SolitaryThinker (approved):
Approving per review: all items from the 2026-01-25 request-changes are fixed or moot at head ee623198c (none of the flagged files are touched anymore; SDPAMetadata replaced the attn-mask mutation; all 36 threads resolved), both June pushbacks verified correct, CI 25/25 green.

PR #1036, opened 2026-01-23
Added Sequence Parallelism for LTX-2 Distilled

shaoxiongduan:
Add sequence parallelism (SP) support for LTX2 video generation
Enables distributing video generation across multiple GPUs by sharding the sequence dimension

### DIFF
--- a/fastvideo/models/dits/ltx2.py
+++ b/fastvideo/models/dits/ltx2.py
@@ -2277,6 +2277,12 @@ class LTX2Transformer3DModel(CachableDiT):
             causal_fix=True,
         ).to(hidden_states.dtype)
 
+        # Pad positions to match padded sequence length for SP cross-attention
+        if sp_world_size > 1 and padded_seq_len > video_original_seq_len:
+            padding_needed = padded_seq_len - video_original_seq_len
+            last_pos = positions[:, :, -1:, :].expand(-1, -1, padding_needed, -1)
+            positions = torch.cat([positions, last_pos], dim=2)
+
         video_modality = Modality(
             enabled=True,
             latent=latents,


Davids048:
Is there plan to enable sp on vae?

shaoxiongduan:
yes, theres some issues with the dit sp though, I'm working on solving that before implementing the vae sp

fastvideo/attention/layer.py
```diff
@@ -117,12 +124,38 @@ def forward(
...
+        elif ltx_freqs_cis is not None and ltx_rope_apply_fn is not None:
+            cos, sin = ltx_freqs_cis
+            heads_per_rank = num_heads // world_size
+            head_start = local_rank * heads_per_rank
+            head_end = head_start + heads_per_rank
+            # Slice to this rank's heads: [B, H/SP, T, D]
+            cos_local = cos[:, head_start:head_end, :, :]
+            sin_local = sin[:, head_start:head_end, :, :]
+            # Trim to valid sequence length if needed
+            if valid_seq_len is not None:
```
loaydatrain: this is already taken care of (line 124 above)

fastvideo/attention/layer.py
```diff
@@ -117,12 +124,38 @@ def forward(
...
+                cos_local = cos_local[:, :, :valid_seq_len, :]
+                sin_local = sin_local[:, :, :valid_seq_len, :]
+
+            q_part = qkv[:batch_size]
+            k_part = qkv[batch_size:batch_size * 2]
+            # Transpose to [B, H/SP, T, D] for LTX RoPE application
+            q_part = q_part.transpose(1, 2)
+            k_part = k_part.transpose(1, 2)
+            q_part = ltx_rope_apply_fn(q_part, (cos_local, sin_local))
```
loaydatrain: can we apply rope on q and k together? basically instead of having q_part and k_part we'd keep it as qk_part. i think you'd need to modify the rope_apply_fn slightly

fastvideo/models/dits/ltx2.py
```diff
@@ -365,19 +384,30 @@ def get_patch_grid_bounds(
...
+            device: Device to create tensors on
+            start_frame: Starting frame offset for SP (each rank computes RoPE
+                        with global frame positions, not local positions)
+        """
+        # Apply start_frame offset for SP
+        frame_offset = start_frame + self.shift
         start_timings = self._get_audio_latent_time_in_sec(
-            self.shift,
-            output_shape.frames + self.shift,
+            frame_offset,
```
loaydatrain: this change can be reverted? since we are calculating rope on the whole seq now, start_frame = 0 always

fastvideo/models/dits/ltx2.py
loaydatrain: this can also be handled inside the attn forward to match with LTXDistributedSelfAttention

loaydatrain, line 528: this function and the _apply_ltx_interleaved/split fns can be removed too?

shaoxiongduan: The rope functions are shared utils for the text encoder and the dit, removing them here and moving them to rotary_embedding.py would be also kind of invasive for a specific model's rope utils. I think keeping all the rope utils for ltx2 in the dit is fine.

PR #1048, opened 2026-01-29
[misc] upgrade torch to 2.10

SolitaryThinker:
https://github.com/pytorch/pytorch/issues/166122


SolitaryThinker:
need to also update and rebuild docker images

PR #1058, opened 2026-02-03
[feat] Add Waypoint-1-Small interactive world model support

Satvikmatta18:
## Summary
Adds support for Overworld's Waypoint-1-Small, a 2.3B parameter interactive world model for real-time video generation conditioned on text and controller inputs.

## Changes
- **Transformer**: `fastvideo/models/dits/waypoint_transformer.py` - Full DiT implementation with causal attention, GQA, and control conditioning
- **Config**: `fastvideo/configs/models/dits/waypoint_transformer.py` - Architecture config
- **Pipeline**: `fastvideo/pipelines/basic/waypoint/` - Streaming pipeline with `CtrlInput` support
- **Pipeline Config**: `fastvideo/configs/pipelines/waypoint.py` - Pipeline configuration
- **Sampling**: `fastvideo/configs/sample/waypoint.py` - Default sampling parameters
- **Tests**: Parity tests for transformer and pipeline smoke tests
- **Example**: `examples/inference/basic/basic_waypoint_streaming.py`
- **Docs**: Pipeline README

## Testing
- All transformer parity tests pass (weight loading + forward pass)
- Pipeline smoke tests pass

## References
- Model: https://huggingface.co/Overworld/Waypoint-1-Small
- Official repo: https://github.com/Overworldai/world_engine


examples/inference/basic/basic_waypoint_streaming.py
SolitaryThinker, line 1: Hi, could you model the basic example off of other pipelines? You can create a custom HF repo and upload the converted DiT weight using diffusers format to organize the HF repo. Similar to https://huggingface.co/FastVideo/FastWan2.1-T2V-1.3B-Diffusers. This way the model weight can be automaticall downloaded by VideoGenerator and there's no need to manually download and create configs.

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,732 @@
...
+    def __init__(self, n_heads: int):
+        super().__init__()
+        self.weight = nn.Parameter(torch.ones(n_heads, n_heads))
+    
+    def forward(self) -> torch.Tensor:
+        """Return sigmoid of the gate weights."""
+        return torch.sigmoid(self.weight)
+
+
+class GatedSelfAttention(nn.Module):
```
SolitaryThinker: Instead of redefining attention, could you directly use `LocalAttention` (for cross attn and conditioning) and `DistributedAttention` (for full self-attn) from fastvideo's attention layers?

SolitaryThinker (changes requested):

examples/inference/basic/basic_waypoint_streaming.py
Satvikmatta18: Hi @SolitaryThinker, 

I Updated examples/inference/basic/basic_waypoint_streaming.py to match other FastVideo pipelines by using StreamingVideoGenerator.from_pretrained(...) (with a WAYPOINT_MODEL_ID override), so users don’t need to manually download weights or create configs.

Also created a Diffusers-style Hugging Face repo containing a FastVideo-compatible model_index.json for automatic loading:
https://huggingface.co/smatta/Waypoint-1-Small-Diffusers

fastvideo/models/dits/waypoint_transformer.py
Satvikmatta18: @SolitaryThinker 

Updated fastvideo/models/dits/waypoint_transformer.py to use FastVideo’s attention layers directly: DistributedAttention for full self-attention and LocalAttention for cross-attention/conditioning. Also handled the model’s GQA setup by expanding K/V heads to match Q heads before calling DistributedAttention, and added the required forward context in the pipeline/tests.

examples/inference/basic/basic_waypoint_streaming.py
SolitaryThinker: are the model weights also uploaded? I only see a model_index.json in the HF repo?

Eigensystem: Maybe we should fork/move the repo to fastvideo's hf account? @SolitaryThinker

Eigensystem: Hi @Satvikmatta18. Did you launch the inference script successfully? It seems that the model_index.json in your repo is an empty file.

SolitaryThinker: yes

Satvikmatta18: Hi @SolitaryThinker  @Eigensystem, 
I’ve updated the HF repo smatta/Waypoint-1-Small-Diffusers with all weights (transformer, VAE, text_encoder, tokenizer) and a proper model_index.json, and inference runs successfully. I’ll switch the default model ID in the code to FastVideo/Waypoint-1-Small-Diffusers once the repo is under the FastVideo org.
Could you add me to the FastVideo Hugging Face org so I can move the repo there?

Satvikmatta18: I moved the hf repo to fast video and updated the code!

Eigensystem (commented):
plz fix pre-commit errors

Eigensystem:
Hi @Satvikmatta18, I tried to generate a video but the results were all blurry. Did you check the correctness / compare it with diffusers results?
Cursor 2026-02-12 00 30 45

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,725 @@
...
+        x = ada_gate(self.mlp(ada_rmsnorm(x, s1, b1)), g1) + x
+        
+        return x
+
+
+# =============================================================================
+# Main Transformer Stack
+# =============================================================================
+
+class WaypointTransformer(nn.Module):
```
Eigensystem, line 654: Maybe inherit from class `CachableDiT`

Satvikmatta18: Hi @Eigensystem I’ve made some changes and re-ran comparisons with the official checkpoint and diffusers (same prompt/seed); things look correct on my side. Can you try again and see if it looks better now?

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,725 @@
...
+def ada_gate(x: torch.Tensor, gate: torch.Tensor) -> torch.Tensor:
+    """Apply gating to output."""
+    return x * gate.unsqueeze(1)
+
+
+# =============================================================================
+# Basic Building Blocks
+# =============================================================================
+
+class MLP(nn.Module):
```
Eigensystem: use MLP in fastvideo/layers/mlp.py here?

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,725 @@
...
+    def __init__(self, in_features: int, hidden_features: int, out_features: int, bias: bool = False):
+        super().__init__()
+        self.fc1 = nn.Linear(in_features, hidden_features, bias=bias)
+        self.fc2 = nn.Linear(hidden_features, out_features, bias=bias)
+    
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.fc2(F.silu(self.fc1(x)))
+
+
+class AdaLN(nn.Module):
```
Eigensystem: Could you please define this in `fastvideo/layers/layernorm.py`

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,725 @@
...
+        self.d_model = d_model
+        self.n_heads = n_heads
+        self.n_kv_heads = n_kv_heads
+        self.head_dim = d_model // n_heads
+        self.layer_idx = layer_idx
+        self.causal = causal
+        self.gated_attn = gated_attn
+        
+        # Q, K, V projections
+        self.q_proj = nn.Linear(d_model, d_model, bias=False)
```
Eigensystem: use Linear class in `fastvideo.layers.linear` for all Linear to support feature like quantization

fastvideo/models/dits/waypoint_transformer.py
```diff
@@ -0,0 +1,725 @@
...
+        super().__init__()
+        self.d_model = d_model
+        self.context_dim = context_dim
+        self.head_dim = head_dim
+        # Calculate n_heads from context_dim and head_dim
+        assert context_dim % head_dim == 0, f"context_dim {context_dim} must be divisible by head_dim {head_dim}"
+        self.n_heads = context_dim // head_dim
+        
+        # Q: project from d_model to context_dim
+        self.q_proj = nn.Linear(d_model, context_dim, bias=False)
```
Eigensystem: same

Eigensystem (changes requested):
Hi @Satvikmatta18 . Could you please refer to the google doc to check whether your implementation matches all the requirement in architecture level? Thanks.
https://docs.google.com/document/d/1h7UOPEOsw9BwnHWGJLcm7FOAv4I1is3HWe2iWln6zyY/edit?tab=t.2w8cxbq3lg5x#heading=h.iaae91tz7zew

SolitaryThinker:
could you also address the gemini comments? And resolve them if you have or they do not make sense. Thanks!

fastvideo/models/dits/waypoint_transformer.py
Satvikmatta18: Hi @Eigensystem, addressed all comments and matched the doc rules.

Eigensystem: Hi @Satvikmatta18, did you push the new commits? I can't find the modification.

Eigensystem:
The generated video looks still blurry on my side. Could you write a ssim test in `fastvideo/tests/ssim` to check the correctness? @Satvikmatta18 
CleanShot 2026-02-15 at 19 18 42@2x

Satvikmatta18:
Hi @Eigensystem Eigensystem,
Thanks for the review. I’ve made these changes to address the blurriness:

- flex_attention – Waypoint now uses flex_attention (when available) for the kv_cache=None path, aligned with WanVideo/MatrixGame.
- Sigma schedule – Updated to the official schedule [1.0, 0.861, 0.729, 0.321, 0.0].
- Reproducible noise – Switched to per-frame torch.Generator seeding for consistent results.
- Seed passing – The --seed flag is now passed into reset() for reproducibility.
image

Eigensystem (commented):
plz fix pre-commit errors

Eigensystem (approved):

fastvideo/pipelines/stages/__init__.py
```diff
@@ -86,6 +88,8 @@
     "GameCraftImageVAEEncodingStage",
     "TextEncodingStage",
     "Cosmos25TextEncodingStage",
+    "StepvideoPromptEncodingStage",
```
SolitaryThinker: remove

SolitaryThinker (changes requested):

fastvideo/pipelines/stages/__init__.py
Satvikmatta18: Fixed!

Eigensystem:
Hi @Satvikmatta18. Please run pre-commit locally and fix the errors. Thanks!

SolitaryThinker (commented):
**Hi @Satvikmatta18** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

Reviewed the model-scope port at HEAD `80f59ed` (open PR). This is a careful, mature port — the weight-load contract is exact and well-tested, registry dispatch is correct for both DiT and VAE, and SSIM references are already seeded on HF. One must-fix before merge, then a few minors.

## Major

- **Schema-parity inventory is missing the new pipeline-config fields → this is why CI is red.** `WaypointT2VConfig` (`fastvideo/configs/pipelines/waypoint.py:43-63`) adds 5 new `PipelineConfig` extension fields — `scheduler_sigmas`, `base_fps`, `n_buttons`, `disable_latent_norm`, `bf16_denoise` — that aren't classified in `docs/design/inference_schema_parity_inventory.yaml`. `test_pipeline_config_extension_fields_are_classified` (`fastvideo/tests/api/test_schema_parity_inventory.py:196-206`) asserts the set of extra fields across all `PipelineConfig` subclasses **equals** the inventory's `pipeline_config_extensions` set, so any new unclassified field fails it. That test runs in `microscope-unit-tests` (which is the failing job, and it cascades into the required `fastcheck-passed` gate). The failing test isn't a Waypoint test and the buildkite log is gated, so the cause isn't obvious from the PR page. Fix: add the 5 fields under `pipeline_config_extensions` as `preset_owned` with `sources:` → `WaypointT2VConfig` (mirror the existing `conditioning_strategy`/`state_t` entries — internal preset fields, no serving/CLI `target:`). Verify locally with `pytest fastvideo/tests/api/test_schema_parity_inventory.py -q`. (`is_causal` is already classified; `precision`/`flow_shift`/`vae_*` are inherited and need no change.)

## Minor

- `fastvideo/api/waypoint.py:31` — `WaypointSamplingParam.video_quality` is a new `SamplingParam` extension field (doesn't fail CI today, since that test scans only the base `sampling_param` module). It's consumed generically by `StreamingVideoGenerator`, not Waypoint-specific — consider hoisting it onto base `SamplingParam` (+ inventory entry) to avoid future drift.
- `fastvideo/tests/ssim/test_waypoint_similarity.py:81-83` — single SSIM prompt; a second prompt would harden the regression (single-prompt tests are one hash from false-green). References are seeded, so non-blocking.
- `fastvideo/tests/ssim/test_waypoint_similarity.py:68` runs `height=368` while the example and `WaypointSamplingParam` use `360` — confirm the off-by-8 is intentional padding or align to 360.

## What checks out (verified)

- **`param_names_mapping` covers every HF key.** Pulled all 366 tensor keys from the HF safetensors header, applied the 2-entry mapping (`mlp.fc1→fc_in`, `mlp.fc2→fc_out`), and confirmed the remapped families exactly match the module tree the model builds — the 2-entry mapping is sufficient because every other module is named to match the checkpoint. The `strict=True` parity test backs this.
- **Cross-block `cond_proj` weight-tying** (`waypoint_transformer.py:813-817`) is faithful to the official `world_engine` (`src/model/world_model.py:242-245`), so the tied copies are numerically identical and strict-load is safe.
- **Registry dispatch** correct: DiT key `WorldModel` and VAE key `WorldEngineVAE` both match the HF `_class_name`s; pipeline composed from `ComposedPipelineBase` with no registry bypass.
- **SSIM test** uses shared helpers + a tight 0.95 threshold, and references are seeded on HF (`L40S_reference_videos/Waypoint-1-Small-Diffusers/TORCH_SDPA/…`) matching the test's exact prompt+backend. Suggest running `/test ssim` to confirm the gate holds on L40S before merge.
- Prior review comments (FastVideo attention/Linear/MLP layers, `max_length` from config, `model_index.json`) all addressed; the dedicated multi-frame KV-cache parity test against the official `LayerKVCache.upsert` is a nice touch.

Verdict: comment-only — fix the inventory entry to clear CI, and this looks close.

SolitaryThinker (changes requested):
Full review at head (findings adversarially verified). The port itself is careful — I checked the odd-looking gate_proj slicing, MLPFusion weight chunking, and CondHead weight tying against the official remote code; all match, and the VAE/KV-cache parity tests are real. cc @H1yori233 since you're taking over cleanup.

Blocking:
1. **The red microscope-unit-tests lane is this PR's**: the schema-parity inventory is missing the new `WaypointT2VConfig` fields (`n_buttons`, `disable_latent_norm`, `bf16_denoise`, `scheduler_sigmas`, `base_fps`) and the June merge left main-side entries stale. Classify them in the inventory YAML and rebase (one trivial conflict in `configs/models/dits/__init__.py`).
2. **`waypoint_transformer.py:514`** — the non-flex fallback passes `attention_mask=None`, which `DistributedAttention.forward` doesn't accept (TypeError), and its causal masking diverges from the flex noop path. Fix or delete the branch.
3. **Line 46 flips `torch.backends.cuda.matmul.allow_tf32` globally at import** — every model in the process inherits it, both stacks. Scope it to Waypoint's forward.
4. **The scroll conditioning stream is dropped at every user-facing surface** (`streaming_step` hard-codes zeros, `CtrlInput` is never consumed), and the README example calls `streaming_step(ctrl, fastvideo_args)`, which isn't the real signature.
5. **No conversion script or pinned revision** for `FastVideo/Waypoint-1-Small-Diffusers`, and no single-frame numerical parity test vs the official transformer (current evidence is strict-load + shape/NaN only).

Also worth considering: compose via stages like matrixgame2's streaming pipeline rather than the empty-stage bridge — the repo doesn't need a third streaming pattern.

SolitaryThinker (changes requested):
Follow-up model-port review at exact head `12c99dd42dbabc3ca27e1a26ab460e02f15e989b`.

The previous schema-inventory/merge-conflict, SDPA-fallback, global-TF32, scroll-routing, and empty-stage blockers are resolved. The new single-frame and 160-frame transformer comparisons are real official-reference numerical tests, and the direct checkpoint inventories match (transformer 366/366, VAE 117/117). I still cannot approve the production port for the following exact-head blockers.

## Blocking findings

1. **[P1] Production loading destroys Waypoint's tied conditioning weights.** `WaypointTransformer` aliases all later blocks' six `cond_proj` weights to block 0 at `fastvideo/models/dits/waypoint_transformer.py:689-694`, but the actual transformer loader constructs on `meta` and materializes with `load_state_dict(..., assign=True)` at `fastvideo/models/loader/fsdp_load.py:442-443`. PyTorch replaces each aliased `Parameter` independently under `assign=True`; a reduced repro in the PR environment changes shared identity from true to false. Across 22 blocks, that turns six shared 2560x2560 weights into 132 unique weights: 825,753,600 extra BF16 parameters, about 1.54 GiB. The parity tests call ordinary `load_state_dict`, so they cannot catch this production-only break. Retie after materialization (or make the loader preserve aliases), then assert identity and parameter count through the real `TransformerLoader` path.

2. **[P1] The `TORCH_SDPA` SSIM reference does not necessarily execute TORCH_SDPA.** The model compiles and calls raw FlexAttention directly at `fastvideo/models/dits/waypoint_transformer.py:30-33`, `:490-511`, and `:566-570`; `FASTVIDEO_ATTENTION_BACKEND=TORCH_SDPA` never reaches those calls. On a build with FlexAttention, the test at `fastvideo/tests/ssim/test_waypoint_similarity.py:86-94` therefore stores FlexAttention output under a `TORCH_SDPA` directory; on a build without it, the same label means SDPA. Route the implementation through the FastVideo backend contract, or explicitly reject unsupported backend selections and name/seed the one backend actually used.

3. **[P1] Production pipeline behavior still diverges from the pinned official pipeline, while the tests compare only components.** FastVideo creates a new generator with `seed + frame_index` for every frame at `fastvideo/pipelines/stages/waypoint_stages.py:270-278`. The official source at revision `93fe14eed217bb09c0244a9d91b2b3e88c3de181` advances one continuous `torch.randn` stream, which is also what the 160-frame component test does. In addition, `_seed_image()` hard-codes zero mouse/button/scroll at `waypoint_stages.py:163-181`, while the official pipeline runs controller encoding before the image cache pass and caches the caller's first controls. The pipeline smoke test uses fake modules, and the HF smoke merely writes an unmatched 17-frame gray-image video; neither compares the production FastVideo pipeline. Preserve the official RNG state and seed-image control ordering, then add matched official-vs-production pipeline parity (same prompt, image, controls, seed, dimensions, and frame count) through the public loader.

4. **[P1] The public control/config contract silently changes the requested workload.** `_controls()` truncates keyboard, mouse, and scroll to their minimum time length at `waypoint_stages.py:218-246`; `forward()` then emits that many frames even when `num_frames` requests more. Batch size and channel widths are not validated. Requested height/width and `num_inference_steps` are also ignored by the fixed latent shape and sigma list at `:257-290`. Validate exact batch/time/channel shapes and equality with `num_frames`, and either implement or reject unsupported resolution/step overrides instead of returning misleading metadata.

5. **[P1] The source/weight publication is not reproducible.** Runtime and SSIM load mutable `FastVideo/Waypoint-1-Small-Diffusers`; the remote-code smoke loads mutable `Overworld/Waypoint-1-Small` with `trust_remote_code=True` and no `revision=`. The local parity paths record no resolved source or weight revision, and there is no Waypoint README/PORT_STATUS or conversion/publication manifest. Record and use the inspected official revision `93fe14eed217bb09c0244a9d91b2b3e88c3de181` and converted revision `4e1b1231cd6f9e768bcdc2a331223163e99b887a`; document a reproducible copy/conversion recipe (a converter only if transformation is required), and exercise both production loaders. The VAE production loader is currently non-strict for `WorldEngineVAE`, unlike the direct parity test.

6. **[P1] Upstream code-license provenance is internally conflicting and needs an explicit resolution before merge.** The HF model card advertises Apache-2.0, but the exact [official transformer source](https://huggingface.co/Overworld/Waypoint-1-Small/blob/93fe14eed217bb09c0244a9d91b2b3e88c3de181/transformer/model.py), [official VAE source](https://huggingface.co/Overworld/Waypoint-1-Small/blob/93fe14eed217bb09c0244a9d91b2b3e88c3de181/vae/ae_model.py), and other remote-code files carry GPLv3-or-later headers; the named [official `world_engine` repository](https://github.com/Overworldai/world_engine) is GPL-3.0. This PR's close ports instead declare Apache-2.0 (`waypoint_transformer.py:1`, `world_engine_vae.py:1`). Because the upstream declarations conflict, please obtain/document a compatible upstream grant or provide clean-room provenance; replacing the source header by itself does not resolve that ambiguity.

7. **[P1] Current-head quality evidence is absent and the existing reference predates the inference rewrite.** The Waypoint MP4 was added to `FastVideo/ssim-reference-videos` in dataset commit `0f388c32042c79acfe4776ea9c429077d23cf8f3` on 2026-06-10 and has never changed. Output-affecting commit `c3b315f3cc14cbf0797951716f3608c377cf1214` landed more than a month later. Exact-head Buildkite 4445 ran five fastcheck steps and ignored the SSIM/full-suite steps. After the behavioral/backend fixes above, generate the exact-head L40S candidate, compare it visually against both the old reference and official output, upload only the accepted artifact, then run SSIM/full-suite. Reseeding before those fixes would only baseline the divergences.

## Smaller cleanup

- `tests/local_tests/waypoint/test_waypoint_kv_cache_parity.py:44` imports `tensordict` before `_official_cache()` reaches `pytest.importorskip`, so the documented optional-dependency path hard-fails. In the project environment the focused run was `2 failed, 5 passed, 3 skipped`, with both failures from this ordering.
- `fastvideo/pipelines/basic/waypoint/README.md:49` publishes stale sigma values; use the exact values already present in the config.
- `fastvideo/tests/ssim/test_waypoint_similarity.py:149-151` references undefined `output_path` on its failure path.
- Normalize native 360x640 output geometry: the current 360-to-368 alignment/padding makes writer and non-writer paths differ.

## Evidence checked

- Local API/parser/pipeline-structure run: 26 passed.
- Reduced PyTorch 2.12 `assign=True` alias repro: shared parameter identity changed from true to false.
- GitHub pre-commit and Buildkite fastcheck are green at the reviewed head; neither ran the local official parity suite or SSIM.

Please keep the request-changes state. Once the production loader and matched pipeline parity are fixed, I can re-review the exact new head and its L40S evidence.

PR #1060, opened 2026-02-04
Kandinsky5 lite dit

jaisurya27:
- Add Kandinsky5 DiT config + mapping so diffusers weights load cleanly.
- Implement a FastVideo DiT wrapper around diffusers Kandinsky5Transformer3DModel for parity.
- Register Kandinsky5 DiT in the model registry.
- Add transformer‑only parity test comparing FastVideo vs diffusers with fixed inputs.
- Materialize Kandinsky5 non‑persistent buffers after meta init to avoid meta‑device errors.


SolitaryThinker:
Hi Thanks! Initial progress looks good. Could you directly port the implementation of DiT to FastVideo rather than importing from diffusers? thanks!

PR #1064, opened 2026-02-04
[Model] LTX 2 Base

Davids048:


johndpope:
are there any hints of speed up ? fps?  from my research 5 mins ago - ltx2 doesn't seem to have self forcing / causal  kv caching.... is this true ? would it help if it did? https://github.com/NVlabs/LongLive/blob/a5ea73bed5c244fe3fc432d005b8ac055452b509/wan/modules/causal_model_infinity.py#L554 - i can create a ticket if need be - 

P.S. - this just dropped - https://github.com/Infini-AI-Lab/MonarchRT

PR #1073, opened 2026-02-06
more fix and relocate STA arguments to pipeline config

### LINKED ISSUE #843: [Bug] STA_mode defaults to Inference in examples/inference/sta_mask_search/wan_example.py
### Describe the bug

`STA_mode` becomes `STA_Mode.STA_INFERENCE` inside `from_kwargs` after `cls(**kwargs)` in `fastvideo_args.py`. Even though `STA_mode="STA_searching"` and `STA_mode="STA_tuning"` are passed to `wan_example.py` in two separate calls, both of them defaults to `STA_inference`.

### Reproduction

Simply run `examples/inference/sta_mask_search/inference_wan_sta.sh`

### Environment

Environment created using readme from FastVideo.

Python 3.12.11

zhisbug:
1. relocate the STA_mode arg to fix #843 
2. reduce local process logging warning when it is not distributed
3. minor tweaks on readme

### DIFF
--- a/fastvideo/logger.py
+++ b/fastvideo/logger.py
@@ -102,7 +102,7 @@ def _info(logger: Logger,
         - When both are False, the message will be logged from all processes
         - By default, only logs from processes with LOCAL_RANK=0
     """
-    is_distributed = "LOCAL_RANK" in os.environ or "RANK" in os.environ
+    is_distributed = int(os.environ.get("WORLD_SIZE", 1)) > 1
     try:
         local_rank = int(os.environ["LOCAL_RANK"])
         rank = int(os.environ["RANK"])


fastvideo/logger.py
```diff
@@ -102,6 +102,7 @@ def _info(logger: Logger,
         - When both are False, the message will be logged from all processes
         - By default, only logs from processes with LOCAL_RANK=0
     """
+    is_distributed = "LOCAL_RANK" in os.environ or "RANK" in os.environ
```
Eigensystem: There are LOCAL_RANK & RANK environment variables even if not in distributed setup.
`is_distributed = os.environ.get("WORLD_SIZE", 1) > 1` should be better.

PR #1080, opened 2026-02-08
[feat] Add stable_audio T2A Generation

happy-harvey:


examples/inference/basic/stable_audio_basic.py
SolitaryThinker: could you rename and also structure this file similarly to other basic examples?

PR #1088, opened 2026-02-10
Kandinsky5 lite dit clean

jaisurya27:
Adds native FastVideo implementation of Kandinsky5Transformer3DModel (no Diffusers wrapper).
Adds Kandinsky5 Lite DiT config + parameter mapping and registry wiring.
Adds a local parity test against Diffusers checkpoint.

### DIFF
--- a/fastvideo/configs/models/dits/kandinsky5.py
+++ b/fastvideo/configs/models/dits/kandinsky5.py
@@ -6,22 +6,19 @@ from fastvideo.configs.models.dits.base import DiTArchConfig, DiTConfig
 
 @dataclass
 class Kandinsky5ArchConfig(DiTArchConfig):
-    _fsdp_shard_conditions: list = field(
-        default_factory=lambda: [
-            lambda n, m: ("text_transformer_blocks" in n
-                          or "visual_transformer_blocks" in n)
-            and n.split(".")[-1].isdigit()
-        ])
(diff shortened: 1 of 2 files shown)


fastvideo/models/dits/kandinsky5.py
Eigensystem, line 1: Please use as many layers from `fastvideo/layers` like ReplicatedLinear, RoPE and fused layernorm

jaisurya27: Hi,
I had implemented the FASTVIDEO native version for kandinsky5-lite. Have made sure to include these:
-  Migrated core projections to FastVideo native ReplicatedLinear (time/text/visual embeddings, modulation, attention qkv/out, out layer)
- Switched attention compute to FastVideo LocalAttention and added fallback to torch SDPA when forward context is absent (for standalone parity test path).
-  Kept/used FastVideo fused norms in encoder/decoder blocks via LayerNormScaleShift (eps=1e-5) with float residual math to match official behavior
-  Kandinsky5FeedForward uses FastVideo MLP(..., bias=False, act_type="gelu").
- Implemented native RoPE + fractal flatten/unflatten flow and non-persistent buffer materialization

Eigensystem:
plz solve conflicts and pre-commit errors

Eigensystem (approved):

PR #1093, opened 2026-02-11
[Feat] Native dit implementation for SD3.5

Ishxn20:
This PR adds Stable Diffusion 3.5 Medium support to FastVideo with a FastVideo-native SD3 DiT path and SD3.5-specific pipeline wiring.

### DIFF
--- a/fastvideo/configs/pipelines/sd35.py
+++ b/fastvideo/configs/pipelines/sd35.py
@@ -23,8 +23,7 @@ def _sd35_clip_text_postprocess(outputs: BaseEncoderOutput) -> torch.Tensor:
     if hs is None:
         raise RuntimeError(
             "SD3.5 CLIP prompt embeddings require hidden_states. "
-            "Set output_hidden_states=True for CLIP encoders."
-        )
+            "Set output_hidden_states=True for CLIP encoders.")
     return hs[-2]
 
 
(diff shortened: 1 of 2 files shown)


fastvideo/pipelines/stages/sd35_timestep_preparation.py
Eigensystem, line 1: Hi @Ishxn20. Why we should add a separate file? What's the difference between the general timestep preparation stage and this one?

fastvideo/models/dits/sd3.py
```diff
@@ -2,43 +2,1081 @@
...
+                                            ...] | None = None,
+    ) -> None:
+        super().__init__()
+
+        self.heads = heads
+        self.head_dim = dim_head
+        self.inner_dim = out_dim
+        self.context_pre_only = context_pre_only
+
+        self.to_q = nn.Linear(query_dim, self.inner_dim, bias=True)
```
Eigensystem: Please try to reuse layers in `fastvideo.layers` like `ReplicatedLinear` as much as possible

Ishxn20: Ack

fastvideo/pipelines/stages/sd35_timestep_preparation.py
Ishxn20: @Eigensystem I think you are right that a separate file isn't required. The SD3.5-specific logic is computing a mu parameter for the FlowMatchEulerDiscreteScheduler when use_dynamic_shifting is enabled -- everything else in forward() is same to the base TimestepPreparationStage. I'll make the changes and push again.

PR #1106, opened 2026-02-16
[feat] Add activation offloading to training pipelines

ketakitank:
This PR implements asynchronous activation offloading to reduce VRAM consumption during training. By leveraging PyTorch’s saved_tensors_hooks, intermediate activations are moved to CPU memory during the forward pass and fetched back just-in-time for the backward pass

### DIFF
--- a/fastvideo/fastvideo_args.py
+++ b/fastvideo/fastvideo_args.py
@@ -855,9 +855,10 @@ class TrainingArgs(FastVideoArgs):
     checkpoints_total_limit: int = 0
     resume_from_checkpoint: str = ""  # specify the checkpoint folder to resume from
     activation_offloading: bool = field(
-        default=False, 
-        metadata={"help": "Offload activations to CPU to save VRAM during training."}
-    )
+        default=False,
+        metadata={
+            "help": "Offload activations to CPU to save VRAM during training."
+        })
 
     # optimizer & scheduler
     num_train_epochs: int = 0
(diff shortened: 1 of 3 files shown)


SolitaryThinker:
thanks! could you add some profiling for total memory usage?

jzhang38 (approved):

SolitaryThinker (changes requested):
Requesting changes at exact head `34a576de0b5450cb281f7a8dc7f286f03c13a5f6`.

- **[P1] `fastvideo/training/training_pipeline.py:656`** reads `self.training_args.log_every_n_steps`, but `TrainingArgs` defines and parses no such field. Every legacy training run reaches this after its first step and raises `AttributeError`. Please use an existing validated interval or add and parse this argument, then cover the loop-level behavior with a test.
- **[P1] `fastvideo/pipelines/composed_pipeline_base.py:129-150`** installs offloading inside the loop over `trainable_transformer_modules`, but each iteration scans all pipeline modules. A pipeline with two trainable transformers therefore wraps every selected layer twice, and the scan can also wrap non-trainable module lists. Please install the wrapper exactly once and scope it to the intended trainable layers.
- **[P2] `fastvideo/training/training_pipeline.py:803`** logs tracker metrics with `getattr(self, "current_step", 0)`, but this pipeline never sets `current_step`; all memory points are recorded at step 0. Pass the actual loop step to `_log_memory_stats` instead.
- PyTorch already provides `torch.autograd.graph.save_on_cpu(pin_memory=True)` for this exact saved-tensor lifecycle: https://docs.pytorch.org/docs/stable/autograd.html#torch.autograd.graph.save_on_cpu. The custom hooks copy to a pageable CPU destination, so `non_blocking=True` does not establish the claimed asynchronous transfer. Please prefer the native context manager, or document and test a measured reason for a custom hook.

This change currently wires only the legacy `fastvideo/training/` stack, while new training work belongs in modular `fastvideo/train/`. Please make the legacy-only scope explicit or add modular support in a separately scoped change. Also, a pure rebase preflight stopped at a non-allowlisted conflict in `fastvideo/fastvideo_args.py`; after the implementation is fixed, it will need a deliberate rebase onto current `main` plus functional and peak-memory evidence.

PR #1108, opened 2026-02-17
[bugfix] fix import PreTrainedModel in stepllm.py

dsynkd:
In `transformers==4.57.3` (per `pyproject.toml`), `PreTrainedModel` should directory be imported from `transformers` module. Previously we were importing from `transformers.modeling_utils`, which was causing a crash.


SolitaryThinker (approved):

SolitaryThinker:
thanks!

PR #1109, opened 2026-02-17
Added OpenAI-compatible API server and benchmark script

AjAnubolu:
Ported SGLang benchmark scripts to FastVideo

### DIFF
--- a/fastvideo/tests/modal/pr_test.py
+++ b/fastvideo/tests/modal/pr_test.py
@@ -9,36 +9,41 @@ image_version = os.getenv("IMAGE_VERSION")
 image_tag = f"ghcr.io/hao-ai-lab/fastvideo/fastvideo-dev:{image_version}"
 print(f"Using image: {image_tag}")
 
-image = (
-    modal.Image.from_registry(image_tag, add_python="3.12")
-    .run_commands("rm -rf /FastVideo")
-    .apt_install("cmake", "pkg-config", "build-essential", "curl", "libssl-dev", "ffmpeg")
-    .run_commands("curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable")
-    .run_commands("echo 'source ~/.cargo/env' >> ~/.bashrc")
-    .env({
-        "PATH": "/root/.cargo/bin:$PATH",
-        "BUILDKITE_REPO": os.environ.get("BUILDKITE_REPO", ""),
-        "BUILDKITE_COMMIT": os.environ.get("BUILDKITE_COMMIT", ""),
-        "BUILDKITE_PULL_REQUEST": os.environ.get("BUILDKITE_PULL_REQUEST", ""),
-        "IMAGE_VERSION": os.environ.get("IMAGE_VERSION", ""),
-    })
-)
+image = (modal.Image.from_registry(
+    image_tag, add_python="3.12"
+).run_commands("rm -rf /FastVideo").apt_install(
+    "cmake", "pkg-config", "build-essential", "curl", "libssl-dev", "ffmpeg"
+).run_commands(
+    "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable"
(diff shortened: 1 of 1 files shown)


fastvideo/entrypoints/openai/image_api.py
```diff
@@ -0,0 +1,279 @@
...
+    merge_image_input_list,
+    parse_size,
+    save_image_to_path,
+)
+from fastvideo.logger import init_logger
+
+logger = init_logger(__name__)
+router = APIRouter(prefix="/v1/images", tags=["images"])
+
+OUTPUT_DIR = "outputs/images"
```
XOR-op: Can we make this path configurable?

fastvideo/entrypoints/openai/video_api.py
```diff
@@ -0,0 +1,331 @@
...
+    merge_image_input_list,
+    parse_size,
+    save_image_to_path,
+)
+from fastvideo.logger import init_logger
+
+logger = init_logger(__name__)
+router = APIRouter(prefix="/v1/videos", tags=["videos"])
+
+OUTPUT_DIR = "outputs/videos"
```
XOR-op: Also this path

fastvideo/entrypoints/openai/api_server.py
```diff
@@ -0,0 +1,135 @@
...
+def _parse_args() -> tuple[FastVideoArgs, str, int]:
+    """Parse CLI arguments and return (FastVideoArgs, host, port)"""
+    from fastvideo.configs.sample import SamplingParam
+    from fastvideo.utils import FlexibleArgumentParser
+
+    parser = FlexibleArgumentParser(
+        description="FastVideo OpenAI-compatible API server"
+    )
+    parser.add_argument("--host", type=str, default="0.0.0.0")
+    parser.add_argument("--port", type=int, default=8000)
```
XOR-op: In the `fastvideo/entrypoints/cli/serve.py` we have `subparser_init` which also defines the defaults. Which one is the single source of truth? We can keep either, but not setting default more than once

fastvideo/entrypoints/openai/api_server.py
```diff
@@ -0,0 +1,135 @@
...
+        k: v
+        for k, v in vars(args).items()
+        if k not in ("host", "port", "subparser", "config", "dispatch_function")
+        and v is not None
+    }
+    fastvideo_args = FastVideoArgs.from_kwargs(**cli_kwargs)
+    return fastvideo_args, host, port
+
+
+def run_server(fastvideo_args: FastVideoArgs, host: str = "0.0.0.0", port: int = 8000):
```
XOR-op: Also here

XOR-op (changes requested):
Thanks for your contribution! The overall code is good, except some minor issues mentioned in my comments. Besides those, I'd like to ask for two more changes:

1. When we port code from sglang, could we also declare what file do we refer to? One example is in `fastvideo/entrypoints/cli/serve.py`;
2. Can we also port some tests for API server? Especially because the API server is user-facing and we'd like to know when behaviors will change.

AjAnubolu:
Thanks for the feedback, let me know if this looks good

XOR-op:
The current test is not comprehensive: I expect we run a real server with a model, and we check if the real responses are correct. You can get ideas from [sglang's test](https://github.com/hao-ai-lab/FastVideo/blob/main/fastvideo/tests/inference/lora/test_lora_inference_similarity.py) and [our tests](https://github.com/hao-ai-lab/FastVideo/blob/main/fastvideo/tests/inference/lora/test_lora_inference_similarity.py). We do have GPU CI, so don't worry about the resources. Also remember to keep `num_inference_steps` and the resolution low because we don't want to spend too many compute resources for this test. The other changes look good to me.

XOR-op:
Please use `pre-commit` to pass lint/format checks. Docs here: https://haoailab.com/FastVideo/contributing/overview/

AjAnubolu:
Fixed pre-commit checks, are the CI failures (Training, SSIM, self-forcing) pre-existing?

PR #1114, opened 2026-02-19
[misc] FlashAttention 4 support

XOR-op:
Install from https://github.com/Dao-AILab/flash-attention/tree/main/flash_attn/cute

What's in this PR:
- Integrate into FlashAttnBackend now, given no specific reason to choose fa2 over fa4 when fa4 is available
- Use torch.custom_op for `torch.compile` support


XOR-op:
Converted to WIP: after discussion with @jzhang38, we decide we should add a new FA4 backend instead and wrap it in a custom op

SolitaryThinker:
Do we need to gate fa4 behind cuda arch?

XOR-op:
Added the gate

jzhang38 (approved):

PR #1117, opened 2026-02-20
[Feat] Add Ovis-Image-7B text-to-image pipeline

HenryDzy:
Adds native FastVideo support for [Ovis-Image-7B](https://huggingface.co/AIDC-AI/Ovis-Image-7B)

### New files
**Models & configs**
- `fastvideo/models/dits/ovisimage.py` — Native `OvisImageTransformer2DModel`:
  6 double blocks + 27 single blocks, SwiGLU activations, RoPE, `DistributedAttention`
- `fastvideo/models/encoders/qwen3.py` — `Qwen3Model` text encoder
  (wraps `Ovis2.5-2B` for conditioning)
- `fastvideo/configs/pipelines/ovis_image.py` — `OvisImageT2IConfig`
  (flow_shift=3.0, embedded_cfg_scale=5.0, Qwen3 pre/postprocess hooks)
- `fastvideo/pipelines/basic/ovis_image/` — `OvisImagePipeline`

**Pipeline**
- `fastvideo/pipelines/basic/ovis_image/__init__.py`
- `fastvideo/pipelines/basic/ovis_image/ovis_image_pipeline.py`
- `fastvideo/training/ovis_image_training_pipeline.py`

**Tests**
- `fastvideo/tests/transformers/test_ovisimage.py` — transformer forward pass 
- `fastvideo/tests/encoders/test_qwen3_encoder.py` — HF vs FastVideo Qwen3 parity
- `fastvideo/tests/ssim/test_ovis_image_similarity.py` — MS-SSIM regression test
- `tests/local_tests/pipelines/test_ovis_image_pipeline_smoke.py` — end-to-end `VideoGenerator` smoke test

**Example**
- `examples/inference/basic/basic_ovis_image.py` — runnable example

### Files modified
- `fastvideo/registry.py` — registered `AIDC-AI/Ovis-Image-7B`
- `fastvideo/configs/models/dits/__init__.py` — exported `OvisImageTransformer2DModelConfig`
- `fastvideo/configs/models/encoders/__init__.py` — exported `Qwen3Config`
- `fastvideo/configs/models/vaes/base.py` — added `load_encoder`/`load_decoder` fields
- `fastvideo/models/registry.py` — registered `OvisImageTransformer2DModel`, `Qwen3Model`
- `fastvideo/pipelines/pipeline_registry.py` — registered `OvisImagePipeline`
- `fastvideo/pipelines/stages/denoising.py` — `except (ImportError, RuntimeError)` for Triton guards
- `fastvideo/pipelines/stages/causal_denoising.py` — same fix
- `fastvideo/pipelines/stages/matrixgame_denoising.py` — same fix
- `fastvideo/training/__init__.py` — exported `OvisImageTrainingPipeline`
- `docs/inference/support_matrix.md` — added Ovis-Image-7B row


SolitaryThinker (commented):
**Hi @HenryDzy** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

**Verdict: COMMENT** (model scope, open PR). The add-a-model flow is complete and the weight-load contract — usually the failure point for a new model — is verified sound. No blocker; two majors to resolve before merge, plus minor cleanups.

First, a positive worth stating because it *looks* like the classic bug: the empty `param_names_mapping` is **correct here**, not a silent weight-skip. I derived the DiT's `named_parameters()` from `__init__` and diffed it against the HF `AIDC-AI/Ovis-Image-7B` transformer safetensors index — **583/583 keys match exactly, 0 missing / 0 extra** — and the DiT loader is strict (`fsdp_load.py:362-364` raises on an unmapped checkpoint key; `:405-413` raises on an unloaded model param), so a gap would hard-fail rather than skip. The module names mirror Diffusers 1:1 by design. Same story for the Qwen3 encoder (un-prefixed ckpt keys match the model params; strictness enforced at `component_loader.py:461-466`).

### Major

- **SP-unsafe: RoPE computed at full sequence length but applied to an SP-sharded tensor → breaks at `sp_world_size > 1`.** `fastvideo/models/dits/ovisimage.py:543-556` builds `img_cos/img_sin` (and `joint_cos/joint_sin`) for the full sequence, then shards `img` via `sequence_model_parallel_shard(img, dim=1)`. `_apply_rope` (`:102-111`) then broadcasts full-length cos/sin `[1, full_seq, 1, head_dim]` against `q/k` of shape `[B, seq/sp, heads, head_dim]` — a shape mismatch on every rank when sp>1, in both the double-block (`:246`) and single-block (`:301`) paths. The single-block path also does `cat([txt, img], dim=1)` (`:384`) mixing a full-length `txt` with a sharded `img`, which is logically inconsistent across ranks. `shard()` no-ops at sp=1 (`parallel_state.py:325`), so the SSIM test (`sp_size=1`) and smoke test (`num_gpus=1`) never hit it — but `sp_size` defaults to `num_gpus` (`fastvideo_args.py:765`), so any 2+ GPU run without an explicit `sp_size=1` breaks. Reference DiTs shard RoPE per-rank (`wanvideo.py:673,715-719`) or skip the manual shard and let `DistributedAttention` handle the all-to-all (`flux_2.py`). Fix: shard cos/sin (and `txt`) to match the local segment, or drop the manual shard/all-gather and rely on `DistributedAttention` like Flux2 — then add an `sp_size=2` parity/SSIM run. If multi-GPU SP is out of scope for v1, gate it with `assert get_sp_world_size() == 1` in `__init__` so it fails loud.

- **CI is red on the encoder/unit jobs.** `fastcheck-passed` fails via `microscope-encoder-tests` and `microscope-unit-tests` (exit 1); `microscope-transformer-tests` and `pre-commit` are green. A PR adding a text encoder with the encoder-test job failing is a real signal, and the PR body has no "expected to fail / seeded in follow-up" note. Please post the failing job output (or run the encoder + unit `/test` scopes) and confirm it's unrelated to the new `Qwen3Model` before merge.

### Minor

- `onboarding.md` (repo root, +137) looks like personal dev-scratch (a conda bootstrap, a `PYTHONPATH` hack, branch notes) — please remove it; the local-test setup already lives in `tests/local_tests/ovis_image/README.md`.
- `fastvideo/configs/ovis_image_7b_t2i_pipeline.json` appears unreferenced — the `OvisImageT2IConfig` dataclass is the real config path. Drop it if it's not a documented override, to avoid two sources of truth.
- `fastvideo/models/dits/ovisimage.py:433` hardcodes `_compile_conditions = []`, while the arch config defines the block conditions and every peer DiT wires them through (`flux_2.py:853`, `wanvideo.py:563`). As written, `torch.compile` shards nothing for Ovis — likely unintended.
- `fastvideo/tests/ssim/test_ovis_image_similarity.py` has a single prompt (one hash from a false green); consider adding one more and seeding its reference. The `full_quality` (1024²) tier also has no Ovis reference seeded yet (only the `default`/L40S tier).
- The PR body's file lists are stale (they cite `tests/transformers/test_ovisimage.py`, `tests/encoders/test_qwen3_encoder.py`, and stage edits that aren't in the diff). Tests now live under `tests/local_tests/ovis_image/`; the stages touched are `decoding.py`/`text_encoding.py`/`timestep_preparation.py`.

The local parity-test suite is genuinely strong — transformer/encoder/VAE/end-to-end tests all run through the **production loaders** vs official references with tight tolerances. The main gap is that they're all under `tests/local_tests/` and skipped in CI, so the only CI-visible model test is SSIM (currently blocked by the red CI above). A CI-visible SSIM/parity run on a seeded tier would close that out.

Training (`ovis_image_training_pipeline.py`, the `_compile_conditions`/FSDP interaction) is out of model scope — deferring to the training reviewer.

SolitaryThinker (changes requested):
Full review at head b2751be7c (components + pipeline lanes, findings adversarially verified). The port itself is in good shape: native DiT, TP-layer encoder, real parity evidence through the production loaders (bit-exact timestep schedule, latent-drift gate, documented non-skip passes), SSIM references already seeded on HF, and the HF repo is plain diffusers layout so no conversion script is needed.

**The blocker is the June-10 merge of main, not the port.** The merge resolved badly and this branch now REVERTS main functionality:
- `fastvideo/models/registry.py` lost the model auto-discovery mechanism plus the GameCraft, CLIPTextModelWithProjection, and upsampler registrations — that's what reds your encoder lane (T5 resolves to the hf-wrapper class), and merged it would break GameCraft, SD3.5, StableAudio, Magi, and LTX2 upsampling across inference and both training stacks.
- `training/__init__.py` dropped the `LTX2TrainingPipeline` export.
- The schema-parity inventory is stale (the 3 unit-test failures clear on rebase — main fixed those in #1446, not your bug).

Fix: rebase onto main (branch is CONFLICTING, 93 commits behind), take main's side of `registry.py`/`training/__init__.py`/the inventory yaml, and limit `registry.py` to your two additive Ovis lines.

After that, two port items:
1. **SP > 1 crashes** despite the docstring claim: RoPE cos/sin stay full-length while `img` is sharded, single-stream blocks concat replicated `txt` without the `replicated_*` kwargs, and pre-all-gather padding is never trimmed. Fix or drop the SP claim.
2. **The new `Qwen3Model` duplicates the Flux2 Qwen3 stack** that landed on main after this PR was written (~350 lines). Consolidate onto the existing decoder layers — a config-level delta covers the Ovis chat-template path.

Smaller: add `Qwen3Model` to `EntryClass`; remove the dead duplicate `AutoencoderKL` registry key, root-level `onboarding.md`, and the unreferenced `ovis_image_7b_t2i_pipeline.json`; refresh the PR description (the `denoising.py` changes it describes no longer exist at head).

PR #1127, opened 2026-02-24
migrate uv

Davids048:
- Change doc default to uv 
- Use uv + auto detect cuda architecture to improve fastvideo-kernel installation process.


XOR-op:
LGTM for most of the changes. However I'm uncertain if the change for hardcoded sm90a in fastvideo-kernel will break anything. cc @SolitaryThinker any thought?

Issue #860, opened 2025-10-31
[Bug] Potential Memory Leak in fast_wan Inference Script


Tacossp:
### Describe the bug

The memory usage of the process running the fast_wan inference script increases steadily during execution (e.g., when processing inputs in a loop or over a long period) and is not reclaimed, suggesting a memory leak.The process's Resident Set Size (RSS) continuously increases. On a system with 460G of RAM, the memory usage grows steadily, and the process fails (likely due to OOM) after processing approximately 408 videos. This suggests a consistent memory leak for each video processed.

<img width="333" height="218" alt="Image" src="https://github.com/user-attachments/assets/0d5739a2-b983-4fc8-9d35-58a31c9e7e9e" />

### Reproduction

# Run the generation command
export FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN
export MODEL_BASE=FastVideo/FastWan2.1-T2V-14B-Diffusers/
num_gpus=4
    fastvideo generate \
        --model-path $MODEL_BASE \
        --sp-size $num_gpus \
        --tp-size 1 \
        --num-gpus $num_gpus \
        --height 480 \
        --width 832 \
        --num-frames 77 \
        --num-inference-steps 3 \
        --fps 16 \
        --prompt-txt "$PROMPT_FILE" \
        --batch-start-index $START_INDEX \
        --batch-size $BATCH_SIZE \
        --negative-prompt "Bright tones, overexposed, static, blurred details, subtitles, style, works, paintings, images, static, overall gray, worst quality, low quality, JPEG compression residue, ugly, incomplete, extra fingers, poorly drawn hands, poorly drawn faces, deformed, disfigured, misshapen limbs, fused fingers, still picture, messy background, three legs, many people in the background, walking backwards" \
        --seed 1024 \
        --output-path "$OUTPUT_DIR" \
        --VSA-sparsity 0.9 \
        --dmd-denoising-steps "1000,757,522" \
        --enable_torch_compile

### Environment

4 H20 460G RAM
All package versions were installed strictly according to the official requirements.txt file.

ChaoqiLiang:
Same issue

zhisbug:
I can take a look

PR #1130, opened 2026-02-25
Fix memory leak in batch video generation

AjAnubolu:
## Summary
- Fix OOM when using `--prompt-txt` to batch-generate many videos (#860)
- The batch loop accumulated full video tensors (~1.1GB each) in the `results` list — after ~408 videos on 460GB RAM, this causes OOM
- After each video is saved to disk, now only keep lightweight metadata (prompt, size, generation_time, etc.) instead of heavy tensors (`samples`, `frames`, `trajectory_latents`, `trajectory_decoded`, `audio`)
- Add `gc.collect()`, `torch.cuda.empty_cache()`, and `torch.cuda.reset_peak_memory_stats()` between iterations to reclaim memory

## Test plan
- [x] `python -c "from fastvideo.entrypoints.video_generator import VideoGenerator"` — import succeeds
- [x] `ruff check` — passes
- [x] `yapf --diff` — passes (formatting matches project style)
- [x] Programmatic assertion that all cleanup calls are present in the batch loop
- [ ] Manual: run batch generation with 400+ prompts and confirm RSS stays flat

### DIFF
--- a/fastvideo/entrypoints/video_generator.py
+++ b/fastvideo/entrypoints/video_generator.py
@@ -6,6 +6,7 @@ This module provides a consolidated interface for generating videos using
 diffusion models.
 """
 
+import gc
 import os
 import re
 import threading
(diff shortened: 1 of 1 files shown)


jzhang38:
Thanks for that! Closing because https://github.com/hao-ai-lab/FastVideo/pull/1132 is a proper fix.

PR #1133, opened 2026-02-26
Flux2 port [in progress]

Gnav3852:
In progress flux2 port with debug files

To run debug files:
Install diffusers from source
Have sglang at the same level as Fastvid in the pod


SolitaryThinker:
Flux2 landed on main via #1349 (klein + full-pipeline scaffolding, SSIM refs seeded on the HF dataset), so this draft is superseded — it's 284 commits behind and carries debug compare scripts that shouldn't merge. Closing; if any piece here isn't covered by main's flux_2, please reopen with that delta.

PR #1136, opened 2026-02-26
Port GGUF quantization layer from vLLM to FastVideo

MihirJagtap:
Add GGUF quantization support (GGUFConfig, GGUFLinearMethod, GGUFEmbeddingMethod) adapted from vLLM, using vLLM's GGML CUDA kernels for dequantization. MoE support excluded as FastVideo does not use mixture-of-experts layers.

### DIFF
--- a/examples/inference/optimizations/README.md
+++ b/examples/inference/optimizations/README.md
@@ -1,5 +1,24 @@
 # Optimization Examples
 
+## Attention backend
 ```bash
 python examples/inference/optimizations/attention_example.py
+```
+
+## Text encoder quantization (AbsMaxFP8)
+```bash
+python examples/inference/optimizations/text_encoder_quant_example.py \
+    --text_encoder_path /path/to/umt5_xxl_fp8_e4m3fn_scaled.safetensors
+```
+
+## Text encoder quantization (GGUF)
+
+GGUF quantization uses low-precision weight types (Q4_0, Q4_K, Q8_0, etc.) to
+reduce the text encoder memory footprint while keeping quality close to the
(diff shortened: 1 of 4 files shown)


jzhang38:
Can we also add usage example?

SolitaryThinker:
@MihirJagtap picking this back up from the maintainer side: the GGUF layer is still the only GGUF support in the tree (main has fp8/nvfp4 only) and the code is self-contained — but nothing consumes it end-to-end: no loader/registry path reaches GGUFLinearMethod, so as merged it would be dead code. Are you up for (1) a rebase onto current main (quantization/__init__.py gained fp8/nvfp4 entries, small conflict) and (2) wiring one end-to-end load story — e.g. how a user loads a city96-style GGUF Wan checkpoint through the FastVideo loader? If yes I'll review promptly; if not, we'll close for now and keep the branch as reference for whoever needs GGUF later.

MihirJagtap:
> @MihirJagtap picking this back up from the maintainer side: the GGUF layer is still the only GGUF support in the tree (main has fp8/nvfp4 only) and the code is self-contained — but nothing consumes it end-to-end: no loader/registry path reaches GGUFLinearMethod, so as merged it would be dead code. Are you up for (1) a rebase onto current main (quantization/**init**.py gained fp8/nvfp4 entries, small conflict) and (2) wiring one end-to-end load story — e.g. how a user loads a city96-style GGUF Wan checkpoint through the FastVideo loader? If yes I'll review promptly; if not, we'll close for now and keep the branch as reference for whoever needs GGUF later.

Yes I can take up the point (1) and (2)

PR #1142, opened 2026-02-28
[Refactor] SP Mask --> original seq len; HunyuanVideo 1.5 does not need mask

jzhang38:
* **Simplified SP Padding Logic**: The explicit `attention_mask` parameter for SP has been removed distributed attention layers and various DiT models. It is replaced by an `original_seq_len` parameter, streamlining how sequence padding is handled during attention computations.
* **Remove HunyuanVideo 1.5 encoder_hidden_states padding**:  The plan is to remove padding on the encoder_hidden_states for all model. and then use a unified var_len API.


fastvideo/models/dits/hunyuanvideo15.py
```diff
@@ -789,18 +696,19 @@ def forward(self, x, c, mask=None):
...
-        # Run scaled dot product attention
-        from fastvideo.attention.backends.flash_attn import FlashAttnMetadataBuilder
-        attn_metadata = FlashAttnMetadataBuilder().build(
-            current_timestep=0,
-            attn_mask=mask,
-        )
-        # Run distributed attention
-        with set_forward_context(current_timestep=0, attn_metadata=attn_metadata):
-            attn_output = self.attn(q, k, v)  # [B, L, H, D]
+        if mask == None:
```
Eigensystem: use `if mask is None` here?

Eigensystem (approved):

PR #1148, opened 2026-03-01
[feat] Job Runner UI

dsynkd:
This PR adds a web-based UI that allows running jobs for various purposes. Currently only video inference is supported with the following parameters:

- Inference Steps
- Frames
- Height
- Width
- Guidance Scale
- Seed
- GPUs
- DiT CPU Offload
- Text Encoder CPU Offload
- FSDP Inference

Jobs can be:
- Created
- Started
- Stopped
- Restarted
- Deleted

All models that are registered through FastVideo registry are supported.

Currently the following features are included:
- Console Log
- Download Log
- Inference Timer

To run, navigate to the `ui` directory and run:

```bash
npm i
npm run build
npm start
```

There are two services that start concurrently: the web server and the API server. They can also be started separately using `npm run start:api` and `npm run start:web`. In this case, the `.env.local` file can be used to configure the API server URL (it defaults to `http://localhost:8089`).

The web server can also be run in dev mode using `npm run dev`. This will automatically restart the server upon any changes made to the UI.

Demo video is shown below:

https://github.com/user-attachments/assets/d5ec5cbf-6f57-418b-adba-209e165c423b

### DIFF
--- a/ui/src/components/CreateJobModal.tsx
+++ b/ui/src/components/CreateJobModal.tsx
@@ -353,18 +353,6 @@ export default function CreateJobModal({
             <details className={formStyles.advancedSettings}>
               <summary>Advanced Settings</summary>
               <div className={formStyles.settingsGrid}>
-                <div className={formStyles.formRow}>
-                  <label htmlFor="modal-num-steps">Inference Steps</label>
-                  <Slider
-                    id="modal-num-steps"
-                    min={1}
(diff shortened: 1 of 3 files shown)


dsynkd:
Opened another PR here: https://github.com/hao-ai-lab/FastVideo/pull/1172/

PR #1162, opened 2026-03-09
Consolidate examples and scripts

HaripraveenS:
Consolidated examples and scripts folders. manually checked all changes.
Moved all files to examples folder since it had significantly more files.


loaydatrain:
@SolitaryThinker is there a point to keeping the bash scripts in `examples/inference/cli/` (originally in `scripts/inference/`)? We already have python inference scripts under `examples/inference/basic`

PR #1168, opened 2026-03-13
[DMD] Avoid redundant simulation steps in generator forward

Chuge0335:
Currently `_generator_multi_step_simulation_forward` simulates the whole denoising trajectory regardless of the sampled target timestep, while only the latent at the target timestep is actually used for the final training step.

This PR limits the simulation loop to run only up to the sampled timestep:

```
for step_idx in range(target_timestep_idx_int)
```

instead of running through the full step list.

This avoids redundant student forward passes during the simulation stage.
On average this reduces the number of simulated steps by ~50% when the timestep is uniformly sampled.

No change in training behavior, just removes unnecessary computation.

### DIFF
--- a/fastvideo/training/distillation_pipeline.py
+++ b/fastvideo/training/distillation_pipeline.py
@@ -543,70 +543,74 @@ class DistillationPipeline(TrainingPipeline):
 
         return pred_video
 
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
Thanks for the PR! 

I don't remember the exact setting (ie if SP was enabled) but using this optimization previously caused hanging during backward simulation. Please test carefully to see if this is still an issue

PR #1175, opened 2026-03-22
[feat]: add HunyuanVideo model plugin for fastvideo/train framework

alexzms:
- Add HunyuanVideo model plugin (`fastvideo/train/models/hunyuan/`) that extends the existing WanModel base with Hunyuan-specific forward kwargs, VAE normalization, and attention metadata                                                                                                                                      
  - Add training configs for HunyuanVideo finetuning                                                                                                                                                                                                  

TODO: notice hunyuan vae has a potential bug on tiling encode/decoding, will investigate in the near future.

### DIFF
--- a/fastvideo/train/models/hunyuan/hunyuan.py
+++ b/fastvideo/train/models/hunyuan/hunyuan.py
@@ -37,7 +37,7 @@ class HunyuanModel(WanModel):
     only the pieces that differ for Hunyuan.
     """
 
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
iirc HY has two encoders, llama and clip. Currently this only expects a single encoder embedding? Could you double check?

SolitaryThinker (approved):

PR #1178, opened 2026-03-22
[bugfix]: fix SP deadlock in negative prompt encoding during training

alexzms:
## Summary
  - Fix NCCL deadlock when training with `sp_size > 1`
  - Root cause: `ensure_negative_conditioning()` created an inference pipeline on rank 0 only, but `post_init()` called `warmup_sequence_parallel_communication()` which does all-to-all on the SP group — other ranks never participated, causing a hang
  - Fix: every rank now encodes the negative prompt independently using a lightweight tokenizer + T5 encoder, eliminating the rank-0 guard and all broadcast logic (small overhead)

  ## Tests
  - [x] Tested with `sp_size=4` on 4 and 8 GPUs. No hangs, no quality degradation
  - [x] Pre-commit passes

### DIFF
--- a/fastvideo/train/models/wan/wan.py
+++ b/fastvideo/train/models/wan/wan.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import copy
-import gc
 from typing import Any, Literal, TYPE_CHECKING
 
 import torch
--- a/fastvideo/train/models/wan/wan.py
+++ b/fastvideo/train/models/wan/wan.py
@@ -36,6 +36,7 @@ from fastvideo.train.utils.module_state import (
     apply_trainable, )
 from fastvideo.train.utils.moduleloader import (
     load_module_from_path, )
+from fastvideo.train.utils.negative_prompt import encode_negative_prompt
 
 if TYPE_CHECKING:
     from fastvideo.train.utils.training_config import (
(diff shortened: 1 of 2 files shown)


fastvideo/train/models/wan/wan.py
```diff
@@ -341,95 +337,46 @@ def ensure_negative_conditioning(self) -> None:
...
+            outputs.attention_mask = text_inputs["attention_mask"]
+            neg_embeds = t5_postprocess_text(outputs).to(device=device, dtype=dtype)
+            neg_mask = text_inputs["attention_mask"].to(device=device, dtype=dtype)
 
-        world_group.broadcast(neg_embeds, src=0)
-        world_group.broadcast(neg_mask, src=0)
+        del text_encoder, tokenizer
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
```
Edenzzzz: This is probably not needed, they are already lazily freed by pytorch and space is reclaimed on OOM. Unless another process is sharing GPU

alexzms: ok

fastvideo/train/models/wan/wan.py
```diff
@@ -341,95 +336,43 @@ def ensure_negative_conditioning(self) -> None:
...
-        world_group.broadcast(embed_shape, src=0)
-        world_group.broadcast(mask_shape, src=0)
+        encoder_config = tc.pipeline_config.text_encoder_configs[0]
+        tok_kwargs = dict(encoder_config.tokenizer_kwargs)
 
-        embed_sizes = tuple(int(x) for x in embed_shape[:embed_ndim].tolist())
-        mask_sizes = tuple(int(x) for x in mask_shape[:mask_ndim].tolist())
+        text_enc_dtype = PRECISION_TO_TYPE[tc.pipeline_config.text_encoder_precisions[0]]
+        tokenizer = AutoTokenizer.from_pretrained(os.path.join(model_path, "tokenizer"))
+        text_encoder = T5EncoderModel.from_pretrained(
```
SolitaryThinker: It should still be possible to just use the prompt encoding stage instead of creating a new instance? I'm pretty sure this is the wrong encoder for `wan`. There's also a `UMT5EncoderModel` which should be loaded by the existing prompt stage

alexzms:
@SolitaryThinker thanks — both points were correct, sorry for the lazy first pass. Pushed a follow-up commit (61f6b39) that addresses them:

- **UMT5 vs T5.** You were right that loading the Wan text encoder via `transformers.T5EncoderModel` is wrong — Wan uses UMT5 (per-layer relative position bias, not shared), so `T5EncoderModel.from_pretrained` was silently giving us bad embeddings for the negative prompt. The repo's own `tests/encoders/test_t5_encoder.py` even uses `UMT5EncoderModel` as the reference for the Wan checkpoint, which I should have caught.
- **Reusing the prompt-encoding stage.** Switched to `TextEncoderLoader` so the encoder class is resolved from `pipeline_config.text_encoder_configs` (UMT5 for Wan) and the postprocess function comes from `pipeline_config.postprocess_text_funcs[0]` instead of being imported by name. Tokenizer is still loaded with `AutoTokenizer.from_pretrained` since `TextEncoderLoader` doesn't handle tokenizers.

The original deadlock fix is preserved: every rank loads its own encoder and encodes the negative prompt independently, no full `WanPipeline` construction, no NCCL collectives. I also force `text_encoder_cpu_offload=False` for this short-lived load so it doesn't initialize an FSDP device mesh (which would re-introduce collectives).

`pre-commit run` passes locally (yapf / ruff / mypy / codespell). Mind taking another look when you have a minute?

SolitaryThinker (approved):

PR #837, opened 2025-10-10
[Feature] Add Cosmos2 i2v pipeline

kevin314:

PR #1224, opened 2026-04-08
[feat] Cosmos 2.5 training support in fastvideo.train

alexzms:
## Summary
Add Cosmos 2.5 (Predict2.5-2B) model plugin for fastvideo/train framework
Add preprocessing pipeline and overfit config for Cosmos 2.5
Fix flow-matching noise schedule, FSDP dtype handling, and VAE normalization'
## Test
Overfit test: 1 GPU, 480×832, 93 frames, 1000 steps
Loss: 0.075 → 0.057, grad norm stable ~0.43
Validation videos verified clean at steps 0, 150, 300, 500, 1000

### DIFF
--- a/fastvideo/train/models/cosmos/cosmos.py
+++ b/fastvideo/train/models/cosmos/cosmos.py
@@ -1,0 +1,341 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Cosmos model plugin (per-role instance).
+
+Subclasses WanModel since Cosmos uses the same
+FlowMatchEulerDiscreteScheduler.  Differences:
+  - transformer class name: CosmosTransformer3DModel
+  - normalize_dit_input("cosmos", ...) instead of ("wan", ...)
+  - forward kwargs: no encoder_attention_mask, needs
+    condition_mask + padding_mask + fps
+  - hidden_states in (B,C,T,H,W) — no permute needed
+  - default flow_shift = 1.0
+  - single T5 text encoder (not dual like Hunyuan)
+"""
+
+from __future__ import annotations
+
+import copy
+import gc
+from typing import Any, Literal, TYPE_CHECKING
+
+import torch
+
+from fastvideo.forward_context import set_forward_context
+from fastvideo.pipelines import TrainingBatch
+from fastvideo.training.training_utils import (
+    normalize_dit_input, )
+
+from fastvideo.train.models.wan.wan import WanModel
+
+if TYPE_CHECKING:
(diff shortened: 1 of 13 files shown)


alexzms:
The CI failure here is unrelated to this PR's changes — it's a test infra issue.

`fastvideo/tests/transformers/test_cosmos.py` (added in #837, untouched here) calls `maybe_download_model("nvidia/Cosmos-Predict2-2B-Video2World", ...)` at **module-import time**, so pytest can't even collect the file when the HF token lacks access:

\```
403 Forbidden: Please enable access to public gated repositories
in your fine-grained token settings to view this repository.
\```

`nvidia/Cosmos-Predict2-2B-Video2World` is a gated repo on HF. The CI's HF token doesn't have "Read access to gated public repositories" enabled, so any PR that touches `cosmos*` files and triggers this test will hit the same 403.

Two options:
1. Update the CI's HF fine-grained token to allow access to gated public repos (one-time fix for the whole repo — affects every cosmos-touching PR).
2. Move the module-level `maybe_download_model` call in `test_cosmos.py` into a fixture and `pytest.skip` on 403 — I'm happy to do that as a separate small PR if it's the preferred path.

This PR's own changes pass `pre-commit`, `pytest --collect-only` on the touched files, and the import smoke test locally. Let me know which path you'd like.

alexzms:
@Eigensystem

fastvideo/train/models/cosmos/cosmos.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+        return training_batch
+
+    def ensure_negative_conditioning(self) -> None:
+        """Create negative (unconditional) prompt embeddings.
+
+        Cosmos 2.5 uses Reason1 (Qwen2.5-VL) which is expensive
+        to load.  For training with ``training_cfg_rate=0`` (no
+        classifier-free guidance dropout), the negative embedding
+        is never used.  We create a zero-valued placeholder that
```
SolitaryThinker: we should throw an error if the training_cfg_rate is set to > 0 here? Otherwise we would have the same deadlock issue as the wan PR right?

SolitaryThinker: sorry not deadlock but wrong negative prompt embed?

alexzms: Thanks for mentioning this. Fixed in most recent commit

SolitaryThinker (approved):

alexzms:
Thanks!

PR #1227, opened 2026-04-09
[feat] Add Cosmos 2.5 T2W training pipeline (LoRA + full fine-tune)

Mister-Raggs:
## Purpose

Adds end-to-end LoRA and full fine-tuning support for Cosmos-Predict2.5-2B (text-to-world) in the FastVideo training framework, along with the preprocessing and example scripts needed to run it. Also fixes several bugs in existing infrastructure that blocked Cosmos 2.5 from working with \`v1_preprocess.py\` and the shared text encoding stage.

## Changes

**New files:**
- \`fastvideo/training/cosmos2_5_training_pipeline.py\` — \`Cosmos25TrainingPipeline\` subclassing \`TrainingPipeline\`. Handles Cosmos 2.5 specifics: flow-matching scheduler (shift=5.0), 18-channel input (16 latent + 1 condition mask + 1 padding mask), Reason1 100352-dim text embeddings, and skips latent normalization (applied inside the VAE encoder).
- \`examples/training/finetune/cosmos2_5/\` — README, preprocessing script, full fine-tune script, LoRA fine-tune script, and \`validation.json\`.

**Bug fixes:**
- \`fastvideo/models/dits/cosmos2_5.py\` — Replace \`nn.Linear\` with \`ReplicatedLinear\` in all attention projections (\`to_q\`, \`to_k\`, \`to_v\`, \`to_out\` in both self-attn and cross-attn). \`get_lora_layer()\` is invisible to plain \`nn.Linear\`, so LoRA adapters were silently not applied. Also fixes two \`fp32\`/\`bf16\` dtype mismatches in \`Cosmos25PatchEmbed.proj\` and \`crossattn_proj\` that crashed validation inference.
- \`fastvideo/dataset/utils.py\` — CFG dropout zero tensor was hardcoded as \`np.zeros((512, 4096))\` (T5-XXL shape). Cosmos 2.5 uses Reason1 embeddings of a different shape, causing a mismatch at training time. Fix: use \`np.zeros(shape)\` derived from the stored embedding schema.
- \`fastvideo/pipelines/preprocess/v1_preprocess.py\` — VAE config was unconditionally overwritten with \`WanVAEConfig(load_encoder=True, load_decoder=True)\`, replacing the Cosmos 2.5 VAE config with a Wan-specific one. Fix: set \`load_encoder\`/\`load_decoder\` flags on the existing config instead.
- \`fastvideo/pipelines/preprocess/preprocess_pipeline_base.py\` — Add \`.float()\` before \`.numpy()\` to handle bf16 text embeddings (numpy does not support bfloat16).
- \`fastvideo/pipelines/stages/text_encoding.py\` — Guard against empty strings with the Qwen2 tokenizer; unwrap \`Qwen2_5_VLProcessor\` to its inner tokenizer for text-only encoding.
- \`fastvideo/dataset/preprocessing_datasets.py\` — Wrap \`AutoTokenizer.from_pretrained\` in try/except for multimodal processors (Qwen2.5-VL) that raise on plain loading.

## Test Plan

```bash
# Preprocess a video dataset
bash examples/training/finetune/cosmos2_5/preprocess_cosmos2_5_t2w.sh

# LoRA fine-tune with validation every 200 steps
bash examples/training/finetune/cosmos2_5/finetune_t2w_lora.sh
```

## Test Results

Ran 2000-step LoRA training on the `wlsaidhi/crush-smol-merged` dataset (47 hydraulic press videos, 480×832, 49 frames) on a single RTX PRO 6000 (94GB VRAM):

- Training completed without errors: final loss `0.067`, grad norm `0.007`, ~2.1s/step
- Validation ran at steps 500, 1000, 1500, 2000 using domain-matched prompts — videos show clear style adaptation vs. base model
- W&B run: https://wandb.ai/raghav-kachroo/cosmos2_5_t2w_lora

<details>
<summary>Training output (final steps)</summary>

```
step: 1980/2000 | loss: 0.0674 | grad_norm: 0.0071 | lr: 9.80e-05
step: 1990/2000 | loss: 0.0689 | grad_norm: 0.0068 | lr: 9.90e-05
step: 2000/2000 | loss: 0.0671 | grad_norm: 0.0073 | lr: 1.00e-04
```

</details>

## Checklist

**Notes:**
- Preprocessing script currently uses `v1_preprocess.py`. Port to `v1_preprocessing_new` is a follow-up (~1-2h).
- 14B model configs and Cosmos 2.5 selective activation checkpointing policies are prototyped locally — can follow up if useful.


alexzms:
Thanks @Mister-Raggs — most of the fixes here are clean. Two things worth flagging before another round:

**1. `flow_shift=5.0` may not actually be in effect during training.** `Cosmos25TrainingPipeline.initialize_pipeline` sets `self.modules["scheduler"]` with the shift, but `TrainingPipeline.train()` (`fastvideo/training/training_pipeline.py:533`) unconditionally overwrites `self.noise_scheduler = FlowMatchEulerDiscreteScheduler()` (no shift) right before the loop, and `_sample_timesteps` reads from that. This is a pre-existing legacy-framework bug (Wan has the same pattern), so not something you need to fix here — but it means the 2000-step run probably wasn't using shift=5.0. Worth a sanity check that the curves are still reasonable; happy to file the framework-side fix as a separate PR.

**2. `text_encoding.py:238` chat-model branch — Copilot's flag is correct.** After `tok = getattr(tokenizer, "tokenizer", tokenizer)`, the non-chat branch uses `tok(...)` but the chat branch still uses `tokenizer.apply_chat_template(...)`. Cosmos 2.5 doesn't hit it, but it's a latent `AttributeError` for any multimodal-processor + chat-model caller. Just swap it to `tok.apply_chat_template(...)`.

After those + a rebase + green CI, should be ready. Thanks again.

Mister-Raggs:
hey @alexzms ! thanks for  the feedback.

1. sanity checked the training curves - loss settles to ~0.2-0.3 by step 200 and stays stable for all 2000 steps, grad norm settles to ~0.1 with no late-run spikes. Training was healthy despite the shift bug, so curves look reasonable.

2. Fixed the text_encoding.py:238 issue - swapped tokenizer.apply_chat_template → tok.apply_chat_template. Also rebased onto latest main. CI running now.

alexzms (approved):
Approved — nice work @Mister-Raggs! The fixes are clean and several of them actually benefit non-Cosmos paths too. Thanks for the fast turnaround on the review feedback.

alexzms:
Hey @Mister-Raggs — SSIM is currently failing. Found the cause:

The empty-string guard at `text_encoding.py:217-220`:

```python
if not processed_text.strip():
    processed_text = "."
```

HunyuanGameCraft (and likely other models) use empty string as the default `negative_prompt`. After this guard, `""` gets rewritten to `"."`, which produces a different tokenization → different negative-prompt embedding → different CFG output → SSIM regression.

The guard was meant for Qwen2/Cosmos 2.5 where empty strings produce 0 tokens, but it's currently applied globally so it bites anything using `negative_prompt=""`.

Suggested fix: scope it to the Qwen2 path only. Easiest is probably a per-encoder-config flag (e.g. `treat_empty_as_dot: bool = False` on `TextEncoderConfig`, set to `True` on `Qwen2_5_VLConfig`), then gate the rewrite on `encoder_config.treat_empty_as_dot`. Or do the rewrite inside a Cosmos 2.5–specific preprocess hook so shared `text_encoding.py` stays untouched.

Let me know if you'd like me to push the fix, or if you want to take it.

Mister-Raggs:
Hey @alexzms, thanks for the quick feedback. I've added the fix, CI is running.

PR #1230, opened 2026-04-14
[CI] Add detect-changes and conditional Docker rebuild with concurrency

Nabhya1807:
This PR improves the Docker image build workflow by adding conditional execution and concurrency control.

Changes:
- Added a detect-changes job using dorny/paths-filter to detect modified Dockerfiles
- Updated build jobs to run only when:
  - manually triggered via workflow_dispatch, OR
  - corresponding Dockerfile changes are detected
- Added workflow-level concurrency to prevent overlapping builds:
  group: docker-image-build
  cancel-in-progress: true

Validation:
- Verified workflow_dispatch triggers correct jobs
- Confirmed detect-changes outputs match modified file paths
- Verified conditional job execution using needs + if logic

Note:
Docker build failures observed in CI are unrelated to this change and originate from existing build scripts.


test.md
Eigensystem, line 1: remove this

Eigensystem:
## Concern: Auto-building on every push to main is too expensive

The current `push` trigger fires on **every commit to `main`** that touches `docker/Dockerfile.*`. Docker image builds are expensive (time, compute, registry storage), and most Dockerfile changes don't warrant an immediate image publish — they may be part of a larger set of changes that aren't ready for release yet.

**Proposal:** Trigger automatic Docker builds on **version tags** instead of push-to-main. For example:

```yaml
on:
  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

This way:
- Images are only published at intentional release points (e.g., `v0.5.0`).
- The `workflow_dispatch` path still allows manual rebuilds when needed outside of a release.
- You avoid burning CI minutes on every merged PR that happens to touch a Dockerfile.

Happy to discuss if there's a different cadence that makes more sense for this project.

Eigensystem (changes requested):

SolitaryThinker:
Closing as superseded by #1447 and #1526, which provide path-scoped CUDA matrix/manifest builds and non-cancelling concurrency in the current single-Dockerfile workflow.

PR #1242, opened 2026-04-20
[feat] Add minimal LoRA finetuning support to the YAML training stack

radicalyyyahaha:
## Purpose

Add minimal LoRA finetuning support to the modular YAML training stack under `fastvideo/train/`.

This PR is scoped to the core LoRA path and runnable LoRA examples for Wan-family and Hunyuan model plugins. MatrixGame training support, the general Wan I2V/image-conditioning training path, and the xlsx run logger were split out so this PR stays focused and reviewable.

## Changes

- Added `fastvideo/train/utils/lora.py` for training-side LoRA injection.
  - Freezes the base transformer.
  - Replaces selected linear layers with trainable LoRA wrappers.
  - Reuses the existing FSDP/HSDP mesh for DTensor wrapping.
  - Uses explicit `Replicate()` placements for replicated LoRA parameters.

- Updated `fastvideo/train/models/base.py` with a shared nested LoRA config interface:

```yaml
models:
  student:
    _target_: fastvideo.train.models.wan.WanModel
    init_from: Wan-AI/Wan2.1-T2V-1.3B-Diffusers
    trainable: true
    lora:
      enable: true
      rank: 16
      alpha: 32
      target_modules: [to_q, to_k, to_v, to_out]
```

- Updated Wan/Hunyuan model plugins to opt into the shared LoRA helper after model-specific transformer loading:
  - `fastvideo/train/models/wan/wan.py`
  - `fastvideo/train/models/wan/wan_causal.py`
  - `fastvideo/train/models/hunyuan/hunyuan.py`

- Updated `fastvideo/train/entrypoint/train.py` to select the needed attention backend for FastWan/TurboWan LoRA configs before model loading when not explicitly overridden.

- Updated `examples/train/run.sh` so training launches from the repository root and uses the module entrypoint consistently.

- Added LoRA example configs:
  - `examples/train/configs/fine_tuning/wan/t2v_lora.yaml`
  - `examples/train/configs/fine_tuning/wan/fast_t2v_lora_vsa.yaml`
  - `examples/train/configs/fine_tuning/wan/fast_ti2v_fullattn_lora_vsa.yaml`
  - `examples/train/configs/fine_tuning/wan/turbo_t2v_lora_sla.yaml`
  - `examples/train/configs/fine_tuning/hunyuan/t2v_lora.yaml`

## Test Results

- `pre-commit`: passed in CI.
- `fastcheck`: passed in CI.
- LoRA/training CI jobs passed:
  - `test-tube-lora-training-tests`
  - `test-tube-train-framework-tests`
  - `test-tube-training-tests`
  - `test-tube-training-tests-vsa`

Multi-GPU validation by @alexzms:

- Wan2.1-T2V-1.3B LoRA overfit on 4x GB200.
- FSDP/HSDP full-shard path verified.
- LoRA rank 16, 240 LoRA layers wrapped.
- Final loss approximately 0.042.
- Checkpoint save path verified.

Current note: the remaining `full-suite-passed` blocker is from `bar-chart-ssim-tests` on `test_matrixgame_similarity.py::Matrix-Game-2.0-Diffusers-Base`, after MatrixGame changes were split out of this PR.

## Checklist

- [x] PR title follows repository format
- [x] LoRA path is disabled by default unless `lora.enable: true`
- [x] Multi-GPU LoRA DTensor wrapping uses the existing FSDP/HSDP mesh
- [x] Explicit replicated DTensor placements are used for LoRA parameters
- [x] Pre-commit passed in CI
- [x] Fastcheck passed in CI
- [x] Multi-GPU LoRA validation completed


fastvideo/pipelines/preprocess/preprocess_pipeline_ode_trajectory.py
```diff
@@ -82,7 +82,7 @@ def preprocess_text_and_trajectory(self, fastvideo_args: FastVideoArgs, args):
             if data is None:
                 continue
 
-            with torch.inference_mode():
```
Eigensystem, line 85: revert this.
`inference_mode` is a new interface which provides better performance by not tracking metadata

fastvideo/pipelines/preprocess/preprocess_pipeline_text.py
```diff
@@ -56,7 +56,7 @@ def preprocess_text_only(self, fastvideo_args: FastVideoArgs, args):
             if data is None:
                 continue
 
-            with torch.inference_mode():
```
Eigensystem, line 59: same

fastvideo/pipelines/preprocess/v1_preprocess.py
```diff
@@ -43,7 +43,7 @@ def main(args) -> None:
         num_gpus=get_world_size(),
         dit_cpu_offload=False,
         vae_cpu_offload=False,
-        text_encoder_cpu_offload=False,
```
Eigensystem, line 47: set as an arg instead of hardcode

fastvideo/pipelines/composed_pipeline_base.py
```diff
@@ -226,11 +226,16 @@ def from_pretrained(cls,
                 setattr(fastvideo_args, key, value)
 
             fastvideo_args.dit_cpu_offload = False
-            # we hijack the precision to be the master weight type so that the
-            # model is loaded with the correct precision. Subsequently we will
-            # use FSDP2's MixedPrecisionPolicy to set the precision for the
-            # fwd, bwd, and other operations' precision.
-            assert fastvideo_args.pipeline_config.dit_precision == 'fp32', 'only fp32 is supported for training'
```
Eigensystem, line 286: revert this

fastvideo/train/models/wan/wan.py
```diff
@@ -73,9 +75,19 @@ def __init__(
         | None = None,
         transformer_override_safetensor: str
         | None = None,
+        lora_rank: int | None = None,
```
Eigensystem: please add these parameter to `ModelBase` class to enable LoRA fine-tuning capabilities across all models

fastvideo/train/models/wan/wan.py
```diff
@@ -138,6 +149,20 @@ def _load_transformer(
                 transformer,
                 checkpointing_type=ckpt_type,
             )
+        if self._lora_rank is not None:
```
Eigensystem: maybe move to base class?

Eigensystem (dismissed):

fastvideo/pipelines/preprocess/preprocess_pipeline_ode_trajectory.py
radicalyyyahaha: Thanks, agreed that `torch.inference_mode()` should remain the default preprocessing mode.

The reason I changed this was a failure when `text_encoder_cpu_offload=True`. In that mode the text encoder goes through the FSDP CPU-offload path (`fully_shard(..., CPUOffloadPolicy)`), not just a simple CPU move. FSDP needs tensor version counters while unsharding/materializing parameters and resharding/offloading them after forward. Tensors created under `torch.inference_mode()` do not track version counters, so the offload path can fail with `RuntimeError: Inference tensors do not track version counter`.

So the distinction is:

- without text encoder CPU offload, `inference_mode()` is fine and more efficient;
- with FSDP CPU offload, `no_grad()` is needed because it disables autograd but keeps normal tensors with version counters.

I will update the PR to keep `inference_mode()` by default and only use `no_grad()` around the text encoder path when CPU offload/FSDP is enabled.

fastvideo/pipelines/preprocess/v1_preprocess.py
radicalyyyahaha: I will avoid hardcoding `text_encoder_cpu_offload=True` and expose it as a CLI argument in `v1_preprocess.py` instead. The default will preserve the previous behavior, while low-memory runs can opt in explicitly.

fastvideo/pipelines/preprocess/preprocess_pipeline_text.py
radicalyyyahaha: same repair

fastvideo/pipelines/composed_pipeline_base.py
radicalyyyahaha: I will revert the `bf16` master-weight change in `ComposedPipelineBase`.

fastvideo/train/models/wan/wan.py
radicalyyyahaha: I will move the common LoRA constructor fields (`lora_rank`, `lora_alpha`, and `lora_target_modules`) into `ModelBase` so model plugins expose a consistent LoRA interface.

fastvideo/train/models/wan/wan.py
radicalyyyahaha: I will move the shared LoRA enable logic into a protected `ModelBase` helper. The base class can own the common checks and call `enable_lora_training()`, while each concrete model still calls the helper after its model-specific transformer loading/checkpointing setup. This keeps the loading path model-specific but makes LoRA activation reusable.

alexzms:
Thanks for the great PR! @radicalyyyahaha Two things before another round:

**1.** This PR is doing quite a lot at once — LoRA infra, MatrixGame model plugin, and the xlsx run logger under `scripts/training/` are three pretty independent pieces. Would you mind splitting? Each one is easier to review (and revert if needed) on its own, and the xlsx logger in particular doesn't share much with the LoRA path.

**2.** You mentioned the GPU/training stack wasn't available locally so e2e wasn't run — totally understandable, but given how much this touches (5 model plugins + the loader + multi-GPU LoRA mesh handling), we'd want at least a 1-step training smoke run per touched model on multi-GPU before merging. Gemini also flagged two issues on `lora.py` (mesh creation + missing `placements`) that look like they'd only show up at ≥2 GPUs, worth double-checking.

---

If you'd like, I can push commits onto your branch to help with the split + the LoRA mesh fixes (keeping authorship on your originals). Or if you'd rather drive it yourself, just let me know.

radicalyyyahaha:
@alexzms Thanks, I’d appreciate your help with the split and the LoRA mesh fixes.

Please feel free to push commits onto my branch. After that, I’ll review the changes, update the PR description to match the narrowed scope, and follow up on the smoke test evidence. I don’t currently have a reliable multi-GPU setup, so if you’re able to run or trigger the 1-step multi-GPU smoke tests, that would be very helpful.

alexzms:
Ran the multi-GPU verification on our side. @radicalyyyahaha

Wan2.1-T2V-1.3B on 4× GB200, FSDP full-shard, LoRA rank=16. Pushed a fix for the Gemini-flagged `lora.py` issues (mesh + missing `placements`) — they're real at ≥2 GPUs; now reuses the FSDP mesh with explicit `Replicate()` placements.

End-to-end LoRA training is verified: 4000-step run is clean, checkpointing works, and a 4-sample overfit (loss 0.10 → 0.04) clearly reproduces the training clips in validation. (Heads up: validating at very low sampling steps hides the LoRA effect — use ~50.)

Thanks for the solid contribution. Next I'll push the narrow-down commits (split out MatrixGame + xlsx logger as discussed), then this should be good to merge.

alexzms:
One more split, @radicalyyyahaha — heads up: the new `fastvideo/train/` stack doesn't support Wan I2V at all today (no I2V configs, no image-conditioning path, T2V-only dataloader). This PR effectively adds Wan I2V training as a side effect (`_uses_image_conditioning` / `_augment_noisy_input_with_image_condition` / the i2v dataloader branch in `wan.py`).

Same as MatrixGame and the xlsx logger — that's an independent feature from minimal LoRA, so I'll pull the Wan I2V image-conditioning out of this PR too. Worth its own PR with proper I2V validation. Keeping this one strictly T2V LoRA; the I2V pieces are easy to re-land separately afterward.

alexzms (approved):
Verified end-to-end on the rebased branch: 4-sample LoRA overfit of Wan2.1-T2V-1.3B on 4×GB200 (HSDP shard=4) — final loss ≈0.042, 240 LoRA layers wrapped, checkpoints saved, no errors.

LoRA is now configured via a nested `models.<role>.lora` block with an explicit `enable` flag (default `false`) for better config readability:

```yaml
  student:
    _target_: fastvideo.train.models.wan.WanModel
    init_from: Wan-AI/Wan2.1-T2V-1.3B-Diffusers
    trainable: true
    lora:
      enable: true
      rank: 16
      alpha: 32
      target_modules: [to_q, to_k, to_v, to_out]
```

radicalyyyahaha:
@alexzms  Thanks again for the split and validation.

I’ll update the PR description so it matches the narrowed LoRA-only scope.

I also checked the remaining merge blocker: `full-suite-passed` is failing in `bar-chart-ssim-tests`, specifically `test_matrixgame_similarity.py::Matrix-Game-2.0-Diffusers-Base`. Since MatrixGame has been split out of this PR, should we treat this as unrelated CI/full-suite infra, or is there anything you’d like me to change here?

Eigensystem (approved):

PR #1248, opened 2026-04-23
[ci] Add CI Performance Regression Tracking Changes

Satyam-53:
## Purpose

Implements T8 performance regression tracking in CI so latency, throughput, and memory regressions are automatically detected during performance runs instead of being caught manually later.

Fixes #

## Changes

- Added baseline comparison script at [compare_baseline.py].

- Added normalized tracking writes to the HF repo path `performance-tracking/{model_id}/{timestamp}_{commit}.json` with fields:
  `model_id`, `timestamp`, `commit_sha`, `gpu_type`, `latency`, `throughput`, `memory`.

- Added regression gate against mean of last 5 prior same-GPU runs, failing when regression exceeds **5% by default** (configurable via environment variable) in [compare_baseline.py].

- Added compact markdown summary output for current vs baseline metrics in [compare_baseline.py].

- Added trend plot generation for tracked metrics in [compare_baseline.py].

- Uploaded markdown summaries and trend plots as Buildkite artifacts.

- Updated perf test output schema to include `throughput_fps` in [test_inference_performance.py].

- Wired baseline comparison as post-test step in `run_performance_tests` in [pr_test.py].

## Test Plan

```bash
Validated through a separate test PR that triggered the full Buildkite CI flow.

Verified end-to-end execution of:
- Fastcheck
- Performance tests
- Baseline comparison step
- Artifact uploads
```

## Test Results

<details>

```
- Successfully passed Buildkite-triggered tests.
- Verified performance tests completed end to end.
- Verified Buildkite artifact uploads for markdown summaries and trend plots.
- Verified metrics JSON uploads to the HF repo.
- Verified regression comparison executed successfully in CI.
```
</details>

## Checklist

- [x] I verified SSIM regression tests pass


Satyam-53:
@Eigensystem Can you please take a look at this PR?

Eigensystem (commented):
Pretty nice work! While digging into how vLLM runs its perf CI (https://github.com/vllm-project/vllm/tree/main/.buildkite/performance-benchmarks), I noticed a few gaps we should close so the coverage is deep enough to actually catch regressions.
### 1. Broaden model/hardware coverage (config-only)
Only `wan-t2v-1.3b.json` exists today. The harness auto-discovers *.json, so we can scale coverage purely by adding files — suggest one JSON per (model, resolution, frames, parallelism) for Wan 1.3B/14B(T2V+I2V), Hunyuan, LTX, etc. vLLM has 17+ JSONs across 6 hardware platforms for comparison.

### 2. Median aggregation + more runs
num_warmup=1, num_measure=3 with mean will flake the 15% gate on noisy hardware. Bump to warmup≥2, measure≥5, and compare medians instead of means.

### 3. Record component-level timings
Only end-to-end latency is tracked today — a VAE or text-encoder regression is invisible unless huge. Emit `text_encoder_time_s`, `dit_time_s`, `vae_decode_time_s` separately and gate each, same spirit as vLLM tracking TTFT/TPOT/ITL independently.

### 4. Emit trend plots as Buildkite artifacts
For main-branch runs, render a simple Plotly HTML of the last ~30 days per `benchmark_id` and upload via buildkite-agent annotate. Free "dashboard" and much faster regression bisect than diffing JSONs on the Volume. vLLM's compare-json-results.py does exactly this.

### 5. Persist baselines somewhere durable
A Modal Volume is private, has no UI, and loses history if recreated. A GH or HF dataset repo gives us durability, a public URL, and a migration path to a real dashboard later.

Satyam-53:
> Pretty nice work! While digging into how vLLM runs its perf CI (https://github.com/vllm-project/vllm/tree/main/.buildkite/performance-benchmarks), I noticed a few gaps we should close so the coverage is deep enough to actually catch regressions.
> 
> ### 1. Broaden model/hardware coverage (config-only)
> Only `wan-t2v-1.3b.json` exists today. The harness auto-discovers *.json, so we can scale coverage purely by adding files — suggest one JSON per (model, resolution, frames, parallelism) for Wan 1.3B/14B(T2V+I2V), Hunyuan, LTX, etc. vLLM has 17+ JSONs across 6 hardware platforms for comparison.
> 
> ### 2. Median aggregation + more runs
> num_warmup=1, num_measure=3 with mean will flake the 15% gate on noisy hardware. Bump to warmup≥2, measure≥5, and compare medians instead of means.
> 
> ### 3. Record component-level timings
> Only end-to-end latency is tracked today — a VAE or text-encoder regression is invisible unless huge. Emit `text_encoder_time_s`, `dit_time_s`, `vae_decode_time_s` separately and gate each, same spirit as vLLM tracking TTFT/TPOT/ITL independently.
> 
> ### 4. Emit trend plots as Buildkite artifacts
> For main-branch runs, render a simple Plotly HTML of the last ~30 days per `benchmark_id` and upload via buildkite-agent annotate. Free "dashboard" and much faster regression bisect than diffing JSONs on the Volume. vLLM's compare-json-results.py does exactly this.
> 
> ### 5. Persist baselines somewhere durable
> A Modal Volume is private, has no UI, and loses history if recreated. A GH or HF dataset repo gives us durability, a public URL, and a migration path to a real dashboard later.

Thank you for these insights.!!
I will look into the coverage and trend plots as Buildkite artifacts although I am not sure at this point how they are plotted so will have to dig into it a bit.
More runs and component level timings thing I already have on my next steps, inspired from SGLang since it also does this. 
For point 5, it is already implemented and tested end to end with using hf_repo as the storage, I will merge that code with this repo in some time.

SolitaryThinker (commented):
Two issues worth addressing before merge.

fastvideo/tests/performance/compare_baseline.py
```diff
@@ -0,0 +1,319 @@
...
+    "PERFORMANCE_TRACKING_ROOT",
+    "/root/data/performance-tracking",
+)
+MAX_REGRESSION = float(os.environ.get("PERF_MAX_REGRESSION", "0.15"))
+
+
+def _should_persist_tracking() -> bool:
+    test_scope = os.environ.get("TEST_SCOPE", "")
+    branch = os.environ.get("BUILDKITE_BRANCH", "")
+    return test_scope == "full" and branch == "main"
```
SolitaryThinker, line 45: 🔴 **Blocking**: this gate is unreachable in the current CI graph.

`BUILDKITE_BRANCH == "main"` is never satisfied:
- `/test full`, `/merge`, and the `ready`-label trigger all POST to Buildkite with `branch: $PR_BRANCH` (PR source branch, not `main`) — see [`ci-slash-commands.yml#L70-L77`](../blob/main/.github/workflows/ci-slash-commands.yml#L70-L77) and [`ci-trigger-full-suite.yml#L67-L72`](../blob/main/.github/workflows/ci-trigger-full-suite.yml#L67-L72).
- No GH workflow triggers Buildkite on push to `main`, Mergify's squash doesn't either, and there is no scheduled Buildkite job in the repo.

Effect: `_write_tracking_record` never runs in CI → the volume stays empty → every run logs `"No baseline records … skipping"` → `_check_regressions` returns `[]` → script always exits 0. The 15% gate fires zero times.

fastvideo/tests/modal/pr_test.py
```diff
@@ -230,10 +236,17 @@ def run_lora_extraction_tests():
                   modal.Secret.from_dict(
                       {"HF_API_KEY": os.environ.get("HF_API_KEY", "")})
               ],
-              volumes={"/root/data": model_vol})
+              volumes={
+                  "/root/data": model_vol,
+                  "/root/data/performance-tracking": performance_tracking_vol,
+              })
```
SolitaryThinker: Nesting `performance_tracking_vol` inside `model_vol`'s mount path (`/root/data` ⊃ `/root/data/performance-tracking`). Modal's docs don't describe this case and commit/reload semantics for the inner volume are not guaranteed. Recommend verifying with a one-off run or flattening to side-by-side mounts (e.g. `/root/perf`).

Satyam-53: This PR is not updated currently. We are no longer creating a separate modal volume for perf tracking.

fastvideo/tests/performance/compare_baseline.py
Satyam-53: yeah, we decided that we won't run the perf. tracking on every PR of github action, rather we'll schedule a buildkite job, right?

fastvideo/tests/performance/hf_store.py
```diff
@@ -0,0 +1,254 @@
...
+            repo_id=HF_REPO_ID,
+            repo_type="dataset",
+            local_dir=local_dir,
+            token=HF_TOKEN,
+            allow_patterns="*.json",
+        )
+    except Exception as exc:
+        print(f"hf_store: sync skipped — {exc}")
+
+    return local_dir
```
SolitaryThinker, line 86: Silent sync failure corrupts the baseline on persist runs.

On a `main + full` build, if `snapshot_download` fails (network blip, auth glitch, repo-access issue), this falls through, returns the empty `local_dir`, and `compare_baseline.main()` then sees `baseline_records == []` for every benchmark → hits the `INITIALIZATION CASE` branch → marks the current record `success=True` → uploads it. You've now seeded a fake "first-run" record on top of a healthy historical dataset.

When `_should_persist_tracking()` is True, sync failures should hard-fail (raise) so the build goes red and a human investigates. Resilience is fine for PR runs / dashboard reads — but not when the same run is also writing to the source of truth.

fastvideo/tests/performance/hf_store.py
```diff
@@ -0,0 +1,254 @@
...
+        api.upload_file(
+            path_or_fileobj=local_path,
+            path_in_repo=path_in_repo,
+            repo_id=HF_REPO_ID,
+            repo_type="dataset",
+            commit_message=f"Perf: {model_id} at {commit_sha}",
+        )
+        print(f"hf_store: uploaded → {HF_REPO_ID}/{path_in_repo}")
+    except Exception as exc:
+        print(f"hf_store: upload failed — {exc}")
```
SolitaryThinker, line 126: Same shape of issue as `sync_from_hf` — on a persist run, a swallowed upload failure means the record is never durable. The build still exits 0, the next run pulls the same stale baseline, and there's no signal that the rolling window stopped advancing. Hard-fail on persist runs.

fastvideo/tests/performance/compare_baseline.py
```diff
@@ -0,0 +1,319 @@
...
+            print(f"No baseline for {record['model_id']} on {record['gpu_type']}. Initializing...")
+            failures = []
+            record["success"] = True  # The first run is always "successful"
+        else:
+            # COMPARISON CASE: Compare against the mean of the last 5 good runs
+            failures = _check_regressions(record, baseline_records, MAX_REGRESSION)
+            if failures:
+                record["success"] = False
+                all_failures.extend(failures)
+            else:
```
SolitaryThinker: `_check_regressions` is called twice and the first result is always discarded:

```python
failures = _check_regressions(record, baseline_records, MAX_REGRESSION)  # ← computed

if not baseline_records:
    failures = []                                                          # ← overwritten
    record["success"] = True
else:
    failures = _check_regressions(record, baseline_records, MAX_REGRESSION)  # ← recomputed
    ...
```

Fold to a single call after the if/else split (or inside the `else` only). Also: the `else` comment still says "mean of the last 5 good runs" but `_baseline_metric` uses `statistics.median` now.

fastvideo/tests/modal/pr_test.py
```diff
@@ -229,12 +240,20 @@ def run_lora_extraction_tests():
...
 def run_performance_tests():
     run_test(
-        "export HF_HOME='/root/data/.cache' && hf auth login --token $HF_API_KEY && pytest ./fastvideo/tests/performance -vs"
+        "export HF_HOME='/root/data/.cache' && "
+        "export PERFORMANCE_TRACKING_ROOT='/tmp/perf-tracking' && "
+        "hf auth login --token $HF_API_KEY && "
+        "pytest ./fastvideo/tests/performance -vs && "
+        "python ./fastvideo/tests/performance/compare_baseline.py && "
+        "python ./fastvideo/tests/performance/dashboard.py"
     )
```
SolitaryThinker: The `&&` chain skips the dashboard when it's most useful: if `compare_baseline.py` exits 1 (regression detected), `dashboard.py` never runs and the post-run hook can't find a `dashboard_<sha>_*.html` to annotate. Same if pytest blows past the absolute threshold.

Suggest decoupling so dashboard renders regardless, while preserving the regression exit code:

```bash
pytest ./fastvideo/tests/performance -vs && \
  PERF_RC=0; python ./fastvideo/tests/performance/compare_baseline.py || PERF_RC=$?; \
  python ./fastvideo/tests/performance/dashboard.py || true; \
  exit $PERF_RC
```

(or pull this into a small wrapper script).

fastvideo/tests/modal/pr_test.py
```diff
@@ -66,6 +70,13 @@ def run_test(pytest_command: str):
     {pytest_command}
     """
 
+    # result = subprocess.run(["/bin/bash", "-c", command],
+    #                         stdout=sys.stdout,
+    #                         stderr=sys.stderr,
+    #                         check=False)
+
+    # sys.exit(result.returncode)
```
SolitaryThinker: Dead block: a commented-out copy of the active `subprocess.run` block immediately below, plus the `sys.exit(...)` that the [`9343d4cd`](https://github.com/hao-ai-lab/FastVideo/pull/1248/commits/9343d4cdf4e47ebe84da9d843730542a421db23d) commit intentionally removed. Leaving it commented out invites someone to "restore" it and reintroduce the Modal-container crash. Just delete it.

SolitaryThinker (commented):
Major progress since the last round — the HF-backed storage, median + 2/5 runs, 5% threshold, dashboard, and dropping the nested Modal volume all land cleanly. My previous two blockers are effectively addressed (pending the actual Buildkite scheduled job for `BUILDKITE_BRANCH=main` + `TEST_SCOPE=full`).

New findings below — most are minor but a few touch correctness on the critical "main + full" path.

**Cleanup batch (please squash before merge)**
- `compare_baseline.py:17` and `:24-25` have dead imports (`HfApi`, `snapshot_download`) and dead module-level constants (`HF_REPO_ID`, `HF_TOKEN`) — already encapsulated inside `hf_store.py`.
- `compare_baseline.py:42-50` redefines `_sanitize` / `_safe_float` despite importing `sanitize` / `safe_float` from `hf_store`. Pick one.
- `compare_baseline.py:1-9` docstring still says "writes to Modal volume", "mean", "15%" — three of four bullets are now wrong.
- `dashboard.py:3-4` imports `shutil` and `subprocess`, neither used.
- `pr_test.py:240-243` puts `HF_REPO_ID` inside `Secret.from_dict`, but it's a public repo identifier, not a secret. Belongs in `image.env({…})`.
- `pr_test.sh:85, 99` — `find … -name "perf_*${SHORT_SHA}*"` (and `dashboard_*${SHORT_SHA}*`) match the SHA anywhere in the filename. Tighten to `perf_${SHORT_SHA}_*` to avoid false matches when the volume contains residual files from another build.

**Future-proofing (not blocking)**
- `perf_reports/` lives on the shared `hf-model-weights` model-weights volume; concurrent perf builds could collide on read/delete. Per-build subdir (`perf_reports/${BUILDKITE_BUILD_ID}/`) would isolate them.
- Dashboard HTML is one Plotly figure per (benchmark × GPU × metric); once Eigensystem's coverage broadening lands, the single `buildkite-agent annotate` could exceed the 100KB display / 1MB hard limits.

SolitaryThinker:
Pushed [`4288e8c1`](https://github.com/hao-ai-lab/FastVideo/pull/1248/commits/4288e8c1) addressing the 5 inline comments from my last review. Resolved those threads.

Summary of changes:
- **`hf_store`**: `sync_from_hf` and `upload_record` gain a `strict=` kwarg. `compare_baseline.main()` passes `strict=persist_tracking` so PR / dashboard reads stay resilient while main-branch writes hard-fail on HF errors.
- **`pr_test.py`**: dashboard now runs after `compare_baseline` regardless of regression result via `{ ... ; PERF_RC=$? ; ... ; exit $PERF_RC; }`. Commented-out `subprocess.run` block deleted, with an inline note about the Modal `sys.exit(0)` gotcha so it stays gone. `HF_REPO_ID` moved from `Secret.from_dict` to `image.env({...})`.
- **`pr_test.sh`**: `find -name` patterns tightened to `${prefix}_${SHORT_SHA}_*`.
- **`compare_baseline.py`**: dropped dead `huggingface_hub` imports + `HF_REPO_ID`/`HF_TOKEN` constants + redefined `_sanitize`/`_safe_float` (now uses the `hf_store` versions). Refreshed module docstring (HF storage / median / 5%). Collapsed the duplicate `_check_regressions` call. Also fixed a pre-existing `unparseable` codespell typo in `hf_store`.

Verified locally: `yapf --diff`, `ruff check`, `codespell`, and `mypy` all clean on the 4 touched Python files; `bash -n` clean on the shell script.

Leaving the ["gate unreachable"](https://github.com/hao-ai-lab/FastVideo/pull/1248#discussion_r3128058200) thread open — that one needs the Buildkite scheduled job for `BUILDKITE_BRANCH=main` + `TEST_SCOPE=full` to be set up externally before merge, since this PR itself doesn't add that trigger.

SolitaryThinker (approved):

PR #1249, opened 2026-04-23
[bugfix]: normalize uint8 pil_image in I2V VAE encoding

Davids048:
## Summary

- In `PreprocessPipelineI2V`, `VideoTransformStage` stores `batch.pil_image` as a `uint8` `[0, 255]` tensor (`preprocess_stages.py:75`) right before it normalizes the rest of the video to `[0, 1]` on the next line. Because Python rebinds the variable instead of mutating the tensor, `batch.pil_image` never gets normalized.
- `ImageVAEEncodingStage.preprocess()` (`image_encoding.py:513`) then fed those `uint8` values straight into `normalize()` (`2x - 1`), producing VAE input in roughly `[-1, 509]` instead of `[-1, 1]` and silently corrupting the I2V first-frame latent. Originally reported by a user in Chinese as "`pil_image` 的范围是 0-255,而 preprocess 会把 0-255 变成 0-1 的流程跳过".
- Fixed at the consumer side. `batch.pil_image` has multiple other downstream consumers (`ImageEncodingStage` with HF `CLIPImageProcessor` which defaults to `do_rescale=True`, `record_schema.py` parquet serialization, `denoising.py`/`causal_denoising.py`) that rely on the existing `uint8` contract, so normalizing at the source would silently break CLIP features and parquet writes. Localizing the fix keeps the change minimal and risk-bounded.

### What changed

In `ImageVAEEncodingStage.preprocess()`, the tensor branch now explicitly handles `uint8 -> float / 255` and validates dtype/range; a `TypeError` is raised for unexpected outer types. The PIL branch is unchanged.

## Before / after evidence

Captured via a one-shot reproducer (`scripts/debug/verify_i2v_pil_image_range.py`, added and then removed in this branch; see commit `d2f7d43`):

|                | uint8 input              | float [0,1] input        | numerically equal |
|----------------|--------------------------|--------------------------|-------------------|
| **Pre-fix**    | `[-1.000, 509.000]` :x:  | `[-1.000, 1.000]`        | False             |
| **Post-fix**   | `[-1.000, 1.000]`        | `[-1.000, 1.000]`        | True              |

## Test plan

New file: `fastvideo/tests/stages/test_image_vae_encoding.py` — 5 CPU-only test functions (7 cases with parametrization), ~0.04s total:

- [x] `test_preprocess_pil_image` — PIL RGB with `{0, 128, 255}` pixels -> output in `[-1, 1]`, endpoints map correctly.
- [x] `test_preprocess_uint8_tensor` — uint8 tensor spanning full range -> correct mapping (this is the direct regression guard for the reported bug).
- [x] `test_preprocess_float01_tensor_matches_uint8_path` — both paths produce numerically identical output.
- [x] `test_preprocess_already_normalized_passthrough` — float tensor already in `[-1, 1]` -> returned unchanged (the `do_normalize=False` branch).
- [x] `test_preprocess_rejects_invalid_inputs` — parametrized: float tensor with max > 1 -> `ValueError`; int32 tensor -> `ValueError`; numpy array -> `TypeError`.

```
$ pytest fastvideo/tests/stages/test_image_vae_encoding.py -v
======================== 7 passed, 15 warnings in 0.04s ========================
```

No end-to-end I2V inference run was done since this change is confined to a single pure-function tensor transform and the regression tests exercise it deterministically; a full `PreprocessPipelineI2V` run would consume a GPU without adding information.

## Follow-up (not in this PR)

`batch.pil_image` carries inconsistent implicit contracts across its consumers: `ImageEncodingStage` assumes `uint8 [0, 255]`, `record_schema` assumes `np.uint8`, `denoising`/`causal_denoising` assume float `[-1, 1]`, and this PR teaches `ImageVAEEncodingStage` to accept `uint8` or float. A separate refactor should split the field (e.g. into `pil_image` / `image_tensor_uint8` / `preprocessed_image`) or add a documented dtype/range annotation, so future consumers don't re-discover which variant they receive.


Eigensystem (approved):

Eigensystem:
Hi @Davids048. Could you please upload the latest SSIM reference videos generated after this change to hf repo? This will allow the SSIM CI tests to pass.

PR #1263, opened 2026-04-27
[ci] publish docker images on release tags

Nabhya1807:
## Purpose
Switch Docker image publishing from push-to-main on Dockerfile changes 
to release-tag-based publishing (`v*.*.*`).

## Changes
- Replace `push` trigger on `main` with tag trigger `v*.*.*`
- Keep `workflow_dispatch` for manual rebuilds
- Remove Dockerfile change detection logic
- Run all image build jobs on tag pushes
- Keep concurrency protection

## Why
Publishing images on every Dockerfile-related merge to `main` is expensive 
and can publish intermediate changes. This makes image publishing happen only 
at intentional release points while preserving manual rebuilds.

## Validation
- Verified workflow YAML parses locally
- Reviewed job trigger logic for tag pushes and manual dispatch


docker/Dockerfile.python3.10
```diff
@@ -64,3 +65,5 @@ RUN source $HOME/.local/bin/env && \
 
 
 EXPOSE 22
+# test change
```
alexzms, line 68: could you confirm that these test lines are needed?

SolitaryThinker:
closing since this is superseded

PR #1272, opened 2026-05-02
[misc] attention hot-path cleanup + denoising loop hoists

alexzms:
## Summary

Small cleanup pass on the attention backends + denoising loop. Numerics bit-exact across 8 Wan2.1 runs, modest e2e improvement (within run-to-run noise on a 14s inference).

- `bsa_attn`: drop `.item()` host/device sync in `_flash_attn_single_*` (use Python `max()` over the cu_seqlens list)
- `sla`: drop redundant `.contiguous()` after `softmax`/`elu`/`relu` feature maps (elementwise, layout already preserved)
- `flash_attn`: cache mask conversions on metadata so all attention layers in one denoising step share one transformation
- `video_sparse_attn`: reuse padded buffer in `tile()` across layers/steps; precompute combined index for `untile()` (one fancy index, not two)
- `denoising`: hoist `guidance_expand` tensor / `use_meanflow` `getattr` / V2V zero-pad allocation out of the per-step loop
- `conditioning`: drop the unreachable post-return block in `ConditioningStage.forward` (CFG is handled inside `DenoisingStage`)

## Test plan

- [x] `pytest fastvideo/tests/` — to run before merging
- [x] Wan2.1-T2V-1.3B e2e: hash bit-exact (sha1 unchanged across 8 runs)

### DIFF
--- a/fastvideo/attention/backends/bsa_attn.py
+++ b/fastvideo/attention/backends/bsa_attn.py
@@ -358,11 +358,13 @@ def _flash_attn_single_mask(
     flat_k = torch.cat(k_list, dim=0)
     flat_v = torch.cat(v_list, dim=0)
 
+    # Compute max_seqlen_k from the Python list before moving to GPU to
+    # avoid a `.item()` round-trip that would force a host/device sync.
+    max_seqlen_q = Sq
+    max_seqlen_k = max(b - a for a, b in zip(cu_seqlens_k[:-1], cu_seqlens_k[1:], strict=False))
+
     cu_seqlens_q_t = torch.tensor(cu_seqlens_q, dtype=torch.int32, device=device)
     cu_seqlens_k_t = torch.tensor(cu_seqlens_k, dtype=torch.int32, device=device)
-
-    max_seqlen_q = Sq
-    max_seqlen_k = int((cu_seqlens_k_t[1:] - cu_seqlens_k_t[:-1]).max().item())
 
     orig_dtype = flat_q.dtype
     compute_dtype = orig_dtype
@@ -445,11 +447,13 @@ def _flash_attn_single_head(
     flat_k = torch.cat(k_list, dim=0)
     flat_v = torch.cat(v_list, dim=0)
 
+    # Compute max_seqlen_k from the Python list before moving to GPU to
+    # avoid a `.item()` round-trip that would force a host/device sync.
+    max_seqlen_q = Sq
+    max_seqlen_k = max(b - a for a, b in zip(cu_seqlens_k[:-1], cu_seqlens_k[1:], strict=False))
+
     cu_seqlens_q_t = torch.tensor(cu_seqlens_q, dtype=torch.int32, device=device)
     cu_seqlens_k_t = torch.tensor(cu_seqlens_k, dtype=torch.int32, device=device)
-
-    max_seqlen_q = Sq
-    max_seqlen_k = int((cu_seqlens_k_t[1:] - cu_seqlens_k_t[:-1]).max().item())
 
     orig_dtype = flat_q.dtype
     compute_dtype = orig_dtype
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -57,10 +57,36 @@ class FlashAttentionBackend(AttentionBackend):
         raise NotImplementedError
 
 
+def _key_padding_mask_from_attn_mask(attn_mask: torch.Tensor, key_len: int) -> torch.Tensor:
+    # Normalize attn_mask to [B, key_len] where True means valid token.
+    if attn_mask.dim() == 4:
+        attn_mask = attn_mask[:, 0, 0, :]
+    elif attn_mask.dim() == 3:
+        attn_mask = attn_mask[:, 0, :]
+    elif attn_mask.dim() != 2:
+        raise ValueError(f"Unsupported attn_mask shape for FLASH_ATTN: {attn_mask.shape}")
+
+    # SDPA additive mask convention: valid=0, masked=-inf/large negative.
+    key_padding_mask = attn_mask if attn_mask.dtype == torch.bool else attn_mask >= 0
+
+    if key_padding_mask.shape[-1] != key_len:
+        raise ValueError("Invalid key padding mask length for FLASH_ATTN: "
+                         f"expected {key_len}, got {key_padding_mask.shape[-1]}")
+    return key_padding_mask
+
+
 @dataclass
 class FlashAttnMetadata(AttentionMetadata):
     current_timestep: int
     attn_mask: torch.Tensor | None = None
+    # Per-step cache of converted/padded masks.  Mask conversion only
+    # depends on shape and on ``attn_mask`` itself, both of which are
+    # constant across all attention layers within one denoising step,
+    # so we share the converted masks across layers instead of redoing
+    # the work on every forward.  Populated lazily by the impl.
+    # Keys are tuples like ("kpad", key_len), ("qpad", batch, q_len),
+    # ("self_attn_pad", qkv_len).
+    _mask_cache: dict | None = None
 
 
 class FlashAttnMetadataBuilder(AttentionMetadataBuilder):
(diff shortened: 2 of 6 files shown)


fastvideo/attention/backends/flash_attn.py
```diff
@@ -57,10 +57,36 @@ def get_builder_cls() -> type["AttentionMetadataBuilder"]:
...
     current_timestep: int
     attn_mask: torch.Tensor | None = None
+    # Per-step cache of converted/padded masks.  Mask conversion only
+    # depends on shape and on ``attn_mask`` itself, both of which are
+    # constant across all attention layers within one denoising step,
+    # so we share the converted masks across layers instead of redoing
+    # the work on every forward.  Populated lazily by the impl.
+    # Keys are tuples like ("kpad", key_len), ("qpad", batch, q_len),
+    # ("self_attn_pad", qkv_len).
+    _mask_cache: dict | None = None
```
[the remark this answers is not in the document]
alexzms: Reachable today via `hunyuanvideo15.py:703` and `hyworld.py:177` — both call `FlashAttnMetadataBuilder().build(attn_mask=...)` directly. Wiring `get_metadata_cls`/`get_builder_cls` is a separate cleanup; not blocking this PR.

fastvideo/attention/backends/sla.py
Satyam-53, line 1: Looks good to me.

fastvideo/attention/backends/flash_attn.py
Satyam-53, line 1: Looks good to me.

fastvideo/attention/backends/bsa_attn.py
Satyam-53, line 1: Looks good to me.

fastvideo/pipelines/stages/conditioning.py
Satyam-53, line 1: Can you please mention exactly where the conditioning is happening now in the pipeline. I could not see that change as a part of this commit. Maybe it would be better to populate that detail (exact file/class name) in the comments as well?

Also, since it is a no op, what do you think about removing this class entirely?

fastvideo/pipelines/stages/denoising.py
Satyam-53, line 1: looks good to me.

fastvideo/attention/backends/video_sparse_attn.py
Satyam-53, line 1: looks good to me

Satyam-53 (commented):
Refactoring and performance changes look good to me except for one of the code files. PTAL at the comment for conditioning.py file.

alexzms:
Thanks for reviewing!

alexzms:
Thanks @Satyam-53!

Pushed an update — comment now points to the actual CFG sites (`denoising.py:364-394`, `:706`, `:930`; all separate-forward passes, fastvideo never used the batched-cat pattern, so the old code in `forward()` was diffusers-style residue rather than the active path).

Re removing the class: only `forward()` is no-op — `verify_input`/`verify_output` still validate CFG fields and handle the Matrix-Game "no prompt embeds → disable CFG" fallback, which isn't replicated elsewhere at this point in the pipeline. So I'd keep the class. Happy to consolidate later as a follow-up if you think it's worth doing.

Satyam-53:
> Thanks @Satyam-53!
> 
> Pushed an update — comment now points to the actual CFG sites (`denoising.py:364-394`, `:706`, `:930`; all separate-forward passes, fastvideo never used the batched-cat pattern, so the old code in `forward()` was diffusers-style residue rather than the active path).
> 
> Re removing the class: only `forward()` is no-op — `verify_input`/`verify_output` still validate CFG fields and handle the Matrix-Game "no prompt embeds → disable CFG" fallback, which isn't replicated elsewhere at this point in the pipeline. So I'd keep the class. Happy to consolidate later as a follow-up if you think it's worth doing.

looks good to me now. We can keep the class for now.

fastvideo/pipelines/stages/conditioning.py
Satyam-53: resolved in later commit.

Satyam-53 (approved):
I have marked it for the approval, maybe please take an approval from @SolitaryThinker as well before merging?

SolitaryThinker:
Hi Alex, this is from my agent setup's /review output. Could you double check to see if this makes sense?

---

### Issue 1: `_mask_cache` is paid-for-but-unused in current usage

In `fastvideo/attention/backends/flash_attn.py`, the new `_mask_cache: dict | None = None` field on `FlashAttnMetadata` only delivers savings if a single `FlashAttnMetadata` instance is reused across multiple `attn_impl.forward()` calls. Every current producer of `FlashAttnMetadata` builds a fresh instance per attention block call and uses it exactly once:

- `fastvideo/models/dits/hunyuanvideo15.py:702-709` — `FlashAttnMetadataBuilder().build(...)` inside the per-block forward, scoped to one `set_forward_context` block over a single `self.attn(...)` call.
- `fastvideo/models/dits/hyworld/hyworld.py:176-192` and `:196-212` — same pattern; the prope path even builds a *second* metadata for the same block.

`DistributedAttention.forward` calls `self.attn_impl.forward(...)` exactly once per call, and within a single forward the cross-attn vs self-attn paths set disjoint cache keys. So in current usage the cache is never hit, even within a single call.

Net effect: the field, the lazy-init guard (`if attn_metadata._mask_cache is None: attn_metadata._mask_cache = {}`), and three `if … not in cache` lookups run on every forward but never return a cached value. Functionally correct, but the bit-exactness on Wan2.1 is also consistent with "this code path was never exercised by Wan2.1" (Wan uses VSA, not FlashAttn-with-mask). The hot-path saving the description advertises will only materialize after a corresponding caller change in `hunyuanvideo15.py` / `hyworld.py` to reuse one metadata across blocks.

Worth either (a) doing that caller change in this PR so the optimization is actually load-bearing, or (b) trimming the cache and keeping just the helper-extraction.

There is also a latent fragility: the `("kpad", key_len)` key does not include a fingerprint of `attn_mask`. If a future caller does reuse one metadata across blocks AND mutates `attn_metadata.attn_mask` in-place between calls, the cached `key_padding_mask` will be stale. Not a bug today, but worth a comment in the dataclass that the cache assumes `attn_mask` is immutable for the metadata's lifetime.

---

### Issue 2: `tile_buf` reuse establishes an implicit "consume before next call" contract

In `fastvideo/attention/backends/video_sparse_attn.py`, `tile()` now returns the shared `attn_metadata.tile_buf` rather than a fresh allocation. The PR's correctness argument — pad positions stay zero across reuses because `non_pad_index` is fixed on the metadata — holds, since `non_pad_index` is set once in `build()` from an `lru_cache`'d helper.

The remaining concern is aliasing on the caller side. Both VSA call paths I checked are safe today:

- `DistributedAttention_VSA.forward` (`fastvideo/attention/layer.py:204-207`): `qkvg = preprocess_qkv(...)` → `chunk(4)` → `attn_impl.forward(q, k, v, gate_compress, ...)`, where `forward` immediately runs `.transpose(1, 2).contiguous()` on each chunk, materializing a copy before returning.
- `fastvideo/models/dits/ltx2.py:1084-1088`: same pattern.

Both consume the buf and copy it before any other layer's `preprocess_qkv` runs, so the reuse is safe under sequential single-stream execution. But this is now an unstated invariant of the backend: any future caller that holds the `tile()` return value across another VSA layer's `preprocess_qkv` call (or that runs them concurrently on different CUDA streams) will silently get corrupted data.

Consider documenting on `tile()` (or `preprocess_qkv`) that the returned tensor is only valid until the next VSA-layer call on the same metadata. Same comment applies more weakly to the `_mask_cache` field — worth noting the cache assumes `attn_metadata.attn_mask` is not mutated for the metadata's lifetime.

---

No correctness blockers — just a gap between claim and reality on Issue 1, and an implicit contract worth documenting on Issue 2.

alexzms:
Thanks @SolitaryThinker — yep, both points hold up. Just pushed `0d2691d1` to address them:

**Issue 1 (`_mask_cache`)** — agreed, every current caller (`hunyuanvideo15.py:702`, `hyworld.py:177` & `:197`) builds a fresh `FlashAttnMetadata` per attention block, so the cache never hits. Dropped the field, lazy-init guard, and the three lookup branches; kept `_key_padding_mask_from_attn_mask` as a top-level helper since the extraction stands on its own. Making the cache load-bearing would need a caller-side refactor to share one metadata across blocks — happy to do that as a follow-up rather than expand scope here.

**Issue 2 (`tile_buf` contract)** — added a docstring on `tile()` and `preprocess_qkv()` spelling out that the returned tensor aliases `attn_metadata.tile_buf` and is only valid until the next VSA-layer call on the same metadata, with a note that the current call sites copy via `.transpose(...).contiguous()` inside `forward()` so the contract holds today.

SolitaryThinker (approved):

Issue #1283, opened 2026-05-04
[Feature] Create a agent skill to handle the intentional performance baseline shifts.


Satyam-53:
### Motivation

Add a manual agent workflow for approved performance baseline reseeding.

  FastVideo performance CI compares current benchmark metrics against a rolling HF baseline, using the median of recent
  successful records for each `(model_id, gpu_type)`. When an intentional runtime or dependency change shifts metrics, such as a
  torch/container upgrade increasing peak memory, CI can fail even though there is no FastVideo logic regression.

  This feature adds a documented agent skill to handle that case safely: after maintainers review and approve the shifted result,
  the workflow backs up existing HF performance records, uses the failed Buildkite `perf_*.json` artifact as the source result,
  creates reviewed `success=true` baseline records with clear provenance, uploads them to `FastVideo/performance-tracking`, and
  allows CI to be rerun against the accepted baseline.

  This keeps performance regressions visible while providing an auditable path for legitimate baseline shifts.

PR #1292, opened 2026-05-07
[ci] Component time performance + reseed hf baseline skill

Satyam-53:
## Purpose

Add component-level timing coverage to inference performance benchmarks so regressions can be tracked separately for text encoding, DiT/denoising, and VAE decode.

Persist the new timing fields into performance tracking records, compare them against rolling baselines, and include them in dashboard plots when present. Add per-component thresholds for the Wan T2V performance benchmark while keeping the dashboard compatible with older records that do not contain component timing columns.

Also improve the performance-tracking CI flow and reseed workflow so reviewed normalized performance artifacts can be used to safely advance accepted baselines after intentional runtime or environment shifts.

  Fixes: Takes care of #1283 along with other CI changes.

  ## Changes

 - Capture component timings from pipeline stage logging and write `text_encoder_time_s`, `dit_time_s`, and `vae_decode_time_s` into performance result JSONs.
  - Enable `FASTVIDEO_STAGE_LOGGING` through the OS environment during performance tests so spawned worker processes collect stage timing data.
  - Map only real pipeline stage classes to component metrics: `TextEncodingStage`, `DenoisingStage`, `DmdDenoisingStage`, and `DecodingStage`.
  - Persist, compare, summarize, and dashboard component timing metrics when present, while remaining compatible with older records that do not include those fields.
  - Add component timing thresholds for the Wan T2V 1.3B benchmark.
  - Upload normalized performance artifacts from `compare_baseline.py` for rolling-baseline comparison failures.
  - Run dashboard generation even when performance CI is red, so failed runs remain observable.
  - Reuse Hugging Face performance tracking syncs within the same CI job, with a freshness check to avoid stale local syncs in persistent environments.
  - Update the performance reseed skill to support reviewed batches of normalized JSONs, validate intra-batch consistency, infer upload fields from existing HF records, back up state under `/tmp`, and offer cleanup after successful reseeding.

  ## Test Plan

 python -m py_compile \
    fastvideo/tests/performance/test_inference_performance.py \
    fastvideo/tests/performance/compare_baseline.py \
    fastvideo/tests/performance/dashboard.py \
    fastvideo/tests/performance/hf_store.py \
    fastvideo/tests/modal/pr_test.py

  bash -n .buildkite/scripts/pr_test.sh
 
 Run end to end buildkite triggered tests, also verified manually through the logs and hf-repo.

  ## Test Results
  <details>

  Command completed successfully along with manual verifications.
  </details>

  ## Checklist


fastvideo/tests/performance/test_inference_performance.py
```diff
@@ -10,17 +10,34 @@
...
+    "T5TextEncoderStage": "text_encoder_time_s",
+    "DiTStage": "dit_time_s",
+    "DenoisingStage": "dit_time_s",
+    "DmdDenoisingStage": "dit_time_s",
+    "TransformerStage": "dit_time_s",
+    "VAEDecodeStage": "vae_decode_time_s",
+    "VAEDecoderStage": "vae_decode_time_s",
+    "DecodingStage": "vae_decode_time_s",
+    "DecodeStage": "vae_decode_time_s",
+}
```
SolitaryThinker, line 30: 8 of these 12 keys are dead — they don't match any class in the repo. Stage names come from `self.__class__.__name__` in `pipelines/stages/base.py`. Only `TextEncodingStage`, `DenoisingStage`, `DmdDenoisingStage`, and `DecodingStage` are real classes. `TextEncoderStage`, `CLIPTextEncoderStage`, `T5TextEncoderStage`, `DiTStage`, `TransformerStage`, `VAEDecodeStage`, `VAEDecoderStage`, `DecodeStage` do not exist (`grep -rn 'class TextEncoderStage\b'` etc. returns nothing).

The dead entries don't break anything — unmapped stage names are silently skipped — but they advertise coverage that doesn't exist. Consider trimming to only stages present in the codebase. If the intent is forward-compatible coverage, note that the codebase actually has additional real denoising-stage classes that aren't mapped today (e.g. `CausalDenoisingStage`, `CausalDMDDenoisingStage`, `CosmosDenoisingStage`, `LongCatDenoisingStage`).

Severity: low–medium (misleading, not buggy).

fastvideo/tests/performance/compare_baseline.py
```diff
@@ -108,7 +158,8 @@ def _check_regressions(
 ) -> list[str]:
     failures: list[str] = []
 
-    for metric in ("latency", "memory"):
+    for metric in ("latency", "memory", "text_encoder_time_s", "dit_time_s",
+                   "vae_decode_time_s"):
```
SolitaryThinker: This hardcoded tuple duplicates `LOWER_IS_BETTER_METRICS` (defined right above and used correctly in `_metric_delta_percent`). Adding a new lower-is-better metric will require updating two places with no signal if you miss one.

Suggest:
```python
for metric in LOWER_IS_BETTER_METRICS:
    ...
```

Iteration order over a set is non-deterministic, which is fine for the failure list. If order matters for log readability, iterate `METRICS` and filter by membership.

Severity: medium (DRY violation, footgun for future metrics).

.buildkite/performance-benchmarks/tests/wan-t2v-1.3b.json
```diff
@@ -36,11 +36,17 @@
...
+      "max_dit_time_s": 10.0,
+      "max_vae_decode_time_s": 10.0
     },
     "default": {
       "max_generation_time_s": 120.0,
-      "max_peak_memory_mb": 30000.0
+      "max_peak_memory_mb": 30000.0,
+      "max_text_encoder_time_s": 5.0,
+      "max_dit_time_s": 10.0,
+      "max_vae_decode_time_s": 10.0
```
SolitaryThinker: The `default` block exists for unbranded GPUs and intentionally relaxes total-generation/memory by ~3–4× vs. L40S (120s vs 34s; 30000MB vs 11000MB). The new component thresholds copy L40S values directly — i.e. the *fastest* device — so a slower fallback GPU (e.g. L4/T4 in some future CI lane) will fail component checks despite passing the relaxed total checks.

Options:
1. Match the same scaling factor (~3×) on the `default` row.
2. Drop the component keys from `default` entirely and let `thresholds.get("max_*")` return `None` — the assertion already short-circuits on `None`, so unknown GPUs simply skip the new checks.
3. If the intent is "fail loudly on unknown GPUs", add a comment in the JSON.

Severity: medium — only matters if non-L40S devices ever run this benchmark.

fastvideo/tests/performance/test_inference_performance.py
```diff
@@ -203,3 +260,41 @@ def test_inference_performance(cfg):
...
+)
+def test_inference_performance(cfg):
+    """Measure generation latency, peak GPU memory, and component-level timings
+    (text encoder, DiT, VAE decode). Assert each against device-aware thresholds.
+    """
+
+    original_env = os.environ.get("FASTVIDEO_STAGE_LOGGING")
+    original_getter = envs.environment_variables["FASTVIDEO_STAGE_LOGGING"]
+    os.environ["FASTVIDEO_STAGE_LOGGING"] = "1"
+    envs.environment_variables["FASTVIDEO_STAGE_LOGGING"] = lambda: True
```
SolitaryThinker: The lambda override is redundant. The original lambda in `envs.py` is `lambda: bool(int(os.getenv("FASTVIDEO_STAGE_LOGGING", "0")))`. After `os.environ["FASTVIDEO_STAGE_LOGGING"] = "1"` the original already returns `True`, so the second assignment is a no-op for in-process callers. Workers (`force_spawn` → `spawn`) re-import `fastvideo.envs` from scratch, so the in-process dict mutation never reaches them — they pick up the env var via the unmodified original lambda.

Setting `os.environ` alone is sufficient. Removing the lambda override would simplify the `try/finally` (no need to capture/restore `original_getter`).

Severity: low (clarity / dead code).

fastvideo/tests/modal/pr_test.py
```diff
@@ -240,17 +240,23 @@ def run_lora_extraction_tests():
...
+        "PERF_RC=0; "
+        "if [ $PYTEST_RC -eq 0 ]; then "
+        "python ./fastvideo/tests/performance/compare_baseline.py; "
         "PERF_RC=$?; "
+        "fi; "
         "python ./fastvideo/tests/performance/dashboard.py || true; "
-        "exit $PERF_RC; }")
+        "FINAL_RC=$PYTEST_RC; "
+        "if [ $FINAL_RC -eq 0 ]; then FINAL_RC=$PERF_RC; fi; "
+        "exit $FINAL_RC")
```
SolitaryThinker, line 275: Two things worth confirming about this shell logic:

**1. Commit "Upload raw perf results for failed performance jobs" doesn't actually do that.** `normalized_perf_*.json` is only written by `compare_baseline._write_normalized_artifact`, and `compare_baseline.py` only runs when pytest passes (`if [ $PYTEST_RC -eq 0 ]; then ... fi`). So when pytest fails because a static threshold is exceeded — exactly the case the new SKILL is designed to recover from — no normalized artifact is produced and the SKILL's preferred input is missing.

   - If the current behavior is intended (the SKILL's "When not to use" disclaims fixed-threshold failures), the commit subject is misleading — those failed runs still won't have normalized artifacts uploaded.
   - If the intent is to surface a normalized artifact on pytest failure too, `compare_baseline.py` (or a small `normalize_only` mode) needs to run unconditionally so it can emit normalized JSON regardless of pytest exit code.

**2. Behavior change on dashboard.py.** Previously dashboard ran only on `pytest && compare_baseline` success; now it runs unconditionally with `|| true`. Presumably intentional for observability when CI is red, but worth a one-line note in the PR description.

Severity: low (semantics / commit-message accuracy).

fastvideo/tests/performance/hf_store.py
```diff
@@ -61,7 +71,19 @@ def sync_from_hf(local_dir: str, *, strict: bool = False) -> str:
     unavailable. Callers that depend on the sync for correctness (e.g. the
     main-branch baseline writer) must pass ``strict=True`` so that misconfig
     or transient HF errors fail loud rather than silently reset the baseline.
+
+    When ``reuse_existing=True``, a previous successful sync in ``local_dir``
+    is reused. This avoids duplicate HF snapshot checks when compare and
+    dashboard scripts run sequentially in the same CI job.
     """
+    marker_path = _sync_marker_path(local_dir)
+    if reuse_existing and os.path.exists(marker_path):
```
SolitaryThinker, line 100: `reuse_existing=True` is sound *only* because Modal containers are ephemeral. The marker file lives at `<local_dir>/.hf_sync_complete`. If `dashboard.py` is ever run twice in the same persistent environment (a developer workstation, a long-lived runner), the second invocation silently reuses whatever data was synced previously — possibly hours/days stale.

In the current Modal CI flow each function gets a fresh container, so this is safe today. Recommend either:
- A doc comment explicitly scoping `reuse_existing` to "second consumer in the same job, never across jobs", or
- A TTL check on the marker timestamp (it already records `synced_at`).

Severity: low (fragility note — current call site is fine).

SolitaryThinker (commented):
Review summary for the component-timings + reseed-baseline-skill PR. Inline comments below cover the issues I'm confident in. Top-level notes:

**Behavior change (worth a mention in PR description):** `dashboard.py` now runs unconditionally with `|| true` after pytest, where previously it only ran when both pytest and `compare_baseline.py` succeeded. That's presumably intentional for trend-graph observability when CI is red, but it's not called out anywhere.

**Things I checked and found OK:**
- Worker `spawn` propagation: `os.environ` set before `VideoGenerator.from_pretrained()` reaches spawned workers.
- `try/finally` env-var restore covers normal completion, exceptions, and KeyboardInterrupt.
- Markdown summary table column counts match the separator (11 columns each).
- `_metric_delta_percent` correctly handles all 6 metrics with positive=worse semantics; `worst_regression_pct` then `max()` is correct.
- Path glob between `compare_baseline._write_normalized_artifact` writer (`PERF_REPORTS_DIR/results/`) and `pr_test.sh:_upload_normalized_perf_results` (`*/results/normalized_perf_*.json`) match.
- `success=true` field in normalized JSON is correctly clobbered by `record["success"] = not failures` before write.

**Asks before merge:**
1. Is the `default`-block component threshold (5/10/10) deliberate, or copy-paste from L40S? (See `wan-t2v-1.3b.json` comment.)
2. Should `compare_baseline.py` (or a normalize-only equivalent) run on pytest failure too, so the SKILL has a normalized source on threshold violations? (See `pr_test.py` comment.)
3. Are the L40S thresholds (especially 10s `max_dit_time_s`) backed by a representative measurement with headroom?

SolitaryThinker:
Branched **off `origin/main`** (independent of this PR) with a follow-up to the docs/env-capture gap that came up while reviewing this work: [`ci/perf-env-capture-and-docs`](https://github.com/hao-ai-lab/FastVideo/tree/ci/perf-env-capture-and-docs) (2 commits, +694 / −17).

### What it adds

**1. New `fastvideo/tests/performance/env_capture.py`** — captures into every `perf_*.json` record (under a new `env` block, `schema_version: 1`):
- GPU: name, count, compute capability, driver version, total memory, UUIDs (via `nvidia-smi`)
- CUDA: runtime version + major
- PyTorch: version, `version_major_minor`, `built_with_cuda`
- Python, OS (system/release/machine)
- `FASTVIDEO_ATTENTION_BACKEND`
- Versions for: torch, transformers, diffusers, huggingface_hub, fastvideo, fastvideo-kernel, flash-attn, sage-attention, xformers, triton, vllm
- Container image (Modal/Docker digest when present)

All capture is best-effort — never raises, missing fields become `None`.

**2. Env-aware baseline filtering** in `compare_baseline.py`:
- Default mode (off): captured env is metadata only; surfaced in markdown, dashboard, and HF records, but doesn't affect baseline selection (preserves current behavior).
- Strict mode (`PERF_STRICT_ENV=1`): rolling baseline is filtered to records whose env tuple matches the current run on `(gpu.name, gpu.count, torch.version_major_minor, cuda.runtime_major, attention_backend)`. Override the tuple via `PERF_COMPARE_KEYS=…`. When zero records match, the comparator logs `no env-comparable records on …` and skips the regression check rather than firing a false-positive against an unrelated environment. **This is the bit that makes the workflow usable for local dev on non-CI GPUs / different torch builds.**

**3. Dashboard env-aware grouping** in `dashboard.py`:
- Time-series now buckets on `(model_id, gpu_type, torch_version, cuda_runtime, attention_backend)` so a torch upgrade no longer silently smears across the trend line. Legacy records (no `env`) get an `n/a` bucket.

**4. Markdown summary** carries an env-keys banner, a strict-mode indicator, and a per-row `Env` column.

**5. New docs page**: [`docs/contributing/performance_benchmarks.md`](https://github.com/hao-ai-lab/FastVideo/blob/ci/perf-env-capture-and-docs/docs/contributing/performance_benchmarks.md) covering the local-dev workflow, the two gates (static thresholds vs rolling baseline), strict-env mode, raw vs normalized schemas, full env-var reference, CI integration, adding a new benchmark, and troubleshooting. Linked from `mkdocs.yml` nav, `docs/contributing/ci_architecture.md`, `docs/contributing/testing.md`, and `fastvideo/tests/AGENTS.md`.

### Smoke tested

| Check | Result |
|---|---|
| `env_capture.capture_env()` on this host (B200 / torch 2.11 / CUDA 12.8) | All fields populated incl. driver_version + UUIDs from `nvidia-smi` |
| `compare_baseline.main()` end-to-end with synthetic baseline | 4.8% regression detected, env tuple shown in markdown |
| `PERF_STRICT_ENV=1` with matching torch version | 5 baselines used, regression check ran |
| `PERF_STRICT_ENV=1` with mismatched torch version | 0 baselines, "no env-comparable records" logged, run still passes |
| `dashboard.build_plots` with synthetic mixed-env data | 4 records → 3 distinct env buckets, 9 plots (3 × 3 metrics) |
| `pre-commit run --files …` on changed files | clean (perf tests excluded from yapf/ruff/mypy per existing repo config; codespell + PyMarkdown pass) |

### Relationship to this PR

The branch is **independent**: it diffs against `origin/main` directly and doesn't depend on this PR's component-time work. If this PR lands first I'll rebase and add a paragraph to the docs page covering the 3 component metrics; if my branch lands first this PR rebases trivially (no overlap on the same lines).

Branch: https://github.com/hao-ai-lab/FastVideo/tree/ci/perf-env-capture-and-docs
Compare: https://github.com/hao-ai-lab/FastVideo/compare/main...ci/perf-env-capture-and-docs

Will hold off on opening a PR until you've eyeballed the diff and the docs page.

Satyam-53:
@SolitaryThinker putting this up here for the suggestion and latter work maybe, while doing the performance regression measurements, we are taking the mean of the num_measurement runs, I feel taking median here as well should also make it more robust. 

Right now, we are using median to compare from the list of successful baseline runs with the mean of the num_measurement runs. 

Also, I have seen component level timings to fluctuate by a significant amount (more than 5%) during subsequent runs, but this was happening yesterday and could be because of Modal flaky downtime. Can do a quick round of retests on this though.

.buildkite/performance-benchmarks/tests/wan-t2v-1.3b.json
Satyam-53: I think we can short-circuit component times to none on the unknown GPUs.

fastvideo/tests/performance/compare_baseline.py
Satyam-53: I agree with the comment. The best fix here is to keep display/log ordering tied to METRICS, since that tuple already defines the report order, and filter it by LOWER_IS_BETTER_METRICS to avoid duplicating the metric list.

fastvideo/tests/performance/hf_store.py
Satyam-53: The TTL guard is in place with a one-hour default and an env override. 
PERFORMANCE_TRACKING_SYNC_REUSE_TTL_SECONDS --- env variable

fastvideo/tests/performance/test_inference_performance.py
Satyam-53: taken care of

fastvideo/tests/performance/test_inference_performance.py
Satyam-53: Agree, fixed it with removing lambda capture.

fastvideo/tests/modal/pr_test.py
Satyam-53: Keeping the current behavior: normalized performance artifacts are produced only when pytest ./fastvideo/tests/performance passes and compare_baseline.py runs. Static-threshold pytest failures will not produce normalized_perf_*.json, and the reseed skill remains scoped to rolling-baseline failures, not fixed-threshold failures.

Updating the inline code comments and calling it out on the PR specifically. Checking if the commit subject can be changed.

Satyam-53 (commented):
1. Is the default-block component threshold (5/10/10) deliberate, or copy-paste from L40S? (See wan-t2v-1.3b.json comment.)
> they were copies from the L40S but have been removed as of now since we are short-circuiting the component time to None for the unknkown GPUs.

2. Should compare_baseline.py (or a normalize-only equivalent) run on pytest failure too, so the SKILL has a normalized source on threshold violations? (See pr_test.py comment.)
> the skill is currently to reseed the rolling performance baseline on hf-repo and not to be used for hard set threshold failures, so it does not make sense to run the compare_baseline.py on pytest failures. If the hard failures hit (which they should not unless a major change is done) they should be changed directly on the config json files, and the skill can take over from there to reseed the perf baseline as well.

3. Are the L40S thresholds (especially 10s max_dit_time_s) backed by a representative measurement with headroom?
> I did run some number of tests to see their fluctuation and then set them with a decent enough headroom. As far as I can remember the headroom currently is 1.5-2x of their median value across runs.

SolitaryThinker:
**Hi @Satyam-53** — this is a code review from one of @SolitaryThinker's AI reviewer agents (Gob). I run these to help triage PRs but @SolitaryThinker hasn't personally verified every finding. If anything below doesn't match what you know about the code, **please ping @SolitaryThinker** — they'll take a closer look. (Apologies for the delay — the review was conducted ~2 days ago at HEAD `4c6c4ac9`, which is still the current HEAD, so all findings still apply.)

## TL;DR

Plumbing is mostly correct and the backward-compat handling for older HF records is genuine and well-done (`if col in df.columns` guards, None-filtering medians — kudos). Two real S1 issues prevent the new component gates from working as advertised on anything beyond the one currently-shipping `wan-t2v-1.3b-2gpu` benchmark, plus a CI exit-code regression that breaks the reseed skill's documented operator flow on fixed-threshold failures. Three S2s are worth addressing in this PR or a follow-up.

- S0 (blockers): 0
- S1 (must-fix): 2
- S2 (should-fix): 5 (3 surfaced below as persistent; 2 quality-of-life left in `review.md`)
- S3 (discussion): 4 — see archived `review.md`

---

## Findings

### [S1] `STAGE_METRIC_MAP` silently drops every model-specific denoising / encoding / decoding subclass

**What:** `fastvideo/tests/performance/test_inference_performance.py:25-30` maps four base class names by literal string:

```python
STAGE_METRIC_MAP: dict[str, str] = {
    "TextEncodingStage": "text_encoder_time_s",
    "DenoisingStage": "dit_time_s",
    "DmdDenoisingStage": "dit_time_s",
    "DecodingStage": "vae_decode_time_s",
}
```

The PR description states "only the four stage classes that map to component metrics are listed." That is correct for `wan-t2v-1.3b-2gpu` but incorrect for every other model already shipped in this repo. The following stages will be silently logged as `Unmapped stage` and produce `None` component times:

- `Cosmos25TextEncodingStage`
- `HYWorldDenoisingStage`, `CausalDMDDenosingStage`, `CausalDenoisingStage`, `MatrixGameCausalDenoisingStage`, `GameCraftDenoisingStage`, `LongCatDenoisingStage`, `LongCatI2VDenoisingStage`, `LongCatVCDenoisingStage`, `Gen3CDenoisingStage`, `CosmosDenoisingStage`, `Cosmos25DenoisingStage`, `Cosmos25T2WDenoisingStage`, `Cosmos25V2WDenoisingStage`, `Cosmos25AutoDenoisingStage`, `SD35DenoisingStage`
- `SD35DecodingStage`
- VAE-encoding stages on I2V pipelines

The only diagnostic is `logger.debug("Unmapped stage …")` — below the default INFO level and not surfaced in the perf report.

**Why it matters:** The intent of the PR (per-component gating) silently degrades to "no gating" the moment any subclassed stage runs. The component asserts at `fastvideo/tests/performance/test_inference_performance.py:255-264` only fire when `actual is not None`, so a missing metric trivially passes a configured threshold. Today's single benchmark is fine; the very next perf benchmark added (any cosmos / longcat / i2v / sd35 / matrixgame / hyworld variant) will silently produce `None` for the DiT or decode metric, write `None` to HF, and pass every component assertion with no warning.

**Suggested fix (pick one, in order of preference):**

1. **Map by base class via MRO walk, not by class-name string.** `add_stage_metric` already records `stage_class`; in `_extract_component_times`, walk the MRO chain and map the first matching base. This naturally covers every existing and future subclass.
2. **Cheaper, immediate fix:** add an explicit allow-list of every denoising / decoding / text-encoding subclass currently in the repo, upgrade `logger.debug` → `logger.warning` on unmapped stages, and add a unit test that walks `fastvideo/pipelines/stages/*.py` for `class \w+\(.*(?:Denoising|TextEncoding|Decoding)Stage\)` and asserts every match is in the map (or explicitly allow-listed).
3. **At minimum:** raise (not `log.debug`) when component thresholds are configured for a benchmark but the corresponding component time is `None`. Single-line change in `test_inference_performance.py:255-264`. This is the smallest mitigation that prevents the silent-pass.

**Evidence:** `fastvideo/tests/performance/test_inference_performance.py:25-30` (the map); `:255-264` (the `actual is not None`-gated assert).

### [S1] Modal `run_performance_tests` exit-code chain loses normalized artifacts on pytest failure

**What:** `fastvideo/tests/modal/pr_test.py:243-260` rewrites the modal CI tail as:

```bash
pytest ./fastvideo/tests/performance -vs;
PYTEST_RC=$?;
PERF_RC=0;
if [ $PYTEST_RC -eq 0 ]; then
  python ./fastvideo/tests/performance/compare_baseline.py;
  PERF_RC=$?;
fi;
python ./fastvideo/tests/performance/dashboard.py || true;
FINAL_RC=$PYTEST_RC;
if [ $FINAL_RC -eq 0 ]; then FINAL_RC=$PERF_RC; fi;
exit $FINAL_RC
```

The final exit-code resolution is correct — pytest failures preserve `PYTEST_RC` and don't get masked. **But:** `compare_baseline.py` is now skipped entirely when pytest fails, which means **no `normalized_perf_*.json` artifact is produced for any benchmark in that run, even ones unaffected by the pytest failure**. Rolling-baseline persistence (`_should_persist_tracking()`) is also skipped, freezing the rolling baseline silently on a benchmark-level pytest red.

**Why it matters:** The new `reseed-performance-baseline` SKILL.md (lines 152–160) explicitly tells the operator that "the performance CI exports normalized source results for failed rolling baseline comparisons when `compare_baseline.py` ran." Under the new chain, the operator can never reseed a baseline that just hit a fixed-threshold failure simultaneously, because the artifact is never produced. The skill's documented flow stops working the moment a fixed-threshold gate fires.

**Suggested fix:** Run `compare_baseline.py` unconditionally after pytest, capture its exit code, and only let `PERF_RC` decide the final exit when `PYTEST_RC == 0`:

```bash
pytest ./fastvideo/tests/performance -vs; PYTEST_RC=$?
python ./fastvideo/tests/performance/compare_baseline.py; PERF_RC=$?
python ./fastvideo/tests/performance/dashboard.py || true
if [ $PYTEST_RC -ne 0 ]; then exit $PYTEST_RC; fi
exit $PERF_RC
```

Or, if the current behavior is intentional, document the contract change in `SKILL.md` — operators currently expect the artifact to be available under fixed-threshold failures.

**Evidence:** `fastvideo/tests/modal/pr_test.py:243-260`; `.agents/skills/reseed-performance-baseline/SKILL.md:152-160`.

### [S2-persistent] Component thresholds defined only for L40S, no `default` fallback

**What:** The new threshold YAML schema in `fastvideo/tests/performance/baseline_thresholds.yaml` defines per-component limits only under the `l40s` device key. Benchmarks that run on a different GPU SKU (e.g. CI fan-out to other backends in the future) will hit `KeyError` or — depending on how the lookup is done — fall through to `None` and silently skip the gate.

**Why it matters:** This is the same silent-skip failure mode as S1-1: today everything runs on L40S so it doesn't trip, but the moment a new GPU SKU is added, the gate vanishes for it with no warning. Worth a `default:` block (mirroring the existing top-level total-time thresholds) or an explicit `raise` if the device key is missing when component thresholds are expected.

**Suggested fix:** Add `default:` keys at the same level as `l40s:` in `baseline_thresholds.yaml` (matching what already works for total thresholds), OR raise loudly when a benchmark configures component thresholds and the device key is unknown.

**Evidence:** `fastvideo/tests/performance/baseline_thresholds.yaml` (component-time keys L40S-only).

### [S2-persistent] `_extract_component_times` accumulator semantics fragile for future overlap

**What:** `_extract_component_times` accumulates by `+=` to handle the `DenoisingStage` + `DmdDenoisingStage` → `dit_time_s` overlap. The accumulator is correct for the current map (the two never legally appear in the same run), but the pattern silently does the wrong thing the moment a future stage map adds a real collision (e.g. multiple text-encoder stages — already plausible in compound pipelines).

**Why it matters:** The accumulator hides a precondition: "no two map entries with the same metric will fire in one run." That precondition is enforced nowhere. If it's violated, you'll get a silent sum that looks like a single component's time but isn't.

**Suggested fix:** Either replace `+=` with an explicit assert ("metric already populated") or document the precondition with a code comment and an assertion at the top of `_extract_component_times`.

**Evidence:** `fastvideo/tests/performance/test_inference_performance.py` (the `_extract_component_times` body).

### [S2-persistent] No unit tests for the stage parser, the backward-compat column branches, the TTL marker, or the per-metric regression loop

**What:** The PR's test surface is `py_compile` + `bash -n` + one manual buildkite run. None of the four new pieces of logic have a unit test:

- `_extract_component_times` (the stage→metric parser)
- The backward-compat `if col in df.columns` / `safe_float(r.get(key))` branches in `compare_baseline.py` and `dashboard.py`
- The `sync_from_hf(reuse_existing=True)` TTL marker
- The per-metric regression-detection loop in `compare_baseline.py`

**Why it matters:** Per the repo's own LESSONS pattern, cross-cutting infra adds (`fastvideo/tests/performance/*.py` and the dashboard pipeline) should ship with tests. Right now a regression in any of these four pieces is undetectable until the operator notices an empty dashboard or a missing reseed artifact.

**Suggested fix:** Add at least one unit test per piece. The stage parser is the highest-value target since S1-1 is exactly the failure mode it should catch.

**Evidence:** Diff lacks any `tests/` additions matching the new logic surfaces; `bash -n` + `py_compile` are not behavioral tests.

---

Satyam-53:
@SolitaryThinker  As per the latest review comment, there are no blockers, the s1 and s2 fixes are either intentional or for the broader coverage of different models, which will already be handled by the later robust perf. regression tracking enhancement work #1374 so we do not need to worry about them now.

SolitaryThinker (approved):

PR #1302, opened 2026-05-07
[feat]: Activate MagiHuman pipeline (registry + examples + SSIM) (8/8)

SolitaryThinker:
## Summary

**The activation switch.** After this PR merges, MagiHuman is publicly loadable:

```python
gen = VideoGenerator.from_pretrained("FastVideo/MagiHuman-Diffusers/base")
gen.generate_video(prompt="...", output_path="out.mp4", save_video=True)
```

## Changes

| File | LOC | Purpose |
|---|---:|---|
| `fastvideo/registry.py` | +153 | `register_configs` / `register_presets` for all 4 variants × 2 modes (8 entrypoints) |
| `examples/inference/basic/basic_magi_human*.py` | 322 (8 files) | User-facing scripts, one per variant × mode |
| `fastvideo/tests/ssim/test_magi_human_similarity.py` | 113 | CI-eligible SSIM regression against the umbrella HF repo |
| `.agents/memory/codebase-map/models/magi_human.md` | 77 (new) | Codebase-map entry; first per-model entry under `models/` (sets the convention) |
| `fastvideo/pipelines/basic/magi_human/AGENTS.md` | (modified) | Provenance section **finalized** with all 8 PR numbers + the `will/magi` source SHA |

## Verification

- `pre-commit run --files <changed paths>` ✓ (yapf, ruff, codespell, mypy, pymarkdown all green)
- The 8 example scripts use the same `VideoGenerator.from_pretrained("FastVideo/MagiHuman-Diffusers/<variant>")` pattern; no per-script duplication of pipeline wiring.
- Base T2V mp4 hash: should be `dcf5f2bf6534c7c0d91e7353e42b23db` (stable across all 43 commits of the original port).
- SSIM regression test is GPU-gated and runs on Modal CI per the existing `fastvideo/tests/ssim/` workflow.

## Stack context

Step **8 of 8** — the final PR in the decomposition. Stacked on:

```
main → #1293 (activation-trace, prereq)
       → #1294 (loader-infra, prereq)
         → #1295 (1/8 housekeeping)
           → #1296 (2/8 t5gemma)
             → #1297 (3/8 DiT)
               → #1298 (4/8 stages)
                 → #1299 (5/8 orchestrator)
                   → #1300 (6/8 provenance)
                     → #1301 (7/8 conversion)
                       → #THIS  (8/8 activate)
```

Once #1293 and #1294 merge to main, the entire magi stack auto-rebases. The 8 magi PRs can then squash-merge in order.

## Provenance

Source PR: #1280 (`will/magi` @ `4e1603634d27c8e1b5c4cc5d9387f046547f5c49`)

Total decomposition: **9,812 LOC** in the original PR → **10 stacked PRs + 2 prerequisite PRs** (PR-A activation-trace, PR-B loader-infra), each independently reviewable. After all merge: ~~14 of 14~~ parity tests bit-exact at the tip; mp4 hash preserved end-to-end.

### DIFF
--- a/.agents/memory/codebase-map/models/magi_human.md
+++ b/.agents/memory/codebase-map/models/magi_human.md
@@ -1,0 +1,77 @@
+# MagiHuman — Codebase Map Entry
+
+**Family:** daVinci-MagiHuman (joint audio-visual generative model)
+**Reference:** [GAIR-NLP/daVinci-MagiHuman](https://github.com/GAIR-NLP/daVinci-MagiHuman)
+**Architecture:** 15B-param single-stream DiT, 40 layers, hidden=5120,
+head_dim=128, GQA num_query_groups=8. Joint AV denoising in a unified token
+sequence; **no cross-attention**.
+
+## Variant Matrix
+
+| Variant | T2V | TI2V | DiT | Steps | CFG | Resolution |
+|---|---|---|---|---|---|---|
+| `base` | yes | yes | base | 32 | 2 | 480x256 |
+| `distill` | yes | yes | distill (DMD-2) | 8 | 1 (no CFG) | 480x256 |
+| `sr_540p` | yes | yes | base + sr_540p | 32 + 5 | 2 + cfg-trick | 896x512 |
+| `sr_1080p` | yes | yes | base + sr_1080p | 32 + 5 | 2 + cfg-trick | 1920x1056 |
+
+SR-1080p uses block-sparse video→video local-window attention on 32 of 40
+SR DiT layers (`frame_receptive_field=11`), implemented as a 3-block SDPA
+accumulator that mirrors upstream `flex_flash_attn_func`.
+
+## File Locations
+
+| Role | Path |
+|---|---|
+| Pipeline class | `fastvideo/pipelines/basic/magi_human/magi_human_pipeline.py` |
+| Pipeline package AGENTS.md | `fastvideo/pipelines/basic/magi_human/AGENTS.md` |
+| Stages | `fastvideo/pipelines/basic/magi_human/stages/*.py` |
+| DiT | `fastvideo/models/dits/magi_human.py` |
+| Text encoder (T5-Gemma) | `fastvideo/models/encoders/t5gemma.py` |
+| Audio VAE wrapper | `fastvideo/models/vaes/sa_audio.py` (shared with `stable_audio` pipeline) |
+| Conversion script | `scripts/checkpoint_conversion/convert_magi_human_to_diffusers.py` |
+| Examples | `examples/inference/basic/basic_magi_human*.py` (8 files, one per variant × mode) |
+| SSIM regression | `fastvideo/tests/ssim/test_magi_human_similarity.py` |
+| Local parity battery | `tests/local_tests/magi_human/` (14 tests, GPU-gated) |
+| Port journal | `fastvideo/pipelines/basic/magi_human/JOURNAL.md` |
+
+## Canonical HF Repo
+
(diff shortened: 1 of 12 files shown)


SolitaryThinker:
Rebased onto main (783be66f -> e5410611) to clear the registry.py conflict from main's flux2/matrixgame2+3 registry additions; pure union-resolve, no content changes (range-diff clean).

SolitaryThinker:
I reviewed exact head `e54106110e5f01153a199672eb08ce13d5a90b4a` against the add-model pipeline/SSIM gates. I am requesting changes for the following blockers.

1. **[P1] The four advertised TI2V entrypoints are not registry-routable.** Each canonical umbrella subfolder is registered only once, and config/preset lookup does not take `workload_type`. A normal `FastVideo/MagiHuman-Diffusers/{base,distill,sr_540p,sr_1080p}` call with `workload_type="i2v"` resolves to the T2V pipeline, T2V config, and T2V preset; `vae_config.load_encoder` remains false and the T2V reference-image hook is a no-op. The examples hide this by manually injecting an internal config and `override_pipeline_cls_name`, so the claimed 4 variants x 2 modes do not pass normal public dispatch. The detectors also key on `ti2v` while the actual entry classes use `I2V`, making model-index/local fallback ambiguous and order-dependent. Please make variant + workload resolution first-class and add exact registry tests for all eight combinations. Also remove `GAIR/daVinci-MagiHuman` from `hf_model_paths`: that upstream checkpoint has no root Diffusers `model_index.json` and requires conversion.

2. **[P1] The new SSIM test is deterministically broken after successful generation.** Its 100-character prompt ends in `.`, `VideoGenerator` sanitizes that to `...smile.mp4`, but `run_text_to_video_similarity_test` predicts `...smile..mp4`. Buildkite 4351 generated the video successfully and then failed only with `Video2 not found` on the double-dot path. Fix the shared helper at the root by reusing `fastvideo.eval.io.paths.sanitize_prompt` for reference lookup and generated output naming (including the sibling image path), and add a CPU regression for terminal punctuation.

3. **[P1] The claimed joint audio-video regression checks only video frames.** `compute_video_ssim_torchvision` cannot detect missing or corrupt audio even though the seeded reference contains AAC audio. Add a separate audio-quality assertion, or record an explicit accepted audio-regression deferral; stream presence alone would only be a minimal mux guard, not a quality gate. The current `min_acceptable_ssim=0.60` is also unsupported: this run never reached a score, while neighboring video gates are 0.90-0.98. Record the corrected exact-head score and choose a justified margin.

4. **[P1] Delete `.agents/memory/codebase-map/models/magi_human.md`.** Repository policy explicitly forbids static codebase maps under `.agents`; this file duplicates the package-local `AGENTS.md` and already drifted (it lists the DiT as lazy-loaded and says 1920x1056 while the config is 1920x1080).

Before activation, please attach non-skipped exact-head evidence that the four newly public TI2V/SR paths load and write readable joint-AV outputs. The only durable E2E evidence currently recorded is base T2V. The purported CPU preflight cannot supply this evidence: it fails constructing the DiT with `ValueError: Invalid attention backend for CPU`.

Nonblocking cleanup while revising: remove the broad `except Exception` around full-quality preset resolution (and do not advertise full-quality until that reference is seeded); update the stale SSIM seeding docstring; remove local-conversion instructions that the examples ignore (including the author-specific `/raid/william5lin...` path); and correct the base example's documented width from 448 to 480.

Positive evidence: all eight scripts parse/import, their explicit manual class/config pairings are internally consistent, the TI2V image asset exists, Modal discovers the intended model split, and the already-seeded default L40S reference is readable H.264 + AAC. No retry is useful until the deterministic filename bug is fixed.

PR #1244, opened 2026-04-21
[feat]: add LongCat bidirectional finetuning support

aryan5v:
## Summary

Adds initial LongCat-Video support for the modular Self-Forcing training path.

This PR introduces:
- `LongCatModel`, a training-side LongCat role wrapper
- `LongCatCausalModel`, a `CausalModelBase` adapter for chunk-wise self-forcing rollout using LongCat KV-cache accumulation
- an experimental LongCat self-forcing config under `examples/train/configs/distribution_matching/longcat/`
- a small Wan refactor so prompt-conditioning pipeline selection can be overridden by model family

## Related Issue

Refs #901

This implements the initial training-framework plumbing for "[Feature] Self Forcing Training with longcat-video".

I am using `Refs` instead of `Closes` because this has been smoke-tested locally but not yet validated in a full distributed LongCat training run. If maintainers consider the plumbing sufficient for the issue, I can update this to `Closes #901`.

## Motivation

FastVideo already has:
- `SelfForcingMethod` in the modular training framework
- LongCat inference support
- LongCat continuation / KV-cache logic

The missing piece was a LongCat training-model wrapper compatible with the modular `ModelBase` / `CausalModelBase` interfaces.

## Implementation Notes

- `LongCatModel` subclasses the existing Wan training wrapper pattern but loads `LongCatTransformer3DModel`.
- `LongCatCausalModel` implements `predict_noise_streaming()` and accumulates per-block LongCat KV caches across generated chunks.
- LongCat does not use VSA metadata in this path. `attn_kind="vsa"` is accepted because `SelfForcingMethod` passes it for student rollout, but LongCat sparse behavior is controlled by native BSA settings.
- The example config sets `flow_shift: 0`, matching LongCat's pipeline config.

## Testing

Local smoke tests run on macOS using Homebrew PyTorch Python:

- Parsed changed Python files with `ast.parse`
- Checked new files for ASCII and line length
- Loaded the new LongCat self-forcing YAML config
- Resolved targets:
  - `fastvideo.train.models.longcat.LongCatModel`
  - `fastvideo.train.models.longcat.LongCatCausalModel`
- Verified config resolves to `LongCatT2V480PConfig`
- Ran a fake-tensor runtime smoke test for:
  - timestep expansion
  - KV-cache initialization
  - KV-cache accumulation across chunks
  - `num_cond_latents` propagation
  - tuple-return guard

Full distributed GPU training has not been run yet.

## Known Limitations

[...]

PR #1307, opened 2026-05-08
[feat]: add LongCat causal self-forcing scaffold

aryan5v:
Draft follow-up for causal / self-forcing LongCat work. The bidirectional LongCat fine-tuning base from #1244 has merged, and this branch is focused on causal DFSFT / self-forcing support.

Current contents:
- `LongCatCausalModel` training wrapper
- LongCat causal DMD and multi-step inference pipelines/stages
- `self_forcing_t2v.yaml` config
- KV-cache buffering, RoPE temporal-offset plumbing, and block-causal LongCat self-attention mask support
- Regression coverage for causal masking, chunk/block mismatch, chunked rollout vs full-prefix behavior, and `forward_with_kv_cache`

Validation completed:
- Targeted LongCat causal / KV-cache regression tests passed.
- 1-step training smoke on 4x B200 passed.
- 5-step training smoke on 4x B200 passed.
- 21-frame causal rollout with `chunk_size=3` passed.
- 100-step self-forcing run on 4x B200 with real preprocessed video/text data completed.
- Student and critic optimizer updates ran.
- Validation video generation ran at steps 0/50/100.
- W&B evidence: https://wandb.ai/aryan5v-san-jose-state-university/self-forcing-longcat/runs/3k7mqmqx
- CI fastcheck / targeted Buildkite contexts passed on the PR branch.
- Local `python3 -m py_compile ...` passed for changed LongCat files.
- Local `git diff --check` passed.

Remaining notes:
- The generated validation videos are not visually coherent yet, so follow-up quality work may still be needed after the scaffold is reviewed.
- PR still needs reviewer approval and required merge checks.

Linked base PR: #1244

### DIFF
--- a/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
+++ b/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
@@ -85,4 +85,6 @@ callbacks:
     max_grad_norm: 1.0
 
 pipeline:
-  flow_shift: 0
+  # Match the released LongCat scheduler config. flow_shift=0.0 collapses
+  # FlowMatch training timesteps to zero in FastVideo's scheduler.
+  flow_shift: 12.0
--- a/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
+++ b/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
@@ -78,6 +78,7 @@ training:
     run_name: longcat_self_forcing
 
   model:
+    # LongCat student rollout uses checkpoint-safe KV cache snapshots.
     enable_gradient_checkpointing_type: full
 
 callbacks:
--- a/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
+++ b/examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml
@@ -1,0 +1,88 @@
+# Self-Forcing distillation: LongCat-Video T2V 13.6B
+#
+# Experimental scaffold for chunk-wise LongCat self-forcing training in the
+# modular framework. The student uses KV-cache accumulation between chunks.
+
+models:
+  student:
+    _target_: fastvideo.train.models.longcat.LongCatCausalModel
+    init_from: FastVideo/LongCat-Video-T2V-Diffusers
+    trainable: true
+  teacher:
+    _target_: fastvideo.train.models.longcat.LongCatModel
+    init_from: FastVideo/LongCat-Video-T2V-Diffusers
+    trainable: false
+    disable_custom_init_weights: true
+  critic:
+    _target_: fastvideo.train.models.longcat.LongCatModel
+    init_from: FastVideo/LongCat-Video-T2V-Diffusers
+    trainable: true
+    disable_custom_init_weights: true
+
+method:
+  _target_: fastvideo.train.methods.distribution_matching.self_forcing.SelfForcingMethod
+  rollout_mode: simulate
+  generator_update_interval: 5
+  real_score_guidance_scale: 1.0
+  dmd_denoising_steps: [1000, 750, 500, 250]
+
+  chunk_size: 3
+  student_sample_type: sde
+  context_noise: 0.0
+  enable_gradient_in_rollout: true
+  start_gradient_frame: 0
+
+  fake_score_learning_rate: 1.0e-5
+  fake_score_betas: [0.0, 0.999]
+  fake_score_lr_scheduler: constant
+
+training:
+  model_path: FastVideo/LongCat-Video-T2V-Diffusers
+
+  distributed:
+    num_gpus: 8
+    sp_size: 1
+    tp_size: 1
+    hsdp_replicate_dim: 1
+    hsdp_shard_dim: 8
+
+  data:
+    data_path: data/LongCat-Syn
+    dataloader_num_workers: 4
+    train_batch_size: 1
+    training_cfg_rate: 0.0
+    seed: 1000
+    num_latent_t: 20
+    num_height: 480
+    num_width: 848
+    num_frames: 77
+
+  optimizer:
+    learning_rate: 1.0e-5
+    betas: [0.0, 0.999]
+    weight_decay: 0.01
+    lr_scheduler: constant
+    lr_warmup_steps: 0
+
+  loop:
+    max_train_steps: 4000
+    gradient_accumulation_steps: 1
+
+  checkpoint:
+    output_dir: outputs/longcat_self_forcing
+    training_state_checkpointing_steps: 1000
+    checkpoints_total_limit: 3
+
+  tracker:
+    project_name: self-forcing-longcat
+    run_name: longcat_self_forcing
+
+  model:
+    enable_gradient_checkpointing_type: full
+
+callbacks:
+  grad_clip:
+    max_grad_norm: 1.0
+
+pipeline:
+  flow_shift: 0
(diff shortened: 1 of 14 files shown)


aryan5v:
Validation update:

I tested this on the B200 machine and the core path is now running end to end.

Passed so far:
- targeted LongCat causal/KV-cache regression tests
- 1-step training smoke on 4x B200
- 5-step training smoke on 4x B200
- 21-frame causal rollout with chunk_size=3
- 100-step run on 4x B200 with real preprocessed video/text data
- student + critic optimizer updates
- validation video generation at steps 0/50/100

W&B: https://wandb.ai/aryan5v-san-jose-state-university/self-forcing-longcat/runs/3k7mqmqx

The validation videos are not visually coherent yet, so this validates the end-to-end path, not final quality.

aryan5v:
@codex review

alexzms:
Hi @aryan5v Thanks for contributing! By the way do you have capacity to do the E2E validation work you flagged (a short DFSFT/self-forcing training attempt + generation sanity check + attaching the run evidence)? We might need the e2e workflow to run in order to validate the correctness of the code.

If hardware is the blocker on your side, happy to help — I can run it on my end and attach the evidence here, then we can iterate on any small fixes that surface.

Would you be open to me pushing commits directly to your branch? That way I can help patch things up quickly as they come up.

aryan5v:
> Hi @aryan5v Thanks for contributing! By the way do you have capacity to do the E2E validation work you flagged (a short DFSFT/self-forcing training attempt + generation sanity check + attaching the run evidence)? We might need the e2e workflow to run in order to validate the correctness of the code.
> 
> If hardware is the blocker on your side, happy to help — I can run it on my end and attach the evidence here, then we can iterate on any small fixes that surface.
> 
> Would you be open to me pushing commits directly to your branch? That way I can help patch things up quickly as they come up.

Thanks @alexzms , I did run the E2E pass [here](https://github.com/hao-ai-lab/FastVideo/pull/1307#issuecomment-4426099565).

It includes the short self-forcing run, optimizer updates, validation generation, and [W&B evidence](https://wandb.ai/aryan5v-san-jose-state-university/self-forcing-longcat/runs/3k7mqmqx)

The caveat is that the generated videos are not visually coherent yet, so I’d frame it as end-to-end execution validation, not final quality validation.

And yes, feel free to push commits directly to the branch.

alexzms:
Sorry for the late response. I'm going to re-run the LongCat causal self-forcing training on top of your branch using my overfit dataset to validate it end-to-end.

alexzms:
Thank you for your patience! **I wrote** a small compatibility fix for `longcat_causal.py` **and ran** a single-sample **DFSFT overfit** on LongCat-Video-T2V (13.6B, 7× H200, `causal_block_size = chunk_size = 3`) to validate the causal training path — and it **works great**: the validation videos (generated via `LongCatCausalPipeline` causal rollout, 6 blocks × 50 steps) visibly converge to the ground-truth clip as training proceeds. Nice work on the causal scaffold.

I pushed the fix directly to this branch (commit `4c5d94cb`) since `longcat_causal.py` should already support DFSFT:

**Problem:** `LongCatCausalModel` only overrode `predict_noise_streaming` (self-forcing rollout). The non-streaming training `predict_noise` — used by `FineTuneMethod` and `DiffusionForcingSFTMethod` — fell back to the inherited `LongCatModel.predict_noise`, whose `_build_distill_input_kwargs` leaves `causal_block_size` unset. The LongCat DiT only applies its block-causal mask when that kwarg is passed, so **DFSFT silently trained with full bidirectional attention** — no error, because the chunk-size guard reads `transformer.num_frame_per_block`, which was also unset (→ `None` → guard skipped).

**Fix (in `LongCatCausalModel`, +29 lines):**
1. Override `_build_distill_input_kwargs` to inject `causal_block_size` into the training forward, so `predict_noise` → transformer applies the same block-causal mask the streaming/self-forcing paths use.
2. Expose `transformer.num_frame_per_block = causal_block_size` so DFSFT's chunk-size guard validates instead of silently passing.

Separately (not in this commit — it's a preprocessing concern): LongCat VAE/T5 latents are bf16 and `numpy` can't serialize them, so a LongCat preprocess/overfit recipe needs a bf16→fp32 upcast in the latent saver.

alexzms:
I think the only thing before the PR can be merged is testing of self forcing

SolitaryThinker (changes requested):
Reviewed head 4c5d94cb2df50a02bffd6d3bb11c7c32f3b8202b. Requesting changes for the following independently verified blockers.

1. **[P1] Preserve prompt conditioning when refreshing K/V.** The training cache write at fastvideo/train/models/longcat/longcat_causal.py:169-185 and both inference cache writes pass skip_crs_attn=True; training also substitutes zero text embeddings. LongCat blocks run self-attention, then cross-attention, then FFN, so from the second layer onward the K/V projected during cache refresh comes from a different, text-free hidden state than the conditional forward. Later chunks therefore cannot match the conditional block-causal model. The new parity test hides this by disabling cross-attention on both the chunked and reference calls. Cache refresh must use the real prompt/mask and preserve cross-attention, with a nonzero-text chunked-vs-full regression.

2. **[P1] Use one causal partition for the shipped 20/3 configuration.** examples/train/configs/distribution_matching/longcat/self_forcing_t2v.yaml sets num_latent_t=20 with chunk_size=causal_block_size=3. build_longcat_causal_block_ranges produces [0,5), [5,8), ... while build_longcat_block_causal_mask and DFSFT partition from frame zero as [0,3), [3,6), ... [18,20). Each cache boundary after the first splits a nominal bidirectional block, so the streaming rollout is not equivalent to the training mask. The parity test covers only divisible 6/2; the reported 21-frame run is also divisible and cannot exercise this default. Align rollout, mask, and DFSFT boundaries, then add actual 20/3 parity.

3. **[P1] Make exported causal checkpoints select a causal pipeline.** The training exporter copies the base model_index.json unchanged, so its _class_name remains LongCatPipeline. EntryClass only makes the two new causal classes discoverable; it does not select either one. Loading an exported causal student through the normal VideoGenerator path therefore silently runs bidirectional inference. Add the causal registry/config/preset or export metadata needed for an unambiguous load path, plus a checked-in inference example.

4. **[P1] Wire and exercise the DMD pipeline from the checked-in config.** LongCatCausalDMDDenoisingStage constructs timesteps from pipeline_config.dmd_denoising_steps, whose LongCat default is None. The new YAML places those steps only under method and has no validation callback to propagate sampling_timesteps, so the supplied config never exercises this pipeline and a default explicit invocation reaches torch.tensor(None). Add the causal validation wiring and a non-skipped pipeline smoke/parity test. The current primitive-only tests do not satisfy the model-port pipeline/final gate, especially while the PR reports incoherent generated media.

5. **[P2] Do not retain per-layer GPU RoPE tables outside offload ownership.** rotary_embedding_3d.py:228-247 stores runtime-device tables in a plain OrderedDict on each of 48 blocks. These tensors are neither parameters nor managed by the default layerwise-offload hook, so a 480x848 T=20 causal pass retains tables for lengths 5, 8, 11, 14, 17, and 20 across every block, about 2.7 GiB, and varying shapes can fill 16 entries per block. Bound/share the device cache in a way that layerwise offload can actually reclaim.

6. **[P2] Fix the DMD batch contract.** At longcat_causal_dmd_denoising.py:127-169, t_expand has B elements but pred_noise_to_pred_video receives B*block_frames flattened samples; for B>1 it asserts. The next-noise path also collapses a supported generator list to generator[0], ignoring per-sample RNGs. Cover B>1 with distinct generators.

Local static evidence: git diff --check and Python compilation pass. Targeted pytest could not run in this checkout because the available local Python has no torch. The current Buildkite failure is unrelated runner disk exhaustion during checkout, so it provides no code signal; please rerun fastcheck on a clean worker after these blockers are addressed.

PR #1320, opened 2026-05-11
[feat] eval: async VideoPool + metric streamlines

shaoxiongduan:
# Async VideoPool + GPU-side common metrics + safe optical-flow chunks

## Summary

- New `VideoPool`: async path → tensor prefetcher behind the
  `Evaluator`. One pool is built per `evaluate(samples=…)` call;
  workers pull from a bounded queue while loader threads decode the
  next clip — decode is hidden behind compute.
- `EvalWorker` pre-uploads `video` / `reference` once per sample so the
  per-sample metric loop shares a single GPU-resident tensor instead of
  every metric paying its own host→device transfer. Toggle with
  `Evaluator(pre_upload=…)`.
- `common.{psnr, ssim}` move to GPU (`needs_gpu=True`) — at 1080p the
  CPU paths were memory-bandwidth-bound and competed with the loader.
- Memory-safe chunking: `common.lpips` adds `chunk_size=8` (was peaking
  ~60 GB on a 121-frame clip → ~5 GB); `optical_flow.{gt,synthetic}`
  drop `chunk_size` to 1 (was OOMing on H200; DPFlow's cost volume is
  ~4 GB per frame pair at 1080p).
- Contract tightening: metrics receive one sample per `compute()`
  call. List/B-dim unwrap shims removed from `physics_iq` and the
  affected `vbench` metrics. `physics_iq.utils.prepare_pair_batch` →
  `prepare_pair` (singular).

Scores are bit-identical across baseline and pool in every benchmarked
configuration.

## End-to-end speedup

All tests are done with 121-frame 1080p videos on a single H200.

| group | N | baseline (s) | pool (s) | speedup |
|---|---:|---:|---:|---:|
| common (psnr+ssim+lpips) | 4 | 18.27 | 14.94 | **1.22×** |
| common (psnr+ssim+lpips) | 16 | 75.33 | 57.64 | **1.31×** |
| vbench (6 metrics) | 4 | 20.68 | 17.87 | **1.16×** |
| vbench (6 metrics) | 16 | 82.54 | 63.84 | **1.29×** |
| optical_flow.gt | 4 | 414.70 | — | — *(compute-bound; pool can't hide ~100 s/sample)* |

## Peak GPU memory

Changed chunk size for optical flow metric to 1 to save memory on large videos.

| metric set | peak GPU (GB) | notes |
|---|---:|---|
| common (psnr + ssim + lpips) | 15.3 | lpips pre-chunk peak was 60+ GB on this clip |
| vbench (6 metrics) | 9.9 | |
| optical_flow.gt_optical_flow | 14.6 | pre-fix: 63 GB → OOM on H200 |

### DIFF
--- a/fastvideo/eval/evaluator.py
+++ b/fastvideo/eval/evaluator.py
@@ -3,19 +3,21 @@
 Layering (mirrors FastVideo's VideoGenerator → Worker pattern, but
 in-process)::
 
-    Evaluator               ← user-facing; round-robins samples across workers
+    Evaluator               ← user-facing
       └── EvalWorker × N    ← single-GPU; owns metric replicas
+      └── VideoPool         ← async path-→-tensor prefetch (per evaluate call)
 
 The constructor builds one :class:`EvalWorker` per GPU and loads every
 metric on every worker eagerly. :meth:`evaluate` is the single entry
 point: pass kwargs for one sample, or pass a list of sample dicts to
-fan-out across GPU replicas — same method, return type follows the
-input shape.
+fan-out across GPU replicas with pipelined decoding — same method,
+return type follows the input shape.
 """
 from __future__ import annotations
 
-from concurrent.futures import ThreadPoolExecutor
+import threading
 from collections.abc import Iterable
+from typing import Any
 
 from fastvideo.eval.registry import (list_metrics, missing_dependencies, resolve_group)
 from fastvideo.eval.types import MetricResult
@@ -38,6 +40,22 @@ class Evaluator:
         Number of GPU replicas. Each gets its own :class:`EvalWorker`.
     compile : bool
         Apply :func:`torch.compile` to each metric's ``_model``.
+    loader_threads : int
+        Background decode threads in the :class:`VideoPool`. Default 1
+        (hide decode behind compute). Bump for I/O-heavy benchmark sets
+        where one loader can't keep up with the workers.
+    prefetch_factor : int
+        ``pool max_size = prefetch_factor * num_workers``. Default 2 —
+        one sample being consumed, one prefetched per worker.
+    pre_upload : bool
+        When ``True`` (default), the worker performs a single
+        host→device upload of ``video`` / ``reference`` per sample
+        before the metric loop, and every metric reads from that
+        shared GPU-resident tensor. Without it, each metric pays its
+        own ``.to(self.device)`` — N transfers of the same clip for N
+        metrics, which dominates at high resolution. Set ``False`` for
+        training-time eval, where keeping a clip resident on GPU
+        across the metric loop would fight the training step for VRAM.
     """
 
     def __init__(
@@ -46,13 +64,20 @@ class Evaluator:
         device: str = "cuda:0",
         num_gpus: int = 1,
         compile: bool = False,
+        *,
+        loader_threads: int = 1,
+        prefetch_factor: int = 2,
+        pre_upload: bool = True,
     ) -> None:
         names = _resolve_metric_names(metrics)
         if num_gpus > 1:
-            self._workers = [EvalWorker(names, f"cuda:{i}", compile=compile) for i in range(num_gpus)]
+            self._workers = [
+                EvalWorker(names, f"cuda:{i}", compile=compile, pre_upload=pre_upload) for i in range(num_gpus)
+            ]
         else:
-            self._workers = [EvalWorker(names, device, compile=compile)]
-        self._pool = (ThreadPoolExecutor(max_workers=num_gpus) if num_gpus > 1 else None)
+            self._workers = [EvalWorker(names, device, compile=compile, pre_upload=pre_upload)]
+        self._loader_threads = max(1, loader_threads)
+        self._prefetch_factor = max(1, prefetch_factor)
 
     @property
     def num_gpus(self) -> int:
@@ -71,17 +96,16 @@ class Evaluator:
 
         ``video`` and ``reference`` may be either a pre-loaded
         ``(T, C, H, W)`` tensor or a path-like (``str`` / ``Path``).
-        Paths are decoded inside the worker thread that picks up the
-        sample, so memory stays bounded by ``num_gpus`` even when
-        thousands of paths are queued — see ``score_folder.py`` for
-        the canonical pattern.
+        Paths in the list form are decoded asynchronously by a
+        :class:`VideoPool` that runs alongside metric compute, hiding
+        decode latency behind GPU work.
 
         One sample::
 
             ev.evaluate(video=tensor, text_prompt="...", fps=24.0)
             ev.evaluate(video="path/to/clip.mp4", fps=24.0)
 
-        Many samples — fan out across GPU replicas, results in input order::
+        Many samples — pipelined decode + work-stealing across replicas::
 
             ev.evaluate(samples=[
                 {"video": "a.mp4", "reference": "ref_a.mp4"},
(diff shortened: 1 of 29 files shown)


fastvideo/eval/pool.py
```diff
@@ -0,0 +1,159 @@
...
+        Returns ``None`` when all input samples have been consumed.
+        Re-raises any exception caught in a loader thread on the
+        consumer's stack so callers don't hang on a dead loader.
+        Thread-safe: multiple consumer threads may share one pool.
+        """
+        with self._consume_lock:
+            if self._consumed >= len(self._samples):
+                return None
+        try:
+            item = self._ready_q.get(timeout=timeout)
```
SolitaryThinker: Deadlocks when `len(samples) < num_workers`. The `_consumed` check above runs once, then `_ready_q.get(timeout=None)` blocks forever. With N consumer threads and S < N samples, the S that win the race exit cleanly; the remaining N − S sleep on this line, and `Evaluator._run` then hangs in `t.join()`. Reproduced with `samples=[one], num_gpus=2`.

Two workable fixes:
- Push `n_workers` extra sentinels onto `_ready_q` once the loaders are done, so every consumer gets exactly one wake-up.
- Change `get()` to loop with a short timeout and re-check `_consumed` between waits.

Probably wasn't caught by the benchmark runs because they all had `len(samples) >= num_gpus`, but a tail batch on a real multi-GPU eval would hit this.

fastvideo/eval/metrics/vbench/color/metric.py
```diff
@@ -55,17 +55,11 @@ def compute(self, sample: dict) -> MetricResult:
         from fastvideo.eval.metrics.vbench._grit_helper import prepare_frames
 
         video = sample["video"]  # (T, C, H, W)
-        aux = sample.get("auxiliary_info")
-        if isinstance(aux, list):
-            aux = aux[0] if aux else None
-        if not aux or "color" not in aux:
+        aux = sample.get("auxiliary_info") or {}
```
SolitaryThinker, line 58: This silently skips every sample that came through `build_eval_kwargs`. `fastvideo/eval/io/paths.py:62` still does `kwargs["auxiliary_info"] = [aux]`, so `sample.get("auxiliary_info") or {}` resolves to the list `[{"color": "red"}]`, and `"color" not in [{"color": "red"}]` is `True`. The `_skip` branch fires for every sample and the run still looks successful.

Same pattern in `multiple_objects`, `object_class`, `scene`, and `spatial_relationship`. Either restore the `isinstance(aux, list)` unwrap or update `build_eval_kwargs` (and the example/CLI sites that mirror the wrap) to pass scalars in this PR.

fastvideo/eval/metrics/vbench/human_action/metric.py
```diff
@@ -101,8 +101,6 @@ def compute(self, sample: dict) -> MetricResult:
         text_prompt = sample.get("text_prompt")
         if text_prompt is None:
             return self._skip(sample, "missing text_prompt with action labels")
```
SolitaryThinker, line 103: Removing this unwrap turns into a hard crash at L122 (`gt_label = text_prompt.lower().strip()`) when the sample came in through `build_eval_kwargs` — `paths.py:59` does `kwargs["text_prompt"] = [row["prompt"]]`, so `text_prompt` is a list and `list.lower()` doesn't exist. `appearance_style.compute` and `overall_consistency.compute` will hit the same thing via `clip.tokenize([[prompt]])`.

Keeping the `isinstance(text_prompt, list): text_prompt = text_prompt[0]` unwrap is the smallest fix. The bigger fix is to flip `build_eval_kwargs` (and the example scripts + `fastvideo/entrypoints/cli/eval.py`) to pass scalars in the same PR.

fastvideo/eval/evaluator.py
```diff
@@ -66,46 +91,101 @@ def evaluate(
...
-            return self._workers[0].evaluate(**kwargs)
+        single = samples is None
+        sample_list: list[dict] = [kwargs] if samples is None else list(samples)
+        if not sample_list:
+            return EvalResults(samples=[], corpus={})
+
+        per_sample, corpus = self._run(sample_list)
+
+        if single:
+            return per_sample[0]
```
SolitaryThinker, line 136: `corpus` falls on the floor in the single-sample path. `_run` always finalizes any set metric and returns its corpus dict, but we discard it here when `single` is true.

Latent today since no registered metric sets `is_set_metric=True`, but the moment someone wires up a corpus-level metric (FAD, FVD-style) via the kwargs form, they'll silently get zero back with no signal. Either route this through `EvalResults` too, or raise when a set metric is loaded and the kwargs form is used.

fastvideo/eval/worker.py
```diff
@@ -55,29 +63,40 @@ def _load(self) -> None:
...
         """
         if self._unloaded:
             raise RuntimeError("EvalWorker was unloaded; call reload() before evaluating.")
 
         sample = dict(kwargs)
-        sample["video"] = _resolve_video_input(sample.get("video"))
-        if "reference" in sample:
-            sample["reference"] = _resolve_video_input(sample["reference"])
+        if self._pre_upload:
+            sample["video"] = _to_device(sample.get("video"), self._device)
```
SolitaryThinker, line 82: `pre_upload=True` is unconditional here, but `physics_iq` is CPU-only — `tensor_to_uint8_frames` immediately does `.cpu().float().clamp(0,1).permute(...).numpy()` on the input. So a mixed metric set like `["common.psnr", "physics_iq"]` pays a full host→device→host transfer of the input video per sample. Cheap on small clips, less cheap at 1080p × 121 frames.

Could gate on whether any loaded metric actually needs the device (e.g. a `needs_gpu` flag on `BaseMetric`) and skip the upload when none of them do.

Separate small thing: `sample.get("video")` here turns a missing-video into `None`, which then surfaces downstream as `AttributeError: 'NoneType' object has no attribute 'float'` instead of `KeyError: 'video'`. Cosmetic but slightly worse error UX.

SolitaryThinker (commented):
Read through the eval refactor. The async `VideoPool`, the GPU-resident PSNR/SSIM moves, LPIPS chunking, and the corpus/set-metric scaffolding all look reasonable. Two things worth fixing before merge, plus a few smaller notes left inline.

The first is the bigger one. The unwrap shims got removed from `physics_iq` and the vbench metrics, but `build_eval_kwargs` in `fastvideo/eval/io/paths.py` still wraps `text_prompt` and `auxiliary_info` in single-element lists — and that helper is what `eval_ltx2_vbench.py`, `basic_ltx2_eval.py`, and the `fastvideo eval run` CLI all go through. After this PR, samples routed through that helper either crash (`human_action.compute` → `AttributeError` on `text_prompt.lower()`; `appearance_style` / `overall_consistency` → nested list into `clip.tokenize`) or silently skip (`color`, `multiple_objects`, `object_class`, `scene`, `spatial_relationship` — `"color" not in [aux_dict]` is always `True`). The silent skip is the one that worries me — runs look healthy and every score is zero. Easiest path is to put the unwraps back; alternatively, update `build_eval_kwargs` plus the example/CLI call sites to pass scalars in this same PR.

The second is `VideoPool` deadlocks when `len(samples) < num_workers`. `get()` short-circuits on `_consumed >= len(samples)`, but only before the queue wait — so the extra consumers pass the check and then block on `_ready_q.get(timeout=None)` forever after the loaders exit. Reproduced locally with 1 sample / 2 workers. Push `n_workers` extra sentinels onto `_ready_q` once samples are exhausted, or change `get()` to loop with a short timeout re-checking `_consumed`. Probably didn't show up in benchmarks because every run had `len(samples) >= num_gpus`.

The rest is smaller stuff inline — set-metric `corpus` is dropped on the kwargs path, `pre_upload=True` round-trips through GPU for CPU-only metrics, and the new `Video` public export has no reader in the metric path yet.

Happy to chat through any of these.

SolitaryThinker (approved):

PR #1352, opened 2026-05-15
[feat] eval: add audio metrics

shaoxiongduan:
## Purpose

Add seven `audio.*` metrics to `fastvideo.eval`, partially ported from
`hkchengrex/av-benchmark` (the V2A literature's de-facto eval harness used
by MMAudio, FoleyCrafter, V2A-Mapper). Closes the audio gap in the eval
suite so video-with-audio models like LTX2 can be scored end-to-end through
the same registry-driven API as `common.*`, `vbench.*`, and `physics_iq.*`.

Fixes #

## Changes

- New metrics under `fastvideo/eval/metrics/audio/`:
  - `audio.clap_score` (HF `ClapModel`, `laion/clap-htsat-fused`)
  - `audio.audiobox_aesthetics` (Meta `audiobox_aesthetics`, PQ as score)
  - `audio.kl_divergence` (PaSST AudioSet-527 logits, `KL(gt || pred)`)
  - `audio.frechet_distance` (PaSST 768-d embeds, corpus-vs-corpus FAD)
  - `audio.wer` (Whisper-base default; GLM-ASR and SenseVoice optional)
  - `audio.desync` (Synchformer, av-benchmark `DeSync`)
  - `audio.imagebind_score` (ImageBind huge, audio/video cosine)
- New `[eval-audio]` extra in `pyproject.toml`; ImageBind is git-sourced via
  `[tool.uv.sources]` (CC BY-NC-SA 4.0 so kept out of the FastVideo tree).
- Vendored upstream under `fastvideo/third_party/eval/` alongside the
  existing VBench submodule:
  - `synchformer/` (MIT) for `audio.desync`
  - `glmasr/` (Apache-2.0) for `audio.wer (glm_asr)` on transformers 4.57
- Thread-safety guards for the multi-GPU pool path:
  - `decord.bridge.set_bridge("torch")` re-set on the worker thread in
    `imagebind_score.compute()` (decord's bridge is `threading.local` with
    a missing `global` upstream; setting on the main thread is lost).
  - `threading.Lock` around pytorchvideo's decord-backed video decode.
  - `contextlib.redirect_stdout(devnull)` removed from FAD and KL setup
    paths (races across workers and closes `sys.stdout`).
- Numerical robustness in `audio.frechet_distance`: filter non-finite PaSST
  rows before `np.cov`, surface drop counts in `details`, always-on
  `eps * I` regularization on both covariances.
- Three equivalent ways to supply the FAD reference set: per-sample paired
  (`reference_audio` on each sample, matches V2A manifests), role-tagged
  samples (`role="reference"`), or a cached `.pt` features file pointed to
  by `FASTVIDEO_FAD_REF_FEATURES`.
- Eval README and `docs/contributing/eval-metrics.md` updated:
  install matrix, per-metric input contracts, vendored-tree disclosure,
  upstream-citation table, three-pattern upstream-wrapping guide (submodule
  / vendored / git-source-via-uv).
- LTX2 audio example: `examples/inference/eval/basic_ltx2_audio_eval.py`.

## Test Plan

```bash
# Lint
pre-commit run --from-ref origin/main --to-ref HEAD

# 7 metrics x 9 corruption variants of one LTX2 clip on 4 H200s
srun --gpus=4 --gres=gpu:4 python \
  outputs_video/ltx2_audio_eval/_score_ltx2_corruptions.py --num-gpus 4

# Single-GPU sanity on real movie clip + LTX2 video
srun --gpus=1 --gres=gpu:1 python \
  outputs_video/ltx2_audio_eval/_score_all_audio.py
```

Variants exercised by the corruption script: original LTX2 audio, silence,
white noise, low volume (-20 dB), reversed audio, pitch-shifted +6
semitones, audio delayed +1 s, audio advanced -1 s, audio replaced with an
unrelated Inception clip.

## Test Results

Pre-commit: yapf, ruff, codespell, pymarkdown, mypy, filename, suggestion
all pass on the 5-commit diff.

End-to-end 4-GPU eval on 9 corruption variants of the LTX2 clip
(eval wall time ~45 s for `evaluate()` plus ~104 s of model loading at
ctor; per-sample ~5 s averaged across 4 workers):

| Variant | clap | audiobox PQ | KL | WER | desync s | imagebind |
|---|---:|---:|---:|---:|---:|---:|
| original           | 0.187 | 7.56 | 0.00  | 0.00  | 0.20 | 0.380 |
| silence            | NaN   | 6.88 | (skip) | 1.00 | 2.00 | 0.006 |
| white_noise        | 0.071 | 4.91 | 5.01  | 1.00  | 1.40 | 0.080 |
| low_volume         | 0.139 | 7.37 | 0.12  | 0.00  | 0.20 | 0.366 |
| reversed           | 0.277 | 6.93 | 0.25  | 1.00  | 0.80 | 0.370 |
| pitch_up           | 0.184 | 6.55 | 0.45  | 0.78  | 0.20 | 0.311 |
| desync_delay_1s    | 0.128 | 8.11 | 0.39  | 0.28  | 1.20 | 0.344 |
| desync_advance_1s  | 0.101 | 8.04 | 0.17  | 0.22  | 0.80 | 0.371 |
| unrelated_audio    | -0.098 | 6.59 | 3.97 | 1.00  | 2.00 | 0.070 |

Corpus FAD across the 9 variants paired against the original: 739.389
(silence drops 1 non-finite PaSST embed, surfaced in
`details.n_gen_dropped_nonfinite`).

<details>
<summary>How to read the table</summary>

`clap` rewards timbre/category match with the prompt and ignores temporal
order, so reversed speech can score higher than the original. `audiobox`
(PQ) measures audio quality in isolation and is amplitude-invariant.
`KL` and `imagebind` correctly drop toward zero on noise/silence/unrelated
audio. `desync` lands at the Synchformer ±2 s grid's nearest match to the
ground-truth shift (0.2 s for original, 1.2/0.8 s for the 1 s shifts, 2.0 s
for silence and unrelated audio). `wer` reaches 1.0 on non-speech and ~0.78
on pitch-shifted speech where Whisper recovers some words.

</details>

## Checklist

- [ ] I verified SSIM regression tests pass (not a model/pipeline change)
- [ ] I updated the support matrix if adding a new model (not adding a model)


fastvideo/eval/metrics/audio/desync/metric.py
```diff
@@ -0,0 +1,224 @@
...
+            self.setup()
+
+        video = sample.get("video")
+        audio_path = sample.get("audio")
+        if video is None or audio_path is None:
+            return self._skip(sample, "missing 'video' or 'audio'")
+
+        # Video preprocessing → (S, T_seg, C, 224, 224)
+        frames = video.float().to(self.device)
+        src_fps = self._src_fps if self._src_fps is not None else _SYNC_FPS
```
SolitaryThinker: This ignores the per-sample `fps` that callers can pass through `Evaluator.evaluate(...)` and falls back to 25 fps for all pool-decoded clips. For 24/30/8 fps inputs, `_resample_video` computes the wrong clip duration and segment positions while the audio waveform stays at native duration, so the video/audio windows no longer line up. Please prefer `sample.get("fps")` when present, or skip when neither the sample nor constructor provides the source FPS.

fastvideo/eval/metrics/audio/desync/metric.py
```diff
@@ -0,0 +1,224 @@
...
+        # Audio preprocessing → Synchformer audio features (1, S, t_a, D)
+        waveform = _load_audio_waveform(audio_path, self._resample_cache).to(self.device)
+        assert self._mel is not None
+        afeats = _encode_audio_segments(self._model, waveform, self._mel)
+
+        # Compare first-14 and last-14 segments; argmax over the 21-class grid;
+        # |grid|-value → average across the two directions.
+        sync_grid = self._grid.to(self.device) if self._grid is not None else None
+        assert sync_grid is not None
+        s_used = min(_NUM_SEG_PER_DIRECTION, vfeats.shape[1], afeats.shape[1])
```
SolitaryThinker: For clips shorter than the expected Synchformer window, `s_used` can be less than 14, but `compare_v_a()` still feeds that shorter token sequence into the vendored transformer. That transformer has a fixed positional embedding for the 14-segment shape and adds it without slicing/padding, so short-but-decodable clips can shape-mismatch here instead of returning a skipped result. Please guard for the required segment count before calling `compare_v_a()` or pad to the expected token length.

fastvideo/eval/metrics/audio/kl_divergence/metric.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class _patch_passt_stft:
+    """Re-enable PaSST's deprecated ``torch.stft(return_complex=False)``."""
+
+    def __init__(self) -> None:
+        self.old_stft = torch.stft
+
+    def __enter__(self) -> None:
+        torch.stft = partial(torch.stft, return_complex=False)  # type: ignore[assignment]
```
SolitaryThinker: This context manager mutates process-global `torch.stft` around each PaSST forward. Multi-GPU eval runs worker threads in the same process, so KL/FAD calls can overlap: one thread can restore `torch.stft` while another PaSST forward still expects the patched callable, or restore another thread's partial wrapper. Current `hear21passt` already passes `return_complex=False` upstream, so please remove this monkeypatch, or protect it with a shared lock if it is truly still required.

fastvideo/eval/metrics/audio/audiobox_aesthetics/metric.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def to(self, device):
+        super().to(device)
+        return self
+
+    def setup(self) -> None:
+        if self._predictor is not None:
+            return
+        from audiobox_aesthetics.infer import initialize_predictor
+        self._predictor = initialize_predictor()
```
SolitaryThinker: `initialize_predictor()` chooses its own device; upstream defaults to `torch.device("cuda")` when CUDA is available. In `num_gpus > 1`, each `EvalWorker` calls `m.to("cuda:i")` before setup, but this line ignores `self.device`, so all AudioBox predictors will load/run on default GPU 0. That removes the intended parallelism and can OOM GPU 0. Please construct or move the predictor onto the worker's assigned device.

SolitaryThinker (commented):
Review findings from the audio metric pass.

SolitaryThinker (approved):

PR #1355, opened 2026-05-15
[fix] Fix causal self-forcing attention settings

mignonjia:
## Purpose

Fix causal self-forcing KV-cache setup so `local_attn_size` and `sink_size` are read from the runtime transformer module.

Fixes #

## Changes

 - Align self forcing parameters with official Self Forcing repo. 
 - Make `local_attn_size` and `sink_size` configurable in yaml files during training.

## Test Plan

```bash
python -m py_compile fastvideo/pipelines/stages/causal_denoising.py
```

Also add wandb run logs

## Test Results

<details>

```
# passed, no output
```

</details>

 - Wan 2.1 T2V self forcing with `local_attn_size=-1` and `sink_size=0`: https://wandb.ai/hapo-exp/longlive/runs/cowj9mjc
 - Wan 2.1 T2V self forcing with `local_attn_size=12` and `sink_size=3`: https://wandb.ai/hapo-exp/longlive/runs/2p79j48k

## Checklist


fastvideo/models/dits/causal_wanvideo.py
```diff
@@ -120,9 +122,12 @@ def forward(self,
                 block_mask=block_mask
             )[:, :, :-padded_length].transpose(2, 1)
         else:
-            frame_seqlen = q.shape[1]
             current_end = current_start + roped_query.shape[1]
             sink_tokens = self.sink_size * frame_seqlen
+            max_attention_size = (
+                21 * frame_seqlen if self.local_attn_size == -1
```
alexzms: Note: with `local_attn_size == -1` this caps the KV window at `21 * frame_seqlen`, so a run with >21 latent frames silently truncates the oldest frames here (no error raised) — a quiet correctness bug. Suggest deriving `21` from config (`num_latent_t` / sliding window) or making it a named constant + a guard that errors when the real frame count exceeds it.

mignonjia: Previously we set this in class init: `self.max_attention_size = 32760 if local_attn_size == -1 else local_attn_size * 1560`
So this 21 is to match the previous hardcoded max length behavior. But I agree that this hardcode is incorrect.

mignonjia: Check my latest commit for this patch.

fastvideo/configs/models/base.py
```diff
@@ -56,13 +56,21 @@ def update_model_config(self, source_model_dict: dict[str, Any]) -> None:
...
         for key, value in source_model_dict.items():
             if key in valid_fields:
                 setattr(self, key, value)
+            elif key in valid_arch_fields:
+                setattr(arch_config, key, value)
             else:
                 logger.warning("%s does not contain field '%s'!", type(self).__name__, key)
                 raise AttributeError(f"Invalid field: {key}")
 
+        if hasattr(arch_config, "__post_init__"):
```
alexzms: why do we need this?

alexzms (approved):
LGTM

PR #1362, opened 2026-05-16
[perf] Quantize frames to uint8 on-device before the post-decode D->H copy

Mister-Raggs:
## Summary

`PostDecodeFrameProcessStage` spends most of its time on a deferred fp32 device→host transfer mischarged to it, plus a single-threaded per-frame CPU `*255`/cast loop.

Fix: cast the decoded video to `uint8` on the source device (typically CUDA) **before** the host copy. The transfer becomes 4× smaller (fp32 → uint8) and the elementwise work runs on the GPU instead of a Python per-frame CPU loop.

**Mechanism proven and independently cross-validated.** The per-path speedup (~−50–66%) is consistent; e2e impact scales with how much of total latency the post-decode path owns on a given model / resolution / GPU:

| Config | PostDecodeStage | End-to-end | Source |
|---|---|---|---|
| Cosmos 2.5, A100, 720×1280×29f, 8 steps | 5.991 s → 2.015 s (**−66%**) | 29.33 s → 24.75 s (**−15.6%**) | author |
| Wan2.1-T2V-1.3B, H100, 720×1280×77f, 30 steps | 0.580 s → 0.290 s (**−50%**) | 202.42 s → 198.92 s (**−1.7%**) | independent A/B by @rich7420 (Modal, avg of 2 runs) |

The post-decode path is ~25% of e2e on Cosmos/A100 but only ~0.3% on Wan/H100, so the same per-path halving lands very differently on e2e. Net-positive everywhere; magnitude is configuration-dependent. VAE and denoising stages are byte-untouched in both runs — the change is correctly scoped to the pixel path. Stacks with `enable_torch_compile` (independently ~−14%; different pipeline stages).

## Root cause

`samples` is a pinned-CPU buffer; `samples.copy_(output_batch.output)` is **non-blocking**, and `gen_time` is logged immediately after it. The full fp32 video D→H (scales with resolution × frames × batch and can be large) therefore completes lazily and the *first* op in the post-decode block blocks on it — so a large transfer is attributed to `PostDecodeFrameProcessStage`. On top of that, the old code ran `(x * 255).to(uint8)` in a Python loop over frames on the CPU tensor.

Evidence:
- nsys (whole process tree, `--trace-fork-before-exec`): fragmented CPU tail + transfer, GPU idle during the stage.
- Synthetic microbench of the exact block: pure CPU compute is only ~0.46 s (T=29) / ~1.8 s (T=125) — far below the ~6 s observed on Cosmos/A100, i.e. the stage isn't compute-bound.
- Real-conditions microbench (pinned dest, non-blocking copy, no pre-sync): reproduces the deferred-transfer tail; the on-device-uint8 path is 2.056 s → 0.034 s (T=29) and 2.304 s → 0.132 s (T=125).
- Real pipeline + independent cross-model A/B: table above.

## Correctness

Not bit-identical to the previous CPU cast: `float → uint8` can differ by ≤1 LSB between CPU and GPU on boundary pixels, and `clamp_(0, 255)` additionally fixes a latent overflow (VAE output slightly outside `[0, 1]` previously wrapped mod 256 in the unclamped cast). Gated by SSIM rather than exact equality:

**MS-SSIM (repo `compute_video_ssim_torchvision`, Cosmos 2.5, A100): 29/29 frames = 1.000000 exactly** (verified all frames read and compared, mean = min = max = 1.0).

## Skip the dead fp32 `samples` copy when `return_frames=False` (commit 2)

After commit 1, the post-decode frames path reads `output_batch.output` directly via the GPU `vid_u8` cast, so `samples` is consumed in exactly one place — the result dict's `samples` field, gated on `batch.return_frames` (`"samples": samples if batch.return_frames else None`). When the caller doesn't ask for `samples`, the pinned ~50 MB fp32 alloc and its D→H copy (and the latent fallthrough `.cpu()`) are dead weight; the typical generate flow (`save_video=True, return_frames=False`) hits this on every call.

Commit 2 extends `skip_pixel_prealloc` to include `not batch.return_frames` and short-circuits the copy on the same condition — removing the residual ~2 s the PR text originally called out as a follow-up. Behavior is unchanged for callers with `return_frames=True` (they still receive the fp32 `samples` tensor) or for the latent fallthrough when `return_frames=True`.

Further device-residency improvement — running `make_grid`/`permute` on-device with a single stacked transfer — is still a candidate follow-up, measured and SSIM-gated separately.

## Scope

Pixel-video path only. `output_type == "latent"` and audio-only paths are byte-unchanged. Callers passing `return_frames=True` still receive the fp32 `samples` tensor as before; commit 2 only skips work when nothing reads it.

## Test plan

- [x] nsys + microbench root-cause attribution
- [x] Real e2e on A100 (Cosmos 2.5 2B): PostDecode −66%, e2e −15.6%
- [x] Independent cross-model A/B (Wan2.1-T2V-1.3B, H100): PostDecode −50%, e2e −1.7%, VAE/denoising untouched
- [x] MS-SSIM gate (commit 1): 29/29 frames = 1.000000
- [x] SSIM re-gate after commit 2 (Wan2.1-T2V-1.3B, RTX A5000, 480×832×29f / 20 steps, FlashAttention-2, `return_frames=False`): BASE (copy runs) vs CAND (copy skipped) = **MS-SSIM mean/min/max = 1.000000 across all frames** — bit-identical, confirming the skip is output-neutral
- [x] CI (buildkite + fastcheck passing)
- [ ] Maintainer review


rich7420:
@Mister-Raggs Thanks for the patch!

A/B results — Wan2.1-T2V-1.3B, H100×1, 720×1280×77f, 30 steps
Same-container A/B on Modal (avg of 2 runs each):
| Stage | Baseline | PR | Δ |
|---|---|---|---|
| PostDecodeFrameProcessStage | 0.580 s | 0.290 s | **−50%** |
| VideoSave | 0.820 s | 0.701 s | −14.5% |
| Decoding (VAE) | 7.325 s | 7.307 s | −0.2% |
| Denoising | 192.701 s | 189.658 s | −1.6% |
| **End-to-end** | **202.42 s** | **198.92 s** | **−1.73%** |

Per-run PostDecode very stable: baseline 0.568 / 0.592 s, treatment 0.296 / 0.285 s.
The mechanism is real and confirmed. PostDecode halved exactly as described — the on-GPU uint8 cast + smaller D→H transfer is working as intended, and VAE/Denoising are untouched.
On Wan T2V, PostDecode was already only ~0.6 s (vs ~6 s on Cosmos 2.5 / A100 where the fix was measured). The savings scale with how much time was being lost in that path — on Wan T2V it was never the bottleneck, so e2e impact lands at −1.7% instead of −15%.
Still a clean net-positive fix. The impact just varies a lot by model/resolution/hardware depending on how large the post-decode buffer is and whether that D→H was already the hotspot.

rich7420 (approved):

SolitaryThinker:
Reviewed for merge. The mechanism is sound and independently reproduced, and the clamp
is a genuine bug fix (out-of-range VAE output used to wrap mod 256). CI on the head is
green. Three notes from a deep pass, all handled and pushed directly onto this branch via
maintainer edit (a merge of current `main` + three additive commits, mirrored as `pr1362-fixes`
instead, `maintainer_can_modify` is on):

1. **Base-drift regression (fixed in `15a164a05`):** `main` picked up #1595's refiner
   size reporting (`_resolve_output_size(samples, ...)`) after your last push. With
   commit 3 leaving `samples` empty when `return_frames=False`, `GenerationResult.size`
   silently fell back to the *requested* geometry in exactly the save flow this PR
   optimizes. Fixed by reading the geometry from `output_batch.output` (shape-only, no
   copy), with a red/green regression test.
2. **Comment accuracy (fixed in `f56f56704`):** `samples.copy_(output)` never passed
   `non_blocking=True` (`git log -S` confirms), so the in-code "deferred non-blocking
   transfer" rationale couldn't be right as written — the win comes from the 4×-smaller
   single uint8 copy plus moving the elementwise math off the single-threaded CPU loop,
   which your A/Bs demonstrate directly. Also dropped a reintroduced "~50 MB" constant
   and scoped the "typical flow" claim to the CLI (the `SamplingParam` API default is
   `return_frames=True`).
3. **Tests (`aa95a4c18`):** CPU regression tests pinning (a) byte-exact equivalence of
   the new path vs the legacy loop for in-range fp32 (batch>1 grid layout, odd frame
   count, uint8 HWC contract — the test also passes against the pre-PR loop, so it pins
   both sides) and (b) the clamp behavior (fails on the old loop).

On @gemini-code-assist's suggestion to move `make_grid` on-device: agree with deferring
— the transfer is already a single stacked copy after this PR, the remaining CPU work is
uint8 memcpy-grade, and that exact loop is also being rewritten by #1660/#1687, so it's
better as a separately measured follow-up. Heads-up @Ishxn20: after this lands, the
`video_generator.py` hunk in #1660/#1687 is superseded and should drop out on rebase.

Approving once `full-suite` passes on the merged state.

SolitaryThinker (approved):

PR #1357, opened 2026-05-16
[misc] Rename MatrixGame to MatrixGame2

H1yori233:
## Purpose

Rename all `MatrixGame` occurrences to `MatrixGame2` to prepare for adding `MatrixGame3` later.

Also bumps HF repo names: 
- `Foundation` → `Base`
- `Base` → `Base-Distilled`
- `GTA` → `GTA-Distilled`
- `TempleRun` → `TempleRun-Distilled`.

Legacy class names, HF paths, and CLI flags are all kept as backward-compat aliases.

## Test

- preset suite: 58/58 ✅
- real inference via Matrix-Game-2.0-Base-Diffusers (old `_class_name=MatrixGameCausalDMDPipeline` → new class) ✅
- worldmodel YAML `_target_` / shell-script torchrun paths all still resolve ✅

## Checklist

- [X] I ran `pre-commit run --all-files` and fixed all issues
- [X] I added or updated tests for my changes
- [X] I updated documentation if needed
- [X] I considered GPU memory impact of my changes

**For model/pipeline changes, also check:**
- [X] I verified SSIM regression tests pass
- [X] I updated the support matrix if adding a new model

PR #1364, opened 2026-05-17
[bugfix] registry: infer workload_type from model registration instead of defaulting to T2V

alexzms:
## Problem

`bar-chart-ssim-tests` (full-suite) fails deterministically on `test_matrixgame_similarity.py::Matrix-Game-2.0-Diffusers-Base`:

```
ValueError: Pipeline 'MatrixGame2I2VPipeline' is not supported for
pipeline type 'basic' and workload type 't2v'
```

This blocks `full-suite-passed` on any PR that runs it (e.g. #1242).

## Root cause

`get_model_info()` in `fastvideo/registry.py` hard-defaults `workload_type` to `WorkloadType.T2V` when the caller doesn't pass one (`build_pipeline` forwards `fastvideo_args.workload_type`, which is `None` for the Matrix-Game-2.0 SSIM run). Matrix-Game-2.0 is registered I2V-only (`register_configs(..., workload_types=(WorkloadType.I2V,))`), so resolving its pipeline under the assumed `t2v` workload fails.

## Fix

When `workload_type is None`, derive it from the model's registered `ConfigInfo.workload_types` rather than assuming T2V:

- model registers exactly one workload → use it (Matrix-Game-2.0 → `I2V`)
- multiple / empty registration → keep the existing `T2V` fallback

Behavior is unchanged for every T2V model and for multi/empty registrations; only single-non-T2V models (the broken case) change. `config_info` is resolved once, slightly earlier — it was already required later in the same function.

## Test evidence

- `pre-commit run --files fastvideo/registry.py`: all hooks pass (yapf/ruff/codespell/mypy).
- Functional check (exact-path registry match, no GPU/network):
  - `FastVideo/Matrix-Game-2.0-Base-Diffusers` → registered `(I2V,)` → inferred `I2V` ✅ (previously `T2V` → the bug)
  - `Wan-AI/Wan2.1-T2V-1.3B-Diffusers` → registered `(T2V,)` → inferred `T2V` ✅ (unchanged, no regression)

## Context

Pre-existing breakage unrelated to #1242; this unblocks the `full-suite-passed` gate there (minimal LoRA training support).

### DIFF
--- a/fastvideo/registry.py
+++ b/fastvideo/registry.py
@@ -781,8 +781,16 @@ def get_model_info(
     elif isinstance(pipeline_type, str):
         pipeline_type = PipelineType.from_string(pipeline_type)
 
+    config_info = _get_config_info(model_path, raise_on_missing=True)
+    assert config_info is not None, "config_info must be resolved"
+
     if workload_type is None:
-        workload_type = WorkloadType.T2V
+        # Derive the workload from the model's registration instead of
+        # assuming T2V. A model registered for a single workload (e.g.
+        # Matrix-Game-2.0, which is I2V-only) otherwise fails pipeline
+        # resolution under the blanket T2V default.
+        registered = config_info.workload_types
+        workload_type = (registered[0] if len(registered) == 1 else WorkloadType.T2V)
 
     if os.path.exists(model_path):
         config = verify_model_config_and_directory(model_path)
@@ -800,9 +808,6 @@ def get_model_info(
 
     pipeline_registry = get_pipeline_registry(pipeline_type)
     pipeline_cls = pipeline_registry.resolve_pipeline_cls(pipeline_name, pipeline_type, workload_type)
-
-    config_info = _get_config_info(model_path, raise_on_missing=True)
-    assert config_info is not None, "config_info must be resolved"
 
     sampling_param_cls = config_info.sampling_param_cls or SamplingParam
 


SolitaryThinker:
**Updated review — the headline problem is now solved on `main` by a different mechanism; this PR's motivation is mostly obsolete.**

## Context update since the original review

Origin/main now ships a new `fastvideo/pipelines/basic/matrixgame2/` package (introduced by #1357 or a follow-up) that directly registers the missing pipeline:

```
origin/main:fastvideo/pipelines/basic/matrixgame2/__init__.py
  __all__ = ["MatrixGame2I2VPipeline", "MatrixGame2CausalDMDPipeline"]
origin/main:fastvideo/pipelines/basic/matrixgame2/matrixgame2_i2v_pipeline.py
  class MatrixGame2I2VPipeline(LoRAPipeline, ComposedPipelineBase): ...
  class MatrixGamePipeline(MatrixGame2I2VPipeline): ...
  EntryClass = [MatrixGame2I2VPipeline, MatrixGamePipeline]
```

This means: `PipelineType.BASIC` resolution of `_class_name="MatrixGame2I2VPipeline"` from `FastVideo/Matrix-Game-2.0-Base-Diffusers` model_index.json now succeeds via the `EntryClass` list, with NO workload_type inference required. The CI failure this PR was trying to fix should disappear once any failing branch is rebased onto current `main`.

I verified this on `origin/main` directly — the failure surface described in the PR description no longer exists on top-of-tree.

## What that means for this PR

The diff in this PR (workload_type-None → infer-from-`ConfigInfo.workload_types`) is independent of the missing-pipeline fix and the original review's S1 finding is now moot — there is no longer a `MatrixGame2I2VPipeline`-not-found error to fix.

The inference logic itself is still mostly sound (see the original S3-1 in the prior review), but its **motivating use case is gone**. Three options:

1. **Close as obsoleted.** The matrixgame2 path is now correct on `main` via the new package; no inference change needed.
2. **Rebase and re-pitch** as a generic cleanup ("infer workload_type from single-registration models") with a new motivating example that ISN'T MatrixGame2. Document which models actually benefit at HEAD, and add a unit test that exercises `get_model_info()` end-to-end (not just the tuple inference). The original S2 finding about test coverage still applies.
3. **Tighten the change** to only opt-in models that explicitly want this behavior (e.g. a registry flag), avoiding a silent global semantics shift on the `workload_type=None` path for any future single-workload model.

My recommendation is option 1 (close) unless you have a concrete second motivating case beyond MatrixGame2 — option 2 risks shipping a behavioral change in the dispatch path without a current concrete need.

## Original review for full context

The detailed walkthrough of the workload-inference logic, the S1 mechanism-vs-claim gap, the cross-check vs #1355, and the test-coverage gap are in the earlier review thread. The S2 ('add an end-to-end `get_model_info()` regression test') and S3 ('multi-workload fallback policy') findings remain valid for option 2 if you pick that path.

— Reviewer notes generated by SolitaryThinker's reviewer agent.

PR #1365, opened 2026-05-18
[perf] Mark LayerwiseOffloadHook entry points torch.compiler.disable (remove per-layer graph break)

Mister-Raggs:
## Summary

`dit_layerwise_offload=True` is the **default**. `LayerwiseOffloadState`'s methods are already `@torch.compiler.disable`, but the `LayerwiseOffloadHook` entry points (`pre_forward`/`post_forward`) were **not**. With `enable_torch_compile`, dynamo traces *into* the hook, hits the disabled callee, and is forced to bail mid-trace — an **implicit graph break at the call site, once per transformer layer every step**. This fragments the per-layer compiled region and blocks CUDA-graph capture (cudagraphs cannot span a graph break).

This marks the two hook entry points `@torch.compiler.disable` so the hook is a clean opaque eager boundary instead of an implicit break. The hook is host-side memory orchestration (CUDA streams, pinned H2D/D2H, param swap) — never compilable by design — so this changes *how dynamo treats the boundary*, not the computation. It completes the decorator pattern already applied to the `LayerwiseOffloadState` methods in the same file.

**This is a correctness/hygiene change, not a speedup PR.** Honest framing below.

## Evidence (Wan2.1-T2V-1.3B, A100-80GB, 480×832×81f, 50 steps)

Enumerated graph breaks via `TORCH_LOGS=graph_breaks` (the DiT compiles in the MultiprocExecutor worker, so `torch._dynamo.explain()` in the caller can't see it; the env-var route does).

| | Before | After |
|---|---|---|
| Total graph breaks | 25 | 13 |
| `layerwise_offload.py:102/104/108` (per-layer) | **present, 5 instances** | **gone** |
| `enable_torch_compile` e2e (default inductor) | 259.7s → 198.1s (**−23.7%**, independent of this change) | unchanged |
| This change's *isolated* e2e effect (compile-on pre vs post) | — | **within noise** (gen 197.2s → 196.3s; denoise 3.80 → 3.78 s/it; the e2e gap is PostDecode jitter, an unrelated stage) |
| MS-SSIM, compiled pre-fix vs post-fix | — | **1.000000 across 81/81 frames** (repo `compute_video_ssim_torchvision`; all frames verified read+compared) |

So: the per-layer break is **eliminated** and the compiled output is **bit-identical**, but on Wan-1.3B/A100 there is **no measurable standalone speedup** — the fusion gain alone is in the noise. The value is (a) it's a real missed-decorator bug in the **default-on, shared** offload path (every model run with `enable_torch_compile` hits it), and (b) it is **one of two breaks** that must be removed before `mode=reduce-overhead` / CUDA graphs can work.

## Characterized follow-up (not in this PR)

With this fix applied, `torch_compile_kwargs={"mode":"reduce-overhead"}` gets *further* — cudagraph capture now starts (it couldn't before) — but still hard-fails:

```
RuntimeError: accessing tensor output of CUDAGraphs that has been overwritten
  ... torch_dynamo_resume_in_forward_at_363  (fastvideo/models/dits/wanvideo.py)
```

The **remaining** `wanvideo.py:363` break — the FlashAttention-2 dispatch (`self.attn1`), a custom CUDA op dynamo can't trace — splits each block into two cudagraph segments and the cross-segment static-buffer handoff violates cudagraph-trees assumptions. Two scoped follow-ups to unlock CUDA graphs (each its own measured PR):

1. Eliminate the `wanvideo.py:363` attention-call break (make the flash-attn dispatch dynamo-traceable via `torch.library`/`allow_in_graph`).
2. Add `torch.compiler.cudagraph_mark_step_begin()` once per denoise step in the pipeline loop (the documented mitigation for this exact error).

This PR is the prerequisite for both; landing it standalone keeps the change focused and SSIM-gated.

## Scope

One file, two decorators. No behavior change to the offload mechanism. Pixel/latent/audio paths unaffected.

## Test plan

- [x] Graph-break enumeration: per-layer offload breaks eliminated (25→13; 102/104/108 gone)
- [x] MS-SSIM compiled pre-fix vs post-fix = 1.000000, all 81/81 frames verified
- [x] `enable_torch_compile` end-to-end still works (−23.7% from compile itself, unchanged by this)
- [x] Cudagraph blocker characterized (remaining attention break documented as follow-up)
- [ ] CI


rich7420 (approved):
thanks for the patch!

SolitaryThinker (approved):
thanks

PR #1373, opened 2026-05-19
[perf]: register FA2/FA3 default flash_attn_func as a torch.library custom op

Mister-Raggs:
## What

Wraps the FA2/FA3 default `flash_attn_func` in a
`fastvideo::_flash_attn_default_forward` `torch.library.custom_op`
(+ `register_fake`), mirroring the existing FP4 `flash_attn_cute`
custom-op template. The external flash-attn call becomes an
opaque-but-traceable graph node instead of forcing a Dynamo graph
break at the call site, bringing the FA2/FA3 path to parity with the
already-registered FP4 path.

## What this does NOT claim (scope, deliberately narrow)

- **Output-neutral.** SSIM mean/min/max = `1.000000` over all frames,
  under both `torch.compile` and eager, hardware-validated on **FA2**
  (H100, Wan2.1-T2V-1.3B).
- **FA3 path:** same wrapper / same code path; FA3's `flash_attn_func`
  has the same single-tensor return (Dao-AILab unified the API). Not
  hardware-validated here — the FA3 from-source build OOM'd on this
  torch 2.10 / cu128 box. Covered by code symmetry and a maintainer's
  separate in-flight FA3 run.
- **Does NOT remove the `wanvideo.py:363` graph break.** Validated on
  H100: break inventory unchanged (128 → 128). That break is dominated
  by the independent `@torch.compiler.disable` on
  `DistributedAttention.forward`, which breaks at the same call site
  and masks the flash-attn opacity addressed here. **No break-count or
  performance claim is made.**

## Why land it independently

Parity with the FP4 path, and a prerequisite/enabler for downstream
`torch.compile` work (the FA2/FA3 call must be a traceable node before
any region containing it can be fused / captured). Aligned with
maintainer feedback that this is a reasonable minimum to land on its
own.

## Test evidence

- SSIM 1.000000 (mean/min/max, all frames), `torch.compile` + eager,
  FA2, H100 / Wan2.1-T2V-1.3B.
- Graph-break inventory diff BASE vs this change: identical
  (128 total, `wanvideo.py:363` unchanged) — documents the no-op on
  break count, consistent with the cause analysis above.

Env: H100 80GB, torch 2.10.0+cu128, flash_attn 2.8.3.

### DIFF
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -83,8 +83,7 @@ if fa_version in ("2", "3"):
         # dynamo graph break on the training path — i.e. pre-PR behavior, no
         # regression. Full autograd parity for the custom op (mirroring the FP4
         # cute template) is a tracked follow-up.
-        if torch.is_grad_enabled() and (q.requires_grad or k.requires_grad
-                                        or v.requires_grad):
+        if torch.is_grad_enabled() and (q.requires_grad or k.requires_grad or v.requires_grad):
             return _fa_default(q, k, v, softmax_scale=softmax_scale, causal=causal)
         return torch.ops.fastvideo._flash_attn_default_forward(q, k, v, softmax_scale, causal)
 elif fa_version == "4":
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -73,6 +73,19 @@ if fa_version in ("2", "3"):
         return q.new_empty(q.shape[0], q.shape[1], q.shape[2], v.shape[-1])
 
     def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
+        # Autograd carve-out. The custom op above registers a forward + fake
+        # kernel but NO backward (register_autograd), so it is opaque to
+        # autograd. Inference runs under no_grad / inference_mode and routes
+        # through the traceable custom op — that is the torch.compile win, and
+        # the only path this PR claims. Training backprops through attention,
+        # so route grad-enabled calls to the original FA2/FA3 `flash_attn_func`
+        # (itself an autograd.Function, so backward is correct) at the cost of a
+        # dynamo graph break on the training path — i.e. pre-PR behavior, no
+        # regression. Full autograd parity for the custom op (mirroring the FP4
+        # cute template) is a tracked follow-up.
+        if torch.is_grad_enabled() and (q.requires_grad or k.requires_grad
+                                        or v.requires_grad):
+            return _fa_default(q, k, v, softmax_scale=softmax_scale, causal=causal)
         return torch.ops.fastvideo._flash_attn_default_forward(q, k, v, softmax_scale, causal)
 elif fa_version == "4":
     # FA4 path: `flash_attn_func` is already a torch.library custom op
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -22,6 +22,55 @@ except ImportError:
         flash_attn_func = flash_attn_2_func
         fa_version = "2"
 
+
+# torch.compile traceability: the FA4/cute path (fa_version=="4") is
+# already a registered torch.library custom op, so dynamo treats it as a
+# graph node. The external FA2/FA3 `flash_attn_func` is NOT — dynamo
+# breaks the graph at the call site (observed: wanvideo.py self-attn,
+# once per layer every step), which fragments the compiled region and
+# blocks CUDA-graph capture. Wrap the FA2/FA3 default call in a custom
+# op (mirrors the FP4 `flash_attn_cute` template) so it becomes an
+# opaque-but-traceable node. The kernel still runs eager inside the op
+# (correct — flash-attn must run eager); only dynamo's treatment of the
+# boundary changes, so numerics are unchanged (SSIM-gate to confirm).
+if fa_version in ("2", "3"):
+    _fa_default = flash_attn_func
+
+    @torch.library.custom_op(
+        "fastvideo::_flash_attn_default_forward",
+        mutates_args=(),
+        device_types="cuda",
+    )
+    def _flash_attn_default_forward(
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        softmax_scale: float | None,
+        causal: bool,
+    ) -> torch.Tensor:
+        return _fa_default(q, k, v, softmax_scale=softmax_scale, causal=causal)
+
+    @torch.library.register_fake("fastvideo::_flash_attn_default_forward")
+    def _flash_attn_default_forward_fake(
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        softmax_scale: float | None,
+        causal: bool,
+    ) -> torch.Tensor:
+        del softmax_scale, causal
+        # FA2/FA3 default path: [batch, seqlen_q, nheads, head_dim_v],
+        # same dtype/device as q (head dim taken from v).
+        return q.new_empty(q.shape[0], q.shape[1], q.shape[2], v.shape[-1])
+
+    def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
+        return torch.ops.fastvideo._flash_attn_default_forward(
+            q, k, v, softmax_scale, causal)
+else:
+    # fa_version == "4": flash_attn_func is already a custom op.
+    def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
+        return flash_attn_func(q, k, v, softmax_scale=softmax_scale, causal=causal)
+
 from fastvideo.attention.backends.abstract import (
     AttentionBackend,
     AttentionImpl,
@@ -228,7 +277,10 @@ class FlashAttentionImpl(AttentionImpl):
             output = self._forward_nvfp4(query, key, value)
 
         else:
-            output = flash_attn_func(
+            # Route through the compilable wrapper so dynamo sees a
+            # registered op (no graph break) for FA2/FA3; identical
+            # kernel + numerics, op runs eager internally.
+            output = flash_attn_func_compilable(
                 query,  # type: ignore[no-untyped-call]
                 key,
                 value,
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -22,7 +22,6 @@ except ImportError:
         flash_attn_func = flash_attn_2_func
         fa_version = "2"
 
-
 # torch.compile traceability: the FA4/cute path (fa_version=="4") is
 # already a registered torch.library custom op, so dynamo treats it as a
 # graph node. The external FA2/FA3 `flash_attn_func` is NOT — dynamo
@@ -36,6 +35,16 @@ except ImportError:
 if fa_version in ("2", "3"):
     _fa_default = flash_attn_func
 
+    # Scope: this op covers exactly the q/k/v + softmax_scale + causal
+    # call shape used by FlashAttentionImpl.forward's default branch
+    # (see `flash_attn_func_compilable(...)` call site below). The
+    # masked/no-pad and varlen / cross-attn paths use different
+    # entry points (`flash_attn_no_pad`, `flash_attn_varlen_*`) which
+    # are intentionally out of scope for this PR — wrapping them is a
+    # natural follow-up. The wrapper's signature is the contract: any
+    # extra kwarg (dropout_p, window_size, alibi_slopes, deterministic,
+    # return_attn_probs, ...) raises TypeError at the call site, so
+    # silent loss of kwargs is not a failure mode.
     @torch.library.custom_op(
         "fastvideo::_flash_attn_default_forward",
         mutates_args=(),
@@ -64,12 +73,19 @@ if fa_version in ("2", "3"):
         return q.new_empty(q.shape[0], q.shape[1], q.shape[2], v.shape[-1])
 
     def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
-        return torch.ops.fastvideo._flash_attn_default_forward(
-            q, k, v, softmax_scale, causal)
-else:
-    # fa_version == "4": flash_attn_func is already a custom op.
+        return torch.ops.fastvideo._flash_attn_default_forward(q, k, v, softmax_scale, causal)
+elif fa_version == "4":
+    # FA4 path: `flash_attn_func` is already a torch.library custom op
+    # (registered in `fastvideo.attention.utils.flash_attn_cute`), so a
+    # passthrough is enough — no extra registration needed.
     def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
         return flash_attn_func(q, k, v, softmax_scale=softmax_scale, causal=causal)
+else:
+    # Defensive: the probe above only ever sets fa_version to "2", "3",
+    # or "4"; an unexpected value means an import/probe regression and
+    # we want a loud error at import, not a silent NameError later.
+    raise RuntimeError(f"Unsupported FlashAttention version: {fa_version!r} — expected "
+                       f"'2', '3', or '4' from the import probe above.")
 
 from fastvideo.attention.backends.abstract import (
     AttentionBackend,
--- a/fastvideo/tests/attention/test_flash_attn_default_custom_op.py
+++ b/fastvideo/tests/attention/test_flash_attn_default_custom_op.py
@@ -1,0 +1,121 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression guard for the FA2/FA3 default-path custom op (PR #1373).
+
+`flash_attn_func_compilable` wraps the default FA2/FA3 `flash_attn_func` in a
+`torch.library.custom_op` so dynamo sees a traceable node (no graph break)
+during inference. That custom op registers a forward + fake kernel but *no*
+backward, so it is opaque to autograd. To keep training working, the
+dispatcher routes grad-enabled calls to the original `flash_attn_func` (an
+autograd.Function) and only sends grad-free (inference) calls through the
+custom op.
+
+These tests pin both halves of that contract:
+  - inference (no grad): output goes through the custom op and matches the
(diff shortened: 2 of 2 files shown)


SolitaryThinker (approved):

SolitaryThinker:
**Hi @Mister-Raggs** — follow-up from one of @SolitaryThinker's AI reviewer agents (Gob) on the failing CI for this PR. I run these to help triage so @SolitaryThinker can prioritize; ping @SolitaryThinker if anything below doesn't match what you know.

## TL;DR

CI build [#3219](https://buildkite.com/fastvideo/ci/builds/3219) is failing on **3 training jobs** (`test-tube-training-tests`, `test-tube-lora-training-tests`, `test-tube-training-tests-vsa`) while inference jobs (`bar-chart-ssim-tests`, `test-tube-performance-tests`) and the microscope/fastcheck suites all **pass**. Root cause looks high-confidence: the new `fastvideo::_flash_attn_default_forward` custom op registers a forward (`custom_op`) and a fake kernel (`register_fake`), but **does not register an autograd implementation**. Inference is unaffected; training breaks because the wrapper has no backward.

## Evidence

### Diff inventory at HEAD `4e99077942a4`

`fastvideo/attention/backends/flash_attn.py` registers exactly two pieces for the FA2/FA3 path:

- `@torch.library.custom_op("fastvideo::_flash_attn_default_forward", ...)` — `flash_attn.py:48`
- `@torch.library.register_fake("fastvideo::_flash_attn_default_forward")` — `flash_attn.py:62`

There is no `setup_context` function and no `torch.library.register_autograd(...)` call for `fastvideo::_flash_attn_default_forward` anywhere in the file.

### The FP4 template the PR claims to mirror has all 4 pieces

`fastvideo/attention/utils/flash_attn_cute.py` (unchanged in this PR, the explicit template per the PR body's "mirrors the existing FP4 `flash_attn_cute` custom-op template" claim):

- `@torch.library.custom_op("fastvideo::_flash_attn_cute_forward", ...)` — `flash_attn_cute.py:16`
- `@torch.library.register_fake("fastvideo::_flash_attn_cute_forward")` — `flash_attn_cute.py:44`
- `def _flash_attn_cute_setup_context(ctx, inputs, output)` — `flash_attn_cute.py:60`
- `torch.library.register_autograd("fastvideo::_flash_attn_cute_forward", _flash_attn_cute_backward, setup_context=_flash_attn_cute_setup_context)` — `flash_attn_cute.py:93`

The varlen variant in the same file (`fastvideo::_flash_attn_cute_varlen_forward` at lines 100, 136, 158, 208) follows the identical 4-piece pattern. The FA2/FA3 wrapper in this PR is the only one missing pieces 3 and 4.

### Failure pattern matches autograd-missing precisely

| Job | Calls backward? | Result |
|---|---|---|
| `bar-chart-ssim-tests` | no (inference + SSIM compare) | **PASS** ✓ |
| `test-tube-performance-tests` | no (inference timing) | **PASS** ✓ |
| `microscope-transformer-tests` / `microscope-unit-tests` | mostly forward | **PASS** ✓ |
| `test-tube-training-tests` | yes | **FAIL** ✗ |
| `test-tube-lora-training-tests` | yes | **FAIL** ✗ |
| `test-tube-training-tests-vsa` | yes (VSA training) | **FAIL** ✗ |

This is the canonical no-backward-on-custom-op pattern. The expected PyTorch error at the failure site is approximately:

```
RuntimeError: trying to backward through the graph a second time, OR
RuntimeError: function fastvideo::_flash_attn_default_forward does not have
              a registered autograd kernel
```

(I couldn't pull the exact assertion text — Buildkite's raw_log endpoint requires `BUILDKITE_API_TOKEN` which isn't on @SolitaryThinker's automation token. If the actual assertion differs from the above two shapes, please let me know — but the structural mechanism is high confidence either way.)

### Why this is new (the wrapper changes behavior, the old call site didn't)

The pre-PR call site at `flash_attn.py:277` called `flash_attn_func(query, key, value, softmax_scale=..., causal=...)` directly. That callable is the `flash_attn` package's pure-Python `flash_attn_func`, which internally invokes an `autograd.Function` (`FlashAttnFunc`). Dynamo took a graph break at the call site but the autograd path worked because the underlying wrapper IS an autograd.Function.

The PR's new wrapper at `flash_attn.py:299` calls `flash_attn_func_compilable(...)`, which routes through `torch.ops.fastvideo._flash_attn_default_forward(...)`. That registered op has no autograd kernel, so the dispatcher does NOT route to FA's internal `FlashAttnFunc.backward` — gradients never flow through the wrapper. At training time, `loss.backward()` fails (or produces zero/None grads, depending on the call shape).

The PR's own validation was inference-only (SSIM=1.0 on Wan2.1-T2V-1.3B on H100). The earlier review (https://github.com/hao-ai-lab/FastVideo/pull/1373#issuecomment-XXX) was also inference-scoped; the training path was never exercised. That's the gap.

## Suggested fix

**Mirror the FP4 template's 4-piece pattern.** Add `_setup_context` + `register_autograd` for `fastvideo::_flash_attn_default_forward`. The backward kernel is straightforward — FA2/FA3 expose a public `_flash_attn_backward` (FA2: `flash_attn._flash_attn_backward`; FA3: `flash_attn_interface._flash_attn_backward`). Sketch:

```python
if fa_version in ("2", "3"):
    if fa_version == "2":
        from flash_attn import _flash_attn_backward as _fa_default_backward
    else:  # "3"
        from flash_attn_interface import _flash_attn_backward as _fa_default_backward

    @torch.library.custom_op(
        "fastvideo::_flash_attn_default_forward",
        mutates_args=(),
        device_types="cuda",
    )
    def _flash_attn_default_forward(q, k, v, softmax_scale, causal) -> torch.Tensor:
        # FA2/FA3 forward returns (out, softmax_lse) when return_attn_probs=False
        # via the underlying autograd Function; for compatibility with the
        # existing register_fake (single-Tensor return), we just take the output.
        # The backward needs softmax_lse + RNG state from the forward pass.
        out, softmax_lse, S_dmask, rng_state = _flash_attn_forward(
            q, k, v, dropout_p=0.0, softmax_scale=softmax_scale,
            causal=causal, return_softmax=False,
        )
        return out  # autograd path will need softmax_lse + rng_state via ctx

    @torch.library.register_fake("fastvideo::_flash_attn_default_forward")
    def _flash_attn_default_forward_fake(q, k, v, softmax_scale, causal):
        del softmax_scale, causal
        return q.new_empty(q.shape[0], q.shape[1], q.shape[2], v.shape[-1])

    def _flash_attn_default_setup_context(ctx, inputs, output):
        q, k, v, softmax_scale, causal = inputs
        # NOTE: setup_context can NOT recompute softmax_lse — it must be saved
        # from the forward. Easiest path: change the custom_op's forward to
        # return (out, softmax_lse) as a tuple, mirror register_fake to a
        # 2-tuple, and ctx.save_for_backward(softmax_lse, q, k, v) here.
        ...

    def _flash_attn_default_backward(ctx, grad_out):
        q, k, v, softmax_lse = ctx.saved_tensors
        dq = torch.empty_like(q)
        dk = torch.empty_like(k)
        dv = torch.empty_like(v)
        _fa_default_backward(
            grad_out, q, k, v, ctx.output, softmax_lse,
            dq, dk, dv,
            dropout_p=0.0,
            softmax_scale=ctx.softmax_scale,
            causal=ctx.causal,
            ...,
        )
        return dq, dk, dv, None, None  # None for softmax_scale + causal

    torch.library.register_autograd(
        "fastvideo::_flash_attn_default_forward",
        _flash_attn_default_backward,
        setup_context=_flash_attn_default_setup_context,
    )
```

The shape contract change required: the forward needs to either save `softmax_lse` via `ctx.save_for_backward` (FA backward can't recompute it from `q/k/v/out` alone — it's a per-head log-sum-exp from the forward attention math), OR the custom_op forward returns `(out, softmax_lse)` as a 2-tuple, and the `register_fake` shape becomes `tuple[Tensor, Tensor]` to match. The 2-tuple shape is exactly what the FP4 template does at `flash_attn_cute.py:_flash_attn_cute_forward` (lines 21-37 return `(out, lse)`) so that's the cleanest mirror.

If you'd rather defer training-path support and keep this PR purely inference-only:

**Alternative (smaller fix, scope-reducing)**: bypass the custom_op when autograd is enabled.

```python
def flash_attn_func_compilable(q, k, v, softmax_scale=None, causal=False):
    if torch.is_grad_enabled() and (q.requires_grad or k.requires_grad or v.requires_grad):
        # Training path: call the original FA wrapper directly (it's an
        # autograd.Function so backward works); pay the graph break.
        return _fa_default(q, k, v, softmax_scale=softmax_scale, causal=causal)
    return torch.ops.fastvideo._flash_attn_default_forward(q, k, v, softmax_scale, causal)
```

This restores training behavior to pre-PR (graph break + working autograd) and keeps the compile-traceability win for inference. It loses the "parity with FP4 path" property the PR body argues for, but it's a 5-line fix and unblocks the 3 failing training jobs.

## What I'd recommend

Pick the principled fix (full `register_autograd`) if the goal is "FA2/FA3 path reaches parity with FP4 for both inference and training" — that matches the PR's stated framing of "parity with the FP4 path". The 2-tuple shape change is a bit more code but mirrors the FP4 template exactly.

Pick the conditional bypass if the goal is "ship inference-only torch.compile traceability now, defer training-path autograd to a follow-up PR". The PR body explicitly carved out "**Output-neutral**" and "**No break-count or performance claim is made**" but did NOT carve out "training is supported", so this would be a legitimate scope adjustment if surfaced in the PR body.

Either way, **add an `opcheck` smoke test** before merging. `torch.library.opcheck(torch.ops.fastvideo._flash_attn_default_forward, args, kwargs)` exercises forward + fake + autograd in one call and would have caught this at unit-test time. (The FP4 path almost certainly has analogous coverage — worth checking and mirroring.)

## CI re-run after fix

Once the fix lands, the 3 failing jobs should flip green without touching `bar-chart-ssim-tests` (it was already green and shouldn't regress). The pre-existing inference SSIM 1.000000 result you reported on Wan2.1-T2V-1.3B/H100 stays valid because the inference path is unchanged.

---

*This triage is from @SolitaryThinker's agent **Gob** (an AI reviewer). Ping @SolitaryThinker if any of the above is off (especially the FA-backward signature — I had to read FA's API from memory since I couldn't pull the actual assertion text from Buildkite without an API token). Earlier inference-only review on this PR remains valid for the inference path; the training-path gap is the new finding.*

SolitaryThinker (changes requested):
please address above comments and fix CI

Mister-Raggs:
Thanks  @SolitaryThinker, confirmed the training break and it's exactly the missing-backward you flagged. The custom op has a forward + fake kernel but no register_autograd, and the default branch was routing every FA2/FA3 call through it, so anything that backprops (the three training jobs) fell over. Inference never hits backward, which is why SSIM/perf stayed green.

For now I've kept this PR scoped to what it set out to do - make inference traceable under torch.compile - and added an autograd carve-out so training goes back to its pre-PR path: grad-enabled calls fall through to the original flash_attn_func (it's an autograd.Function, so backward is correct), at the cost of the graph break on the training path. Inference keeps the custom op and stays traceable. No numerics change either way. Also added an opcheck test plus inference-parity and a training-backward test so this can't regress silently again.

The fuller fix, giving the custom op its own backward so training is also traceable, mirroring the FP4 cute template's 4-piece pattern, I'd rather do as a follow-up. FA2 and FA3 have different backward signatures and it needs real training-numerics validation, which is more than this PR was scoped for. I'll open a tracked issue for it.

One small correction on scope: the masked/varlen path isn't wrapped in this PR (those still call the original funcs) - that wrapping is on a separate branch, so this PR's only training-path change is the default branch.

SolitaryThinker:
**Hi @Mister-Raggs** — this is a differential re-review from one of @SolitaryThinker's AI reviewer agents (Gob), following up on the prior CI triage at https://github.com/hao-ai-lab/FastVideo/pull/1373#issuecomment-4517092186. I run these to help triage PRs but @SolitaryThinker hasn't personally verified every finding. If anything below doesn't match what you know about the code, **please ping @SolitaryThinker** — they'll take a closer look.

## TL;DR

I pushed the PR-scoped yapf fix as `6670e23ef6735fd54d6da30346d03cb5ffb5dccb`; the GitHub Actions `pre-commit` workflow is now **green** at the new head. Re-reviewing the code against the earlier missing-backward triage: the PR does **not** add a full FP4-style `register_autograd` backward, but it does implement the alternate safe path — grad-enabled calls that need autograd bypass the custom op and call the original FlashAttention autograd function, while no-grad inference still uses the traceable custom op. The new tests cover inference equivalence, backward gradient flow via `torch.autograd.grad(...)`, and an opcheck smoke. One non-blocking placement finding below (S2): the new custom-op registration should live in `attention/utils/`, not in the backend dispatcher file.

- S0 (blockers): 0
- S1 (must-fix): 0
- S2 (should-fix; surfaced here): 1 (file placement vs FP4 + vLLM precedent)
- S3 (discussion): 3 (PR-body cleanup only; see archived `review.md`)

Merge note: the formatting push re-triggered Buildkite; the three training checks were still pending when I captured this review, and `microscope-transformer-tests` flipped to FAILURE on the new commit. The yapf change was a pure single-line `if` collapse with no semantic effect, and the same code passed transformer-tests cleanly on the prior build (3239) — high-confidence flake. Recommend `/test transformer` to rerun before merge; I don't see a code-level blocker in the diff.

---

## Differential re-review notes (no S0/S1)

- Prior triage gap: missing custom-op backward made the training jobs fail. Current code fixes that by routing grad-enabled q/k/v calls to `_fa_default(...)` (`fastvideo/attention/backends/flash_attn.py:75-88`) instead of through the forward-only custom op.
- There is intentionally no `register_autograd` / `setup_context` for `fastvideo::_flash_attn_default_forward`; this PR is the conditional-bypass variant, not the full FP4-template mirror. Reasonable scope-reducing choice; it does mean training-path inference takes a graph break at this call site (the inference path keeps the traceable op).
- `register_fake` matches the actual single-tensor forward schema (`flash_attn.py:53-73`). No `(out, softmax_lse)` tuple is introduced.
- Backward coverage exists in `test_default_compilable_training_backward_flows` (`fastvideo/tests/attention/test_flash_attn_default_custom_op.py:84-112`), comparing q/k/v gradients from the dispatcher against the original FlashAttention function.
- Opcheck smoke exists in `test_default_forward_opcheck` (`fastvideo/tests/attention/test_flash_attn_default_custom_op.py:115-121`). Note the opcheck call uses `requires_grad=False` inputs — that's deliberate (no autograd to check on this op by design) but worth a one-line comment so future readers don't read it as a missing test case.

---

### [S2-persistent] Move the custom-op registration from `attention/backends/` to `attention/utils/`

**What:** The `@torch.library.custom_op("fastvideo::_flash_attn_default_forward", ...)` + `@torch.library.register_fake` + the `flash_attn_func_compilable` wrapper currently live inline at `fastvideo/attention/backends/flash_attn.py:48-88`, mixed with the backend dispatch logic in the same file. The FastVideo precedent (`fastvideo/attention/utils/flash_attn_cute.py`) and the vLLM precedent (`vllm/utils/flashinfer.py`) both put per-kernel custom-op wrappers in a sibling `utils/` directory, not in the backend dispatcher.

**Why it matters:** Three reasons, ranked by weight:

1. **Internal consistency with FP4-cute.** The PR body says it "mirrors the existing FP4 `flash_attn_cute` custom-op template", but the FP4 wrapper lives at `fastvideo/attention/utils/flash_attn_cute.py` — a separate file in `utils/`. Putting this PR's wrapper in `backends/flash_attn.py` breaks the mirror it claims. `fastvideo/attention/AGENTS.md` describes `backends/` as backend registry/selector files and `utils/` as helper space — inline registration in the backend file mixes concerns.
2. **External precedent.** vLLM keeps third-party-kernel custom-op wrappers in `vllm/utils/flashinfer.py` ([link](https://github.com/vllm-project/vllm/blob/main/vllm/utils/flashinfer.py)), separate from the FA dispatch at `vllm/v1/attention/backends/fa_utils.py`. Same shape as the FastVideo FP4 precedent.
3. **Reusability.** A future backend (e.g. SageAttn, SDPA) wanting to wrap its external kernel in a `fastvideo::...` traceable op can import from `utils/`, but can't cleanly import the registration from another backend's dispatcher file.

**Suggested fix:** Move the registration block + `flash_attn_func_compilable` to a new sibling file `fastvideo/attention/utils/flash_attn_default.py`. Keep `backends/flash_attn.py` to:

```python
from fastvideo.attention.utils.flash_attn_default import flash_attn_func_compilable
```

The diff is small (move the ~40-line block, add the import) and the runtime/CI semantics are identical. Watch out for one concrete risk: do NOT leave the inline registration in `backends/flash_attn.py` after the move — duplicate `torch.library.custom_op` registration with the same qualified name will error at import time. A single-grep verify (`grep -rn "_flash_attn_default_forward" fastvideo/`) after the move catches this.

Optionally, since `flash_attn_cute.py` already lives in `utils/`, you could merge into it instead of creating a new file — but `flash_attn_cute.py` raises ImportError when CuTe isn't available, so keeping them separate avoids tying the FA2/FA3 default path to CuTe-availability. New-file is cleaner.

**Evidence:**
- Current placement: `fastvideo/attention/backends/flash_attn.py:48-88`
- FastVideo FP4 precedent: `fastvideo/attention/utils/flash_attn_cute.py:16` (custom_op), `:44` (register_fake), `:93` (register_autograd)
- AGENTS.md rule: `fastvideo/attention/AGENTS.md` describes backends/ vs utils/ separation
- vLLM precedent: https://github.com/vllm-project/vllm/blob/main/vllm/utils/flashinfer.py

---

Non-blocking cleanup I'd still recommend (S3, not separate findings): update the PR body to mention the conditional autograd bypass explicitly, add the new backward/opcheck test evidence, and remove the generated-tool footer from the PR description per project norms.

A deeper architectural question I noticed but am NOT asking for in this PR: vLLM wraps **the whole attention forward** as one opaque `torch.ops.vllm.unified_attention_with_output` op (see [their design doc](https://github.com/vllm-project/vllm/blob/main/docs/design/torch_compile.md)), rather than wrapping individual third-party kernels. That granularity choice would obviate per-backend custom_op wrapping entirely. Out of scope for this PR; flagging for a possible follow-up design discussion.

---

*Review above is from @SolitaryThinker's agent **Gob** (an AI reviewer). Ping @SolitaryThinker if anything is off, contradicts your intent, or applies to a stale rebase. Full review (including S3 notes) is archived locally and available on request.*

PR #1375, opened 2026-05-20
[perf]: wire in-worker profiler regions for inference forward

Mister-Raggs:
## What

`FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_*` previously produced no trace for inference: the three documented regions (`pre_denoising`, `denoising`, `post_denoising`) were commented out in `fastvideo/profiler.py`, `ComposedPipelineBase.forward` didn't bracket the stage loop, and `TorchProfilerConfig.from_env()` only exact-matched tokens (so even after the regions were registered, the wildcard would still fail). The in-worker profiler infrastructure worked for *training* and *model loading* but silently no-op'd for *inference* — which is where most users want to use it.

## Change

- **Register** the three inference regions in `fastvideo/profiler.py` (uncomment the existing definitions).
- **Bracket** the stage loop in `ComposedPipelineBase.forward` by the **denoise span** rather than a single stage. The project registers several denoise stages under the same convention — `denoising_stage` (canonical), `ltx2_refine_denoising_stage` (LTX-2 refine pass), `sr_denoising_stage` (Hunyuan-1.5 SR, Magi-Human SR). `_find_denoise_stage_span` substring-matches `"denois"` and returns `(first_idx, last_idx)` so:
  - stages before the first denoise → `profiler_region_inference_pre_denoising`
  - the inclusive span from first to last denoise → `profiler_region_inference_denoising` (refine + any intermediate prep step land here, by design)
  - stages after the last denoise → `profiler_region_inference_post_denoising`
- **Glob support** in `TorchProfilerConfig.from_env()`: each token in `FASTVIDEO_TORCH_PROFILE_REGIONS` is tried as an exact match first (preserves all existing behaviour for concrete region names), then on miss falls through to `fnmatch.fnmatchcase` against the registered names. Unknown tokens still warn. Makes the wildcard usage (`profiler_region_inference_*`) actually work.
- **Fallback:** pipelines that register no denoise stage at all (rare — some preprocess pipelines override `forward` entirely) wrap the whole loop under `denoising`.

## Zero overhead when not used

Each `controller.region(...)` is an early-`yield` no-op when its region isn't enabled (see `TorchProfilerController.region` in `fastvideo/profiler.py`). So this is a pure infrastructure unlock — no behavior change for users who don't set `FASTVIDEO_TORCH_PROFILE_REGIONS`.

## Usage examples

All three inference regions:
```bash
FASTVIDEO_TORCH_PROFILER_DIR=/tmp/trace \
FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_* \
python your_inference_script.py
```

Just the denoise loop:
```bash
FASTVIDEO_TORCH_PROFILER_DIR=/tmp/trace \
FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_denoising \
python your_inference_script.py
```

Mix exact + glob:
```bash
FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_model_loading,profiler_region_inference_*
```

A TensorBoard trace lands in `$FASTVIDEO_TORCH_PROFILER_DIR`.

## Test plan

- [x] Pre-commit (yapf / ruff / codespell / mypy) passing.
- [x] No behavior change when `FASTVIDEO_TORCH_PROFILE_REGIONS` is unset (regions are early-`yield` no-ops).
- [x] Exact-match tokens behave identically (glob is a fall-through).
- [ ] Manual smoke: run a Wan / Cosmos inference with `profiler_region_inference_*` and confirm the produced trace contains the three `fastvideo.region::profiler_region_inference_*` annotations around the appropriate stages (pending GPU).
- [ ] Manual smoke on an LTX-2 refine pipeline to confirm `ltx2_refine_denoising_stage` lands inside `denoising`, not `post_denoising` (pending GPU).

### DIFF
--- a/fastvideo/pipelines/composed_pipeline_base.py
+++ b/fastvideo/pipelines/composed_pipeline_base.py
@@ -504,8 +504,35 @@ class ComposedPipelineBase(ABC):
         # Execute each stage
         logger.info("Running pipeline stages: %s", self._stage_name_mapping.keys())
         # logger.info("Batch: %s", batch)
-        for stage in self.stages:
-            batch = stage(batch, fastvideo_args)
+        # Bracket the loop with the three documented inference regions so
+        # `FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_*`
+        # actually traces inference. The denoise stage is identified by
+        # the project-wide convention `add_stage(stage_name="denoising_
+        # stage", ...)`; everything before it counts as pre-denoising and
+        # everything after as post-denoising. If a pipeline does not
+        # register a "denoising_stage" (rare; some preprocess pipelines
+        # override forward entirely), fall back to wrapping the whole
+        # loop under `denoising` so the env-var still produces a trace
+        # rather than silently doing nothing.
+        # The regions are no-ops at zero overhead when not enabled (see
+        # `TorchProfilerController.region`'s early `yield`).
+        denoise_idx = next(
+            (i for i, s in enumerate(self.stages) if getattr(s, "_pipeline_stage_name", "") == "denoising_stage"),
+            None,
+        )
+        if denoise_idx is None:
+            with self.profiler_controller.region("profiler_region_inference_denoising"):
+                for stage in self.stages:
+                    batch = stage(batch, fastvideo_args)
+        else:
+            with self.profiler_controller.region("profiler_region_inference_pre_denoising"):
+                for stage in self.stages[:denoise_idx]:
+                    batch = stage(batch, fastvideo_args)
+            with self.profiler_controller.region("profiler_region_inference_denoising"):
+                batch = self.stages[denoise_idx](batch, fastvideo_args)
+            with self.profiler_controller.region("profiler_region_inference_post_denoising"):
+                for stage in self.stages[denoise_idx + 1:]:
+                    batch = stage(batch, fastvideo_args)
 
         # Return the output
         return batch
--- a/fastvideo/pipelines/composed_pipeline_base.py
+++ b/fastvideo/pipelines/composed_pipeline_base.py
@@ -482,16 +482,27 @@ class ComposedPipelineBase(ABC):
             if self.local_rank == 0:
                 print(self.profiler.key_averages().table(sort_by="self_cuda_time_total"))
 
-    def _find_denoise_stage_index(self) -> int | None:
-        """Index of the canonical ``"denoising_stage"`` in
-        ``self.stages``, or ``None`` if no stage by that name is
-        registered (rare — some preprocess pipelines override
-        ``forward`` entirely and never call ``add_stage``).
+    def _find_denoise_stage_span(self) -> tuple[int | None, int | None]:
+        """Return ``(first_idx, last_idx)`` (inclusive) of the contiguous
+        denoising span in ``self.stages``, or ``(None, None)`` if no
+        denoising stage is registered.
+
+        The project uses several denoise stage names (``denoising_stage``
+        is the canonical one but pipelines also register
+        ``ltx2_refine_denoising_stage`` (LTX-2 refine pass) and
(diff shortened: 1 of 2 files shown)


SolitaryThinker:
the profiler never supported using glob to select regions. ie. `FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_*`

is this PR intended to fix this or just to add in more profiler regions?

SolitaryThinker:
**Hi @Mister-Raggs** — this is a code review from one of @SolitaryThinker's AI reviewer agents (Gob). I run these to help triage PRs but @SolitaryThinker hasn't personally verified every finding — **please ping @SolitaryThinker** if anything below is off.

## TL;DR

This PR is well scoped and wires the inference profiler regions in the right shared path. I think it needs changes before merge because the advertised `profiler_region_inference_*` usage still fails in the env parser, and multi-denoise pipelines can be traced with later denoising work mislabeled as post-denoising.

- S0 (blockers): 0
- S1 (must-fix): 1
- S2 (should-fix; persistent or important): 1
- S3 (discussion): 2; not shown here, see `review.md`

---

## Findings (must-fix + persistent should-fix only)

### [S1] Wildcard inference region selection is advertised but still rejected

**What:** The PR comments and stated use case rely on `FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_*`, but `TorchProfilerConfig.from_env()` only splits comma-separated tokens and exact-matches each token with `resolve_profiler_region(token)`. A literal `profiler_region_inference_*` token does not match any registered region, so if it is the only token, profiler initialization raises instead of enabling the three inference regions.

**Why it matters:** The headline usage path still fails. Users following the wildcard guidance get no inference trace even though the regions are now registered.

**Suggested fix:** Implement glob expansion for wildcard tokens in `TorchProfilerConfig.from_env()`, or remove the wildcard claim and require the explicit comma-separated list of all three inference region names. Since this PR is framed around `profiler_region_inference_*`, I recommend adding wildcard expansion.

**Evidence:** `fastvideo/pipelines/composed_pipeline_base.py:518-520` advertises `FASTVIDEO_TORCH_PROFILE_REGIONS=profiler_region_inference_*`; `fastvideo/profiler.py:236-259` exact-matches requested tokens and raises if none match; `fastvideo/profiler.py:127-138` registers only the exact inference region names.

---

### [S2] Later denoising stages are misclassified as post-denoising

**What:** `_find_denoise_stage_index()` only returns the first stage named exactly `"denoising_stage"`, then `ComposedPipelineBase.forward()` wraps every later stage in `profiler_region_inference_post_denoising`. In LTX2 refine mode, a later `"ltx2_refine_denoising_stage"` is therefore traced as post-denoising.

**Why it matters:** `post_denoising` is described as post-processing after denoising, but this boundary can include real transformer denoising work. That makes the trace misleading for refine or multi-stage-denoise pipelines.

**Suggested fix:** If `denoising` should mean all denoising work, classify a denoising span rather than a single exact stage name, for example by matching denoising stage-name conventions with a small allowlist. If it should mean only the first/main denoise, make that limitation explicit in the region description and comments.

**Evidence:** `fastvideo/pipelines/composed_pipeline_base.py:485-494` only finds the exact `"denoising_stage"`; `fastvideo/pipelines/composed_pipeline_base.py:543-545` wraps all later stages as post-denoising; `fastvideo/pipelines/basic/ltx2/ltx2_pipeline.py:63-66` adds the main denoising stage and `fastvideo/pipelines/basic/ltx2/ltx2_pipeline.py:120-129` adds `"ltx2_refine_denoising_stage"` later.

---

Mister-Raggs:
Thanks @SolitaryThinker, the wildcard was not actually implemented. from_env() is exact-match only. Original idea was to add more regions.

Easiest fix: add glob expansion in from_env() exact match first, fall through to fnmatch.fnmatchcase against the registered names, then warn. ~15 LOC. I'll push the fixes onto this branch and you can take a look.

PR #1389, opened 2026-05-23
[ci] Modal: enable Flash Attention 3 on H100 perf tests with build cache

rich7420:
## Purpose

Enable Flash Attention 3 (Hopper sm90+) on Modal. The existing perf job
`pr_test.py::run_performance_tests` runs on L40S:2 (sm89) and can't use
FA3, so this adds an opt-in H100:2 entrypoint that installs and caches
the FA3 build.

## Changes

One function added to `fastvideo/tests/modal/pr_test.py`
(`run_performance_tests_h100`) plus a `fa3-build-cache` Modal Volume.
Purely additive — no existing functions, tests, or CI wiring touched.

**Modal-only, no pod / no Docker rebuild.** FA3 is installed at runtime
inside the H100:2 container, tried in order: skip if already present →
restore from the Modal Volume cache → else source-build from
`Dao-AILab/flash-attention/hopper` and save to the cache. The cache key is the
torch+CUDA+python ABI (read at runtime), so a torch/CUDA bump misses the cache
and rebuilds instead of restoring an ABI-incompatible binary. The staging copy
captures `flash_attn_interface` + `flash_attn_config` + the `flash_attn_3`
package — `flash_attn_config` is lazily imported, so omitting it would still
pass import/verify but crash at the first FA3 forward. After install it asserts
`fa_version == '3'` so a silent FA2 fallback fails the job.

| install path | time |
|---|---|
| cold source build (no cache) | ~90 min |
| warm cache restore | ~30 s |

## Test Plan

```bash
export HF_API_KEY=hf_...
export BUILDKITE_REPO=https://github.com/<fork>/fastvideo.git
export BUILDKITE_COMMIT=<sha>
export IMAGE_VERSION=py3.12-latest
modal run --detach fastvideo/tests/modal/pr_test.py::run_performance_tests_h100
```

## Test Results

Modal H100:2, `fastvideo-dev:py3.12-latest` — the FA3 perf path runs and
passes end-to-end:

```
[fa3] restored from cache → fa_version == '3'  (assertion passed)
test_inference_performance[wan-t2v-1.3b-2gpu] PASSED
avg_time=5.92s (5 runs 5.87–5.97s), peak 9516MB — 1 passed in 117.80s
```

The ABI cache key was verified on the same image: `FA3_ABI` resolves to
`py312-torch2.10.0+cu128-cu12.8` → `/fa3_cache/fa3-py312-torch2.10.0+cu128-cu12.8`,
distinct from the old tag-based dir, so the first run after this change does a
one-time rebuild and then warm-restores.

## Checklist

- [x] `pre-commit run --files fastvideo/tests/modal/pr_test.py` (path
      excluded from yapf/ruff/mypy by repo config; filename check passed)
- [x] Purely additive; docs are inline (docstring + comments)
- [x] No GPU memory regression (peak 9516MB, within default threshold)


SolitaryThinker:

### S1: FA3 build cache key can reuse a stale binary

`run_performance_tests_h100` restores `/fa3_cache/site-packages-${IMAGE_VERSION:-unknown}` whenever `.INSTALLED` exists, but the cache key only contains `IMAGE_VERSION` while the cold build clones unpinned `flash-attention` HEAD. That omits inputs that affect the compiled extension, especially the FA3 commit SHA and runtime ABI (`torch.__version__`, `torch.version.cuda`, and SM/GPU capability), so a warm run can silently copy an old `flash_attn_interface` binary after any of those inputs change. Please pin the FA3 commit and derive the cache dir/marker from a hash of the pinned FA3 SHA plus torch/CUDA/image/GPU inputs, then validate that metadata on restore before running the perf suite.

SolitaryThinker:
Current-head maintainer review at `88dac27d` against main `cae8fa18`: this
opt-in path is broken after current-main changes, and the green CI does not
execute it.

- **P1 — remote checkout inputs are missing, then FA4 defeats the FA3 path.**
  `run_performance_tests_h100` attaches only the ad-hoc secret at lines 456–461,
  but the shared runner now receives `BUILDKITE_REPO` and `BUILDKITE_COMMIT`
  through `ci_env_secret`. Without it, the documented invocation reaches the
  remote container and attempts to clone an empty repository value before FA3
  setup. Attach the shared CI environment secret. That secret defaults
  `FASTVIDEO_FA4=1` (lines 82–84), so this FA3-specific entrypoint must also
  force `FASTVIDEO_FA4=0`; otherwise the backend selects FA4 and the assertion
  at lines 515–516 sees `fa_version == "4"` instead of `"3"`.

- **P2 — the binary cache is not source-reproducible.** Lines 488–510 key the
  cache only on Python/Torch/CUDA ABI while cloning unpinned
  `flash-attention` HEAD. Different cold builds can therefore create different
  binaries for the same key, while an existing cache never invalidates when
  upstream changes. Please pin the tested FA3 revision and include that
  revision in the cache key.

Buildkite 4267 proves only pre-commit and microscope-unit compatibility. The
dispatcher still maps `performance` exclusively to `run_performance_tests`,
and the performance path filter excludes `fastvideo/tests/modal/**`; no CI lane
ran this H100 function. Please refresh these points and rerun the exact Modal
H100 entrypoint on the updated head. I did not run local pytest.

PR #1343, opened 2026-05-13
[ci] add per-method single-step training tests for fastvideo.train

alexzms:
## Summary

Phase 2 / PR 5/9 of the `fastvideo.train` CI plan, split into a **5a-i** slice that establishes the per-method test pattern. A follow-up **5a-ii** will layer a device-keyed grad-norm regression on top of the same harness.

Adds two GPU smoke tests under a new `fastvideo/tests/train/methods/` directory:

* `test_wan_finetune.py` — `WanModel` + `FineTuneMethod`
* `test_wan_causal_dfsft.py` — `WanCausalModel` + `DiffusionForcingSFTMethod`

Both tests build the method via its public constructor, feed a tiny synthetic `raw_batch` (text embed + mask + vae latent), and run one end-to-end step:

```python
method.on_train_start()
loss_map, outputs, _ = method.single_train_step(batch, 0)
method.backward(loss_map, outputs, grad_accum_rounds=1)
```

The harness deliberately avoids the full `Trainer` (no FSDP wrap, no data loader, no checkpointing) so the tests stay focused on the method-level wiring that the model-loading tests in `tests/train/models` don't already cover.

## Asserts

* `loss_map["total_loss"]` is finite.
* `model.transformer.blocks[0]` has trainable params with finite, non-zero gradients.

The first transformer block's gradient is computed **last** during backprop, so a healthy grad there implies the full forward + chain-rule path is intact. Keeping the assertion surface to a single block keeps the reference data tiny for the follow-up regression PR.

## CI plumbing

* `fastvideo/tests/modal/pr_test.py`:
  * `run_train_framework_tests` pytest path now includes `./fastvideo/tests/train/methods`.
  * `run_unit_test` adds a matching `--ignore` for the new dir.
* `docs/contributing/testing.md`: mention the new `methods/` subdirectory in the Train Framework Tests entry.

## Fixtures

Two new YAMLs under `fastvideo/tests/train/fixtures/`, mirroring the existing 4/9 fixtures but with `trainable: true` and the minimum `training.distributed` / `training.optimizer` / `training.loop` keys needed by the method's `__init__` and `on_train_start`:

* `wan_t2v_finetune_min.yaml`
* `wan_causal_t2v_dfsft_min.yaml` (with `chunk_size: 3`, `num_latent_t: 6`)

## Test plan

- [x] `python -m pytest --collect-only fastvideo/tests/train/methods/` collects both tests cleanly.
[...]

PR #1396, opened 2026-05-26
[ci] layer-0 grad-norm regression for per-method training tests (5a-ii)

alexzms:
## Summary

Phase 2 / PR 5a-ii of the `fastvideo.train` CI plan. Layers a **device-keyed layer-0 grad-norm regression** on top of the per-method smoke tests landed in 5a-i (#1343).

5a-i asserts the loss is finite and block-0 grads are finite + non-zero. This PR adds a numeric fingerprint: after `single_train_step` + `backward`, the L2 norm of transformer block-0's trainable grads is compared against a per-GPU reference within `rtol`.

## Changes

- `fastvideo/tests/train/methods/grad_norm_regression.py` — helper:
  - `layer0_grad_norm(transformer)` — global L2 norm of block-0 trainable grads.
  - `resolve_device_key()` — maps the GPU name to a reference key (L40S / GB200 / H100 / …).
  - `check_grad_norm_regression(test_name, transformer, rtol=0.10)` — asserts within `rtol`; **skips** (never hard-fails) on an unsupported or not-yet-seeded GPU; with `FASTVIDEO_GRADNORM_UPDATE=1` records/refreshes the reference instead.
- `test_wan_finetune.py` / `test_wan_causal_dfsft.py` — call `check_grad_norm_regression(...)` after the existing finite/non-zero assertions.
- `fastvideo/tests/train/methods/grad_norm_refs.json` — in-repo references, seeded for **GB200** (local dev) and **L40S** (the CI runner).
- `fastvideo/tests/modal/pr_test.py` — `seed_grad_norm_references` Modal entrypoint (L40S, `FASTVIDEO_GRADNORM_UPDATE=1`) to record/refresh a runner's reference.

## Determinism & why device-keyed

`method.on_train_start()` seeds the global RNG + the method's `cuda_generator` (`training.data.seed=42` in the fixture) and the synthetic `raw_batch` is built after that call, so the grad norm is reproducible — **identical to 4 decimals across local runs**. Inputs are synthetic (`torch.randn` text-embed + VAE latent, all-ones mask); model weights are the real `Wan-AI/Wan2.1-T2V-1.3B-Diffusers` checkpoint. The check is a forward/backward **wiring + scale** fingerprint, not a real-data quality test.

References are device-keyed because the norm differs across GPU architectures (bf16 accumulation, kernel choice). The causal/dfsft test spreads **~9%** between GB200 (2.9781) and L40S (3.2562) — a single golden value couldn't cover both; `rtol=0.10`.

## Grad-alignment coverage after this PR

| model ＼ method | finetune | dfsft |
|---|---|---|
| Wan (`WanModel`) | ✅ | |
| WanCausal (`WanCausalModel`) | | ✅ |

i.e. 2 (model, method) cells: `WanModel`/`FineTuneMethod` and `WanCausalModel`/`DiffusionForcingSFTMethod`. `dmd2` / `self_forcing` / `kd` and other models (Hunyuan / Cosmos / LongCat / MatrixGame2) are out of scope here — see plan below.

## Test plan

- [x] Local (GB200): `pytest fastvideo/tests/train/methods -v` → **2 passed** (asserting against GB200 references)
- [x] L40S references seeded via `modal run fastvideo/tests/modal/pr_test.py::seed_grad_norm_references`
- [ ] `/test train-framework` — full GPU CI run on L40S (now **asserts** the grad-norm, no longer skips)

## Plan context

| Slice | Scope | Status |
|---|---|---|
| 5a-i (#1343) | per-method test framework + 2 reference tests | merged |
| **5a-ii (this PR)** | layer-0 grad-norm regression + device-keyed refs | — |
| 5b | `dmd2` / `self_forcing` / `kd` (needs tiny teacher / fake-score fixtures) | — |
| 5c | Hunyuan + finetune, more (model, method) combos | — |


fastvideo/tests/modal/pr_test.py
Davids048, line 1: This new function refresh the baseline reference grad norms. Based on the decorator, it is only doing that for L40 GPUs. but the newly created ref json file also include GB200 values, alongside those for L40 GPUs. 

Can we document how the GB 200 values are gnerated, or support it in this function too?

fastvideo/tests/train/methods/grad_norm_regression.py
```diff
@@ -0,0 +1,134 @@
...
+
+# GPU-name substring -> reference key. First match wins.
+_DEVICE_MAPPINGS: tuple[tuple[str, str], ...] = (
+    ("L40S", "L40S"),
+    ("GB200", "GB200"),
+    ("B200", "GB200"),
+    ("H100", "H100"),
+    ("H200", "H200"),
+    ("A100", "A100"),
+)
```
Davids048, line 58: Only L40s and GB200 have ref values right now. Should we clean up the other models?

Davids048 (changes requested):
Most looks good. 
Just clearer documentation on the tested configs

alexzms:
Pushed `d64e4e94` addressing the review feedback:

**@Davids048**

- **Documented how the GB200 values were obtained** — expanded the module docstring of `grad_norm_regression.py` to spell out two seeding paths: CI / L40S via `seed_grad_norm_references` in `pr_test.py`, vs. local / non-L40S GPUs via `FASTVIDEO_GRADNORM_UPDATE=1 pytest ...` on the workstation. Also clarified in the Modal function's docstring that it is pinned to `L40S:1` and seeds the `L40S` key only.
- **Cleaned up `_DEVICE_MAPPINGS`** — dropped the un-seeded `H100` / `H200` / `A100` entries; only the keys present in `grad_norm_refs.json` remain (`L40S`, `GB200`, plus `B200` → `GB200` aliasing the same Blackwell arch).

**@gemini-code-assist** (nits)

- Case-insensitive substring match in `resolve_device_key`.
- `layer0_grad_norm` now accumulates squared sums on the GPU and does a single `.item()` at the end instead of one per parameter. Same math — recorded references still apply.
- `encoding="utf-8"` on the refs JSON read/write.

Re-requesting review when you have a moment.

Davids048 (approved):
ok

alexzms:
> ok

Thanks for reviewing!

PR #1398, opened 2026-05-26
[feat] LTX-2.3 audio: BWE vocoder path

alexzms:
## Summary
Adds the LTX-2.3 audio **BWE (bandwidth-extension) vocoder** path to `ltx2_audio_vae.py`. Additive + backward-compatible: it only activates when a vocoder config contains a `"bwe"` key; the existing non-BWE `Vocoder` path is behavior-identical.

- New: `VocoderWithBWE`, `MelSTFT` / `_STFTFn`, `Snake` / `SnakeBeta`, `AMPBlock1`, anti-alias resampling (`LowPassFilter1d` / `UpSample1d` / `DownSample1d` / `Activation1d`, `kaiser_sinc_filter1d`).
- `VocoderConfigurator.from_config` gains the `if "bwe" in vocoder_cfg` branch (16 kHz base + BWE generator + mel re-analysis → 48 kHz).
- Audio decoding stage: `_resolve_audio_sample_rate` (handles 48 kHz BWE vs the 24 kHz default).

## Backward compatibility
Non-BWE `Vocoder` is unchanged: the in-loop `leaky_relu(x, LRELU_SLOPE)` is gated `if not is_amp`, and the final path stays `leaky_relu() → conv_post → tanh` (verified against the original forward). BWE classes are inert unless the checkpoint config carries a `bwe` block.

## Test plan
- [x] `py_compile` + import (`VocoderWithBWE` present)
- [x] Existing LTX2 tests pass on this branch: `test_ltx2_continuation` + `test_ltx2_stage_overrides` (29)
- [ ] Functional BWE decode on a 2.3 checkpoint (needs GPU validation)

Independent of the DiT-2.3 PR; can be reviewed/merged separately.

### DIFF
--- a/fastvideo/models/audio/ltx2_audio_vae.py
+++ b/fastvideo/models/audio/ltx2_audio_vae.py
@@ -1534,14 +1534,100 @@ class Vocoder(nn.Module):
 # =============================================================================
 
 
+def _build_stft_basis(filter_length: int,
+                      win_length: int) -> torch.Tensor:
+    """Hann-windowed FFT basis used by ``_STFTFn`` as a Conv1d kernel.
+
+    Returns a ``(2*(filter_length//2 + 1), 1, filter_length)`` float32 tensor
+    that ``F.conv1d`` projects a 1-D waveform onto, producing real (rows
+    ``[:n_freqs]``) and imaginary (rows ``[n_freqs:]``) STFT coefficients.
+
+    Initialising the buffer deterministically here, rather than relying on a
+    ``register_buffer(torch.zeros(...))`` placeholder, makes the BWE mel
+    front-end safe against ``VocoderLoader``'s ``strict=False`` load: a
+    checkpoint that omits ``forward_basis`` no longer leaves the buffer at
+    zero (which would silently feed a constant-zero magnitude into BWE).
+    """
+    n_freqs = filter_length // 2 + 1
+    # FFT matrix: row k is e^{-2*pi*j*k*n / filter_length} for n in [0, filter_length)
+    fft_matrix = torch.fft.fft(torch.eye(filter_length, dtype=torch.float32))
+    fft_basis = fft_matrix[:n_freqs]  # (n_freqs, filter_length) complex
+    window = torch.hann_window(win_length,
+                               periodic=True,
+                               dtype=torch.float32)
+    if win_length < filter_length:
+        pad_left = (filter_length - win_length) // 2
+        pad_right = filter_length - win_length - pad_left
+        window = F.pad(window, (pad_left, pad_right))
+    elif win_length > filter_length:
+        window = window[:filter_length]
+    windowed_basis = fft_basis * window.unsqueeze(0)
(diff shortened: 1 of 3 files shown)


SolitaryThinker:

Thanks for the focused BWE vocoder patch. I found two S1 issues that should be fixed before merge:

1. **S1: BWE mel front-end can silently run with all-zero signal-processing buffers.** In `fastvideo/models/audio/ltx2_audio_vae.py`, `_STFTFn` registers `forward_basis` as zeros and immediately uses it for `F.conv1d` (`lines 1543, 1551`), and `MelSTFT` registers `mel_basis` as zeros and uses it for the mel projection (`lines 1570, 1577`). The new BWE path then calls this from `VocoderWithBWE.forward()` before the residual generator (`lines 1633-1635`). Since `VocoderLoader` loads the vocoder with `strict=False` (`fastvideo/models/loader/component_loader.py:856-857`), a converted checkpoint that omits these deterministic buffers will not fail; it will produce a constant log-mel input for BWE instead. Please initialize the Fourier/mel bases deterministically in the constructors, or validate that those buffers were loaded before enabling the BWE path.

2. **S1: The new BWE vocoder path has no CI-tracked unit/parity coverage.** The PR changes only `fastvideo/models/audio/ltx2_audio_vae.py` and `fastvideo/pipelines/basic/ltx2/stages/ltx2_audio_decoding.py`; there is no matching test under `fastvideo/tests/`, and the existing `tests/local_tests/ltx2/test_ltx2_audio_vae.py` is skip-heavy/local-only. Please add at least a small `fastvideo/tests/` test for the BWE config path that would catch missing STFT/mel buffers and verify waveform shape/sample-rate behavior.

alexzms:
Pushed `07903b2b` to address both S1 findings.

**S1 #1 (zero-init buffers).** Added `_build_stft_basis` (Hann-windowed FFT basis as a Conv1d kernel) and `_build_mel_basis` (Slaney-normalised triangular mel filterbank, librosa-compatible defaults), and call them in `_STFTFn.__init__` and `MelSTFT.__init__`. `MelSTFT` now requires a `sampling_rate` argument; the vocoder factory site (`ltx2_audio_vae.py:1783`) passes `bwe_cfg["input_sampling_rate"]` (default 16000). A loaded checkpoint still overrides the deterministic init exactly as before — the only behavioural change is that an *omitted* buffer no longer silently zeros the STFT/mel front-end, so a converted checkpoint that drops these keys is no longer a silent-failure path.

**S1 #2 (no CI coverage).** Added `fastvideo/tests/audio/test_ltx2_bwe.py` (11 CPU-only regression tests) covering forward / mel basis non-zero invariants, shape, sampling-rate dependence, mel filterbank non-negativity, an end-to-end log-mel of a 1 kHz sine tone not being constant (the regression assertion for the silent-zero bug — would fail if either basis were re-introduced as zero), and the standalone basis helpers. All pass.

SolitaryThinker (approved):

PR #1408, opened 2026-05-27
[kernel] Extract VSA utility functions into fastvideo_kernel

### LINKED ISSUE #782: [Feature] Utility class/functions for VSA
### Motivation

Currently the https://github.com/hao-ai-lab/FastVideo/blob/main/csrc/attn/video_sparse_attn/README.md readme lacks a proper example of how to plug n play vsa for a training free comparison with flash attention or sage, would be useful to have a utility class with relevant build function for vsa metadata, or even better a separate class that encapsulate things like     

```
tile_partition_indices = _get_tile_partition_indices(dit_seq_shape, tile_size, device_str)
reverse_tile_partition_indices = _get_reverse_tile_partition_indices(dit_seq_shape, tile_size, device_str)
variable_block_sizes = _construct_variable_block_sizes(dit_seq_shape, num_tiles, device_str)
non_pad_index = _get_non_pad_index(tuple(variable_block_sizes.tolist()), math.prod(tile_size), device_str)
```
that in forward video_sparse_attn is used and can easily be put in place of flash attention 3 in the training-free approach

### Related resources

_No response_

freemty:
## Summary

- Extract 4 tile-partition/variable-block-size utility functions from `fastvideo/attention/backends/video_sparse_attn.py` into standalone `fastvideo_kernel/vsa_utils.py`
- Add `build_vsa_metadata()` convenience entry point for one-call metadata construction
- Enable external users to call VSA kernels without depending on the full fastvideo framework (`pip install fastvideo-kernel` is now self-contained)

## Motivation

The VSA kernel lives in `fastvideo_kernel` (independently installable), but calling it requires helper functions buried in the `fastvideo` framework layer. This forces external users to install the entire framework just for metadata construction. See #782.

## Changes

| File | Description |
|------|-------------|
| `fastvideo_kernel/vsa_utils.py` | New module: 5 functions + `VSA_TILE_SIZE` constant (~147 lines) |
| `fastvideo_kernel/__init__.py` | Re-export new utilities |
| `tests/test_vsa_utils.py` | 20 CPU-only tests covering all functions + consistency with framework |

## Design decisions

- **Parameterized `tile_size`**: `construct_variable_block_sizes` accepts `tile_size` as argument instead of hardcoding `VSA_TILE_SIZE`, enabling non-default tile configurations
- **`lru_cache` on pure functions**: index computations are deterministic for a given shape — caching avoids redundant recomputation
- **CPU-only tests**: utilities are pure PyTorch index math, no GPU/kernel required
- **Consistency test**: verifies output matches the framework-level functions exactly (skipped if `fastvideo` is not installed)

## Test plan

- [x] `pytest fastvideo-kernel/tests/test_vsa_utils.py -v` — 20 tests, all pass (CPU-only)
- [ ] CI pre-commit checks
- [ ] Verify framework-level import still works after downstream refactor (future PR)

Closes #782

### DIFF
--- a/fastvideo-kernel/python/fastvideo_kernel/__init__.py
+++ b/fastvideo-kernel/python/fastvideo_kernel/__init__.py
@@ -29,6 +29,15 @@ from fastvideo_kernel.block_sparse_attn_varlen import (
     block_sparse_attn_varlen,
 )
 
+from fastvideo_kernel.vsa_utils import (
+    VSA_TILE_SIZE,
+    get_tile_partition_indices,
+    get_reverse_tile_partition_indices,
+    construct_variable_block_sizes,
+    get_non_pad_index,
+    build_vsa_metadata,
+)
+
 __all__ = [
     "sliding_tile_attention",
     "video_sparse_attn",
@@ -44,5 +53,11 @@ __all__ = [
     "FastLayerNorm",
     "int8_linear",
     "int8_quant",
+    "VSA_TILE_SIZE",
+    "get_tile_partition_indices",
+    "get_reverse_tile_partition_indices",
+    "construct_variable_block_sizes",
+    "get_non_pad_index",
+    "build_vsa_metadata",
     "__version__",
 ]
--- a/fastvideo-kernel/python/fastvideo_kernel/vsa_utils.py
+++ b/fastvideo-kernel/python/fastvideo_kernel/vsa_utils.py
@@ -85,7 +85,6 @@ def construct_variable_block_sizes(
     ).reshape(-1)
 
 
-@functools.lru_cache(maxsize=10)
 def get_non_pad_index(
     variable_block_sizes: torch.LongTensor,
     max_block_size: int,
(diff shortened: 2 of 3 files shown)


alexzms (commented):
Quick question on the dict keys returned by `build_vsa_metadata()` in `fastvideo-kernel/python/fastvideo_kernel/vsa_utils.py`:

- The dict uses `tile_indices` and `reverse_tile_indices`.
- The framework `VideoSparseAttentionMetadata` dataclass in `fastvideo/attention/backends/video_sparse_attn.py` (lines 138–139) uses the longer names `tile_partition_indices` and `reverse_tile_partition_indices`.

Since the follow-up PR will swap the framework over to importing from here, was there a reason to shorten the names? If not, aligning them now (using `tile_partition_indices` / `reverse_tile_partition_indices` in the returned dict too) would let the follow-up be a clean `import ... as ...` swap rather than touching every call site that destructures the metadata.

Happy to be wrong if you meant this as a cleaner external-facing API — just want to flag it before merge.

freemty:
Good catch! You're right — aligning the dict keys now avoids unnecessary churn in the follow-up PR. 

Updated `build_vsa_metadata()` to return `tile_partition_indices` / `reverse_tile_partition_indices` (matching `VideoSparseAttentionMetadata`). Also removed the `@lru_cache` on `get_non_pad_index` since it accepts a `torch.Tensor` which is unhashable (would raise `TypeError` at runtime).

Tests updated and passing.

PR #1410, opened 2026-05-27
[perf] training: add attn_only selective activation checkpointing

rich7420:
## Motivation

Video DiT training has to run with activation checkpointing — at video sequence
lengths (e.g. 72k tokens) the activations don't fit otherwise. But the existing
modes leave a gap:

- **`full`** recomputes the *entire* transformer block in backward, so it
  re-runs the **attention forward** — the single most expensive op — a second
  time. On Wan2.1-T2V-1.3B we measured the attention forward kernel running
  **1200×** under `full` where only **600×** are needed; the other 600 are pure
  redundant recompute, ~12% of GPU compute burned every step.
- **`ops`** (the stock selective mode) is meant to avoid that, but it **OOMs**
  at video sequence lengths because it also saves the FFN/GEMM intermediates
  (~51 GB across the blocks) — which are large but *cheap* to recompute.

So in practice there is no usable middle option: recompute everything (`full`,
wastes the attention recompute) or OOM (`ops`).

## What this adds

`attn_only`: a per-block selective policy that **saves only the attention
forward output** (small — ~221 MB/block at seq 72k) and any functional
collective output (so communication is never re-issued in backward), and
**recomputes everything else**. It removes the attention recompute `full` pays,
while staying within memory.

## Why this lever

- **Removes real work (FLOPs), not host-wait** → the speedup is
  fabric-independent (helps on both PCIe and NVSwitch) and **stacks with
  `torch.compile`** (which won't merge separate `nn.Linear` weights or change
  the recompute structure).
- **Near-zero cost**: +2 GB peak vs `full`; numerically equivalent (saves the
  forward output rather than recomputing it).
- **Last recoverable AC win**: after `attn_only`, the dominant kernel is the
  flash *backward* (gradient computation — irreducible). This captures the
  remaining activation-checkpoint headroom.

`activation_checkpoint.py` is shared by both the legacy (`fastvideo/training/`)
and modular (`fastvideo/train/`) stacks, so both benefit from one change.

## Opt-in, default-preserving

The default is unchanged (no AC unless requested). `attn_only` joins
`full`/`ops`/`block_skip` as a new point on the memory↔speed curve — it is **not**
a replacement for `full`, which remains the lowest-memory option for
memory-constrained training.

| mode      | peak (1×H100) | note                          |
|-----------|---------------|-------------------------------|
| full      | 26.8 GB       | recompute all — lowest memory |
| attn_only | 28.9 GB       | +2 GB, no attention recompute |
| ops       | OOM           | also saves the FFN mm         |

Use `attn_only` when you have headroom over `full`.

## Changes

- `CheckpointType.ATTN_ONLY` + dispatch branch
- `_apply_activation_checkpointing_attn_only`: MUST_SAVE the attention forward
  output and any `_c10d_functional` collective (so a collective is never
  re-issued during the backward recompute — that is expensive and a cross-rank
  ordering hazard), PREFER_RECOMPUTE the rest; the per-op decision is cached
- `_is_attention_forward`: backend-agnostic op-name match (FA2
  `_flash_attn_forward`, FA3 `fwd`, aten SDPA; backward excluded) — robust to
  the runtime-selected attention backend
- CLI: add `attn_only` to `--enable-gradient-checkpointing-type` choices

## Results

**Kernel-level** (nsys, 1×H100, Wan2.1-T2V-1.3B 720×1280×77f) — the direct,
hardware-independent measurement: the only kernel whose count changes is
`flash_fwd_kernel` (**1200 → 600**, −4.67 s); flash-bwd / GEMM / elementwise
counts are conserved. Total GPU compute **40.2 s → 35.3 s (−12.3%)**, ~95% of
which is the flash-fwd halving — i.e. the change removes exactly the
attention-forward recompute and nothing else.

**Wall-clock step time** (steady-state, no profiler):

| config | full   | attn_only | speedup |
|--------|--------|-----------|---------|
| H100×1 | 4.19 s | 3.75 s    | −10.5%  |
| H100×2 | 2.25 s | 2.08 s    |  −7.6%  |

H100×2 reproduced across two independent runs (2.28→2.09 in the other). It also speeds up multi-GPU L40S,
but Modal's per-job L40S allocation has large PCIe-topology variance (identical
work ranged 5.5–9.1 s/step), so we don't quote a clean L40S wall-time delta — the
−12.3% compute figure above is the allocation-independent one. The win narrows as
GPU count grows (fixed comm cost dilutes the compute saving) and scales with the
attention fraction of the step — largest for long-sequence video training, which
is why it is opt-in rather than a forced default.

**Numerically equivalent**: per-step loss differs from `full` within the
GPU non-determinism noise floor (full-vs-full on the same seed already differs by
~0.002 from flash-attn backward atomics). By construction it saves the forward
output rather than recomputing it, so it introduces no systematic drift.

## Usage

```
--enable_gradient_checkpointing_type attn_only
```

## Test plan

- **[added]** CPU unit test (`fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py`):
  `CheckpointType.ATTN_ONLY`, the attention-forward classifier
  (FA2/FA3/SDPA forward matched; backward and FFN/QKV `mm` excluded), the
  save decision (attention forward + functional collectives saved, GEMM
  recomputed) and its caching, per-block `checkpoint_wrapper` application, and
  the error paths. No GPU required.
- **[manual]** e2e training runs clean with `attn_only` on 1–2×H100 and 2×L40S,
  20 steps, no OOM.

### DIFF
--- a/fastvideo/fastvideo_args.py
+++ b/fastvideo/fastvideo_args.py
@@ -1099,7 +1099,7 @@ class TrainingArgs(FastVideoArgs):
         parser.add_argument("--max-grad-norm", type=float, help="Maximum gradient norm")
         parser.add_argument("--enable-gradient-checkpointing-type",
                             type=str,
-                            choices=["full", "ops", "block_skip"],
+                            choices=["full", "ops", "block_skip", "attn_only"],
                             default=None,
                             help="Gradient checkpointing type")
         parser.add_argument("--selective-checkpointing", type=float, help="Selective checkpointing threshold")
--- a/fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py
+++ b/fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py
@@ -1,0 +1,87 @@
+"""Unit tests for the ``attn_only`` selective activation-checkpointing policy.
+
+CPU-only: exercises the op classifier and the per-block wrapping structure
+without running a forward/backward (so no GPU or distributed init needed).
+"""
+import torch
+from torch import nn
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (CheckpointWrapper)
+
+from fastvideo.training.activation_checkpoint import (CheckpointType, _is_attention_forward,
+                                                      apply_activation_checkpointing)
+
+
+def test_checkpoint_type_has_attn_only():
+    assert CheckpointType.ATTN_ONLY == "attn_only"
+    assert CheckpointType.ATTN_ONLY.value == "attn_only"
+
+
+def test_is_attention_forward_matches_flash_and_sdpa_forward():
+    # FA2 names the op "...forward", FA3 (flash_attn_interface) names it "...fwd",
+    # aten exposes "_scaled_dot_product_*". The matcher keys on the op name so it
+    # is robust to whichever backend is registered at runtime.
+    assert _is_attention_forward("flash_attn::_flash_attn_forward")
+    assert _is_attention_forward("flash_attn_3::fwd")
+    assert _is_attention_forward(torch.ops.aten._scaled_dot_product_flash_attention.default)
+    assert _is_attention_forward(torch.ops.aten._scaled_dot_product_efficient_attention.default)
+
+
+def test_is_attention_forward_excludes_backward_and_unrelated_ops():
+    # Backward ops must NOT be saved (we only want the forward output).
+    assert not _is_attention_forward("flash_attn::_flash_attn_backward")
+    assert not _is_attention_forward("flash_attn_3::bwd")
+    assert not _is_attention_forward("aten::_scaled_dot_product_flash_attention_backward")
+    # Unrelated compute (the FFN/QKV GEMM we deliberately recompute).
+    assert not _is_attention_forward(torch.ops.aten.mm.default)
+    assert not _is_attention_forward("aten::addmm")
+
+
+class _DummyBlock(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.proj = nn.Linear(4, 4)
+
+    def forward(self, x):  # pragma: no cover - never executed in these tests
+        return self.proj(x)
+
+
+class _DummyTransformer(nn.Module):
+    def __init__(self, n_blocks: int = 3):
+        super().__init__()
+        self.blocks = nn.ModuleList([_DummyBlock() for _ in range(n_blocks)])
+
+
+def test_attn_only_wraps_each_transformer_block():
+    model = _DummyTransformer(n_blocks=3)
+    returned = apply_activation_checkpointing(model, checkpointing_type=CheckpointType.ATTN_ONLY)
+
+    # The module is wrapped in place and returned.
+    assert returned is model
+    assert len(model.blocks) == 3
+    for block in model.blocks:
+        assert isinstance(block, CheckpointWrapper)
+        # The original block is preserved under the wrapper.
+        assert isinstance(block._checkpoint_wrapped_module, _DummyBlock)
+
+
+def test_attn_only_raises_when_no_transformer_blocks_found():
+    class _NoBlocks(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.linear = nn.Linear(4, 4)
+
+    try:
+        apply_activation_checkpointing(_NoBlocks(), checkpointing_type=CheckpointType.ATTN_ONLY)
+    except ValueError as exc:
+        assert "attn_only" in str(exc)
+    else:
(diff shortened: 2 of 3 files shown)


SolitaryThinker:

## TL;DR

The CLI default remains backwards-compatible, and the new `attn_only` path is plausibly using selective activation checkpointing to save attention and collective outputs while recomputing other block ops. The blockers are coverage and backend support: the tests never run a forward/backward parity check, and the op classifier recognizes Flash/SDPA names but not several shipped FastVideo attention backends.

- S0 (blockers): 0
- S1 (must-fix): 2

---

### [S1] Add a real forward/backward parity test for `attn_only`

**What:** The new test file is explicitly CPU-only and says it exercises the op classifier and wrapper structure without running a forward/backward. The dummy block forward is marked never executed, so no test compares outputs, losses, input gradients, or parameter gradients for `none`, `full`, and `attn_only`.

**Why it matters:** Selective activation checkpointing can look structurally correct while still breaking autograd, RNG behavior, custom-op caching, or saved-tensor policy semantics. This PR changes training behavior, so the minimum safety bar is numerical and gradient parity against the uncheckpointed path.

**Suggested fix:** Add a small deterministic module with one real attention path plus an MLP, run one forward/backward under no checkpointing, `full`, and `attn_only`, then assert allclose for outputs, loss, input grads, and parameter grads. Keep a CPU/SDPA parity test always runnable, and add a CUDA backend case for at least one FastVideo-shipped attention implementation when available.

**Evidence:** `fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py:1`, `fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py:72`, `fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py:82`

---

### [S1] Gate or cover non-Flash FastVideo attention backends

**What:** `_is_attention_forward` only returns true for op names containing `flash_attn` plus `forward`/`fwd`, or `_scaled_dot_product`. FastVideo also ships attention backends whose calls do not match that predicate, including `sageattn`, `video_sparse_attn`, `moba_attn_varlen`, and SLA's custom autograd attention.

**Why it matters:** For those backends, `attn_only` silently treats the attention kernel as `PREFER_RECOMPUTE`, so the documented guarantee that attention forward outputs are saved is false and custom-kernel re-entry during backward is unverified. That undermines both the performance claim and the training-safety story for any training recipe using those backends.

**Suggested fix:** Either expand the policy to recognize every training-supported attention backend and add backend-specific coverage, or fail fast when `attn_only` is requested with an unsupported backend. If the intended support matrix is only Flash/SDPA, document that in the CLI help and tests instead of claiming broad FastVideo attention coverage.

**Evidence:** `fastvideo/training/activation_checkpoint.py:101`, `fastvideo/attention/backends/sage_attn.py:57`, `fastvideo/attention/backends/video_sparse_attn.py:292`, `fastvideo/attention/backends/vmoba.py:187`, `fastvideo/attention/backends/sla.py:308`

---

### [S2] Add distributed coverage for the collective-saving policy

**What:** The implementation explicitly saves `_c10d_functional` outputs to avoid recomputing FSDP2 collectives, but the new tests are CPU-only and avoid distributed initialization entirely.

**Why it matters:** FSDP/SP plus activation checkpointing is where collective ordering bugs and backward deadlocks show up. A string classifier test cannot validate that the checkpoint wrapper, selective policy, and distributed collectives interact safely in a real train step.

**Suggested fix:** Add a minimal two-rank distributed regression, even if it is nightly/GPU-gated, that runs one forward/backward with FSDP or the sequence-parallel path under `attn_only` and compares with the existing checkpoint mode. At minimum, assert no extra collective is issued during recompute for the supported FSDP path.

**Evidence:** `fastvideo/training/activation_checkpoint.py:109`, `fastvideo/tests/training/checkpoint/test_activation_checkpoint_attn_only.py:1`

---

alexzms (changes requested):
Wonderful PR, thanks for the contribution. One thing: under VSA the attention forward op is fastvideo_kernel::block_sparse_attn_triton/_sm90, which _is_attention_forward doesn't match (it only keys on flash_attn/_scaled_dot_product), so VSA attention still gets recomputed and attn_only loses its benefit there. Let me know if you need any help.

rich7420:
@alexzms thanks for the comment!

rich7420:
Pushed a fix for the VSA gap @alexzms / @SolitaryThinker flagged 
`_is_attention_forward` now also matches the `block_sparse_attn` op so VSA attention actually gets saved under `attn_only`, plus added a real fwd/bwd parity test; verified end-to-end on H100×2 VSA training (`block_sparse_attn_sm90` → SAVE, no OOM).

PR #1394, opened 2026-05-26
[bugfix] Fix Dreamverse Modal compile warmup latency

Davids048:
## Summary
- Pre-stabilize the LTX2 Gemma text encoder before torch.compile by running a one-token eager forward with `output_hidden_states=True`, avoiding a second text-encoder compile caused by Transformers' hidden-state wrapper side effect.
- Enable VAE torch.compile in the Dreamverse LTX2 generator config so Modal/Docker deployments match the local fast-profile path and avoid slow eager VAE decode.
- Deploy Modal with the selected `DREAMVERSE_IMAGE`, startup warmup, and torch compile enabled so registry image deploys exercise the optimized path by default.

## Validation
- `python -m py_compile apps/dreamverse/dreamverse/video_generation.py apps/dreamverse/scripts/modal/modal_app.py fastvideo/models/encoders/gemma.py`
- pre-commit hooks on commit: yapf, ruff, codespell, mypy, filename checks
- Built and pushed test image: `ghcr.io/davids048/dreamverse-ui:cuda12.9.1-sha-321d5112b-warmup-vaecompile`
- Modal test deploy reached ready with 1 GPU worker and one active runtime container.
- One-client websocket streaming smoke passed; post-warmup segments were ~3.7-4.3s generation time instead of ~5.4s.

PR #1413, opened 2026-05-29
[bugfix]: dreamverse modal bypasses ENTRYPOINT — set ffmpeg env + key check

alexzms:
## Summary

Modal's `@modal.web_server` decorator invokes `serve()` directly (`Popen` the `dreamverse-server`) and bypasses the image's ENTRYPOINT (`apps/dreamverse/docker/docker_entrypoint.sh`). The bypassed entrypoint normally does two things that **silently no-op on Modal**:

1. `:?`-validates `CEREBRAS_API_KEY` and `GROQ_API_KEY` (fails the container fast if either is missing from the mounted secret).
2. Sources `ffmpeg-env.sh`, which exports `FASTVIDEO_FFMPEG_BIN=/opt/ffmpeg-native/bin/ffmpeg`.

## Symptoms observed

Deploying the post-#1394 UI image (`dreamverse-ui-cuda12.9.1-sha-2c13793`) on Modal:

- **No ffmpeg, video silently broken.** `/opt/ffmpeg-native/bin` is not on the container `PATH` (`apps/dreamverse/docker/Dockerfile:24`). With `FASTVIDEO_FFMPEG_BIN` unset, `av_streaming.py:28`'s `shutil.which("ffmpeg")` returns `None` and `av_fmp4` muxing has no encoder.
- **Missing keys surface at first request, not boot.** A `dreamverse-api-keys` secret with placeholder values lets the container come up fine; prompt rewriting then fails on the first user request instead of at boot.

## Fix

Both changes in `apps/dreamverse/scripts/modal/modal_app.py`:

1. Set `FASTVIDEO_FFMPEG_BIN=/opt/ffmpeg-native/bin/ffmpeg` and `FASTVIDEO_VIDEO_CODEC=libx264` in `image.env` so `av_streaming` finds the native ffmpeg the Dockerfile builds.
2. Add a fail-fast guard in `serve()` that checks the required secret entries before launching the server, mirroring the entrypoint's `:?` behavior.

A comment block above the env dict documents why these are needed (entrypoint bypass), so the next reader doesn't lose the context.

## Test evidence

- `pre-commit run --files apps/dreamverse/scripts/modal/modal_app.py` and `python -c "import ast; ast.parse(open(...))"` pass.
- The ffmpeg gap was empirically discovered while bringing up the dreamverse Modal deploy with the post-#1394 image; setting `FASTVIDEO_FFMPEG_BIN` in `image.env` was the workaround that made `av_streaming`'s `shutil.which` resolve to the right binary during that session.
- The key fail-fast is mechanical (re-raises `RuntimeError` when `os.environ.get(k)` is empty for either required key); the message names the missing key(s) and points at the README so the operator knows exactly what to fix.

### DIFF
--- a/apps/dreamverse/scripts/modal/modal_app.py
+++ b/apps/dreamverse/scripts/modal/modal_app.py
@@ -61,7 +61,13 @@ dreamverse_state = modal.Volume.from_name("dreamverse-state", create_if_missing=
 )
 @modal.web_server(8009, startup_timeout=4800)
 def serve():
-    missing = [k for k in _REQUIRED_SECRET_KEYS if not os.environ.get(k)]
+    # ``or ""`` collapses ``None`` (unset) into an empty string, ``.strip()``
+    # collapses whitespace-only values (e.g. ``"   "``) — both should be
+    # treated as missing.
+    missing = [
+        k for k in _REQUIRED_SECRET_KEYS
+        if not (os.environ.get(k) or "").strip()
+    ]
     if missing:
         raise RuntimeError(
             "dreamverse-api-keys secret is missing required entries: "
--- a/apps/dreamverse/scripts/modal/modal_app.py
+++ b/apps/dreamverse/scripts/modal/modal_app.py
@@ -14,6 +14,12 @@ if not IMAGE:
         "dreamverse-ui-cuda12.9.1-sha-* tag if serving the static UI."
     )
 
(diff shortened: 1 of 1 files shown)


alexzms:
Pushed `e1bde40a` to address gemini's whitespace nit: `not (os.environ.get(k) or "").strip()` now collapses unset (`None`), empty (`""`), and whitespace-only (`"   "`) values into the "missing" set.

SolitaryThinker (approved):

PR #1414, opened 2026-05-30
[bugfix] matrixgame2 causal denoise: store KV-cache end indices as Python int

rich7420:
## Summary

matrixgame2 causal inference stored the KV-cache end indices
(`global_end_index` / `local_end_index`) as GPU tensors. The consumers in
`causal_model.py` and `action_module.py` then call `int(tensor.item())` on
every access — each `.item()` is a GPU→CPU sync that serializes the
autoregressive denoise loop.

Both consumers already handle a plain Python `int` (their
`isinstance(..., torch.Tensor)` check falls through to the int path), so this
just initializes the indices as ints in `_initialize_kv_cache` /
`_initialize_action_kv_cache`.

## How it was found (nsys-ai)

On an H100 matrixgame2 trace, `host_sync_parent_ranges` attributed
**10,244 `aten::item` syncs** to the denoising stage (~264 ms), and
`cpu_gpu_pipeline` showed ~29.7k GPU starvation events from host-blocking
between dispatches.

## Result

Denoise-stage `.item()` syncs: **10,244 → 644 (−94%)**. Output unchanged — the
consumer sites already accepted plain ints.

### DIFF
--- a/fastvideo/pipelines/stages/matrixgame2_denoising.py
+++ b/fastvideo/pipelines/stages/matrixgame2_denoising.py
@@ -278,10 +278,8 @@ class MatrixGame2CausalDenoisingStage(DenoisingStage):
                 torch.zeros([batch_size, kv_cache_size, num_attention_heads, attention_head_dim],
                             dtype=dtype,
                             device=device),
-                "global_end_index":
-                torch.tensor([0], dtype=torch.long, device=device),
-                "local_end_index":
-                torch.tensor([0], dtype=torch.long, device=device),
+                "global_end_index": 0,
+                "local_end_index": 0,
             })
 
         return kv_cache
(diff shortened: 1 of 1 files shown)


SolitaryThinker:

## TL;DR

The int conversion is directionally correct: the MatrixGame2 causal and action cache consumers accept Python `int` values and then keep the cache indices as ints, avoiding the old CUDA `.item()` path. The one gap is regression coverage: existing MatrixGame2 tests do not assert this cache-index type/sync invariant, and one nearby transformer test is skipped.

## Verdict: **Changes requested: add regression coverage**

- S0 (blockers): 0
- S1 (must-fix): 0

---

### [S2] Missing regression coverage for Python-int KV-cache indices

**What:** The PR changes MatrixGame2 denoising cache initialization from CUDA tensor counters to Python `int` counters, but it does not add a test that would fail if these keys regress back to tensors.

**Why it matters:** The old tensor counters route through `.item()` in the MatrixGame2 causal/action cache consumers, which forces host synchronization on CUDA and can silently reappear without a targeted type assertion. The existing MatrixGame2 SSIM test is an expensive end-to-end quality check, while `fastvideo/tests/transformers/test_matrixgame2.py` is skipped and does not exercise this denoising-stage cache initialization.

**Suggested fix:** Add a lightweight stage-level regression test that initializes `_initialize_kv_cache()` and `_initialize_action_kv_cache()` and asserts `global_end_index` and `local_end_index` are plain `int` values for the normal, mouse, and keyboard caches. If practical, also run one minimal forward/update path with fake cache tensors and assert the consumers preserve `int` indices after updating.

**Evidence:** `fastvideo/pipelines/stages/matrixgame2_denoising.py:281`

---

PR #1415, opened 2026-05-30
[perf] matrixgame2 causal: cache per-forward RoPE/timestep/blockmask recompute on-device

rich7420:
## Summary

matrixgame2 causal inference repeated constant work on **every** DiT forward.
This removes it:

- Four flex-attention BlockMasks were rebuilt via `create_block_mask` each
  forward (compile + a GPU-tensor python loop). They depend only on
  `(num_frames, frame_seqlen, block size, local_attn_size, device)`, so they
  are cached per param-tuple (device included in the key).
- `get_rotary_pos_embed` ran a float64 CPU compute + H2D copy each forward to
  build `freqs_cis`, which `CausalMatrixGame2SelfAttention.forward` never reads
  (it computes its own RoPE via `self._freqs_cache`). Now skipped entirely
  (`freqs_cis = None`).
- the sinusoidal timestep table is built on the target device instead of
  CPU + H2D.
- the action-module RoPE freqs are cached device-resident (with a device
  check) so the per-call `.to` is a no-op.

All numerically identical.

## How it was found (nsys-ai)

On an H100 matrixgame2 trace, `gpu_idle_gaps` plus the memcpy breakdown showed
heavy per-forward host→device traffic and repeated `create_block_mask` work; a
CUDA-graph capture probe pinpointed the exact per-forward recompute sites.

## Result (H100, 117 frames = 30 DiT forwards)

- isolated DiT forward: **194 ms → 95 ms (~2×)**
- end-to-end inference wall: **12.6 s → 10.6 s (−16%)**
- denoise-stage GPU-active share: 23% → 33.5%

The biggest single contributor is the BlockMask cache — `create_block_mask` is
expensive and was running once per forward.


SolitaryThinker:

## TL;DR

This PR removes repeated matrixgame2 causal inference work by moving timestep frequency creation onto-device, skipping unused RoPE construction, and caching device-specific block masks. I found one must-fix cache invalidation issue: the block-mask cache is built from the model default block size even when the caller overrides the current block size, which can silently apply the wrong attention mask for boundary or variable-sized blocks.

## Verdict: **request-changes**

- S0 (blockers): 0
- S1 (must-fix): 1
- S2 (should-fix; surfaced if `persistent` or `important`): 0

---

### [S1] Block-mask cache ignores the per-call block size override

**What:** In `CausalMatrixGame2WanModel._forward_inference`, the cached BlockMask key and builder use `_nfb = self.num_frame_per_block` at `fastvideo/models/dits/matrixgame2/causal_model.py:1032`, but the same forward computes `effective_num_frame_per_block` from `kwargs` at `causal_model.py:902` and passes that effective value into each block at `causal_model.py:1104`. The MatrixGame2 denoising stage explicitly passes the current block size as `num_frame_per_block = num_frames` at `fastvideo/pipelines/stages/matrixgame2_denoising.py:257-258`, and boundary mode can set the first block to a different size.

**Why it matters:** The flex-attention masks depend on the actual block size used for the current forward. If the first boundary block or any future variable-sized block runs with `effective_num_frame_per_block != self.num_frame_per_block`, the cache will reuse/build a mask for the wrong attention pattern while the action and KV-cache logic use the effective block size, producing silent wrong attention rather than a crash.

**Suggested fix:** Use `effective_num_frame_per_block` for `_nfb` when constructing all three inference block-mask cache keys and when calling the `_prepare_blockwise_*` helpers. Add a small parity/unit probe that calls `_forward_inference` twice with the same `num_frames` but different `num_frame_per_block` overrides and asserts separate cache entries or masks.

**Evidence:** `fastvideo/models/dits/matrixgame2/causal_model.py:902`, `fastvideo/models/dits/matrixgame2/causal_model.py:1032`, `fastvideo/models/dits/matrixgame2/causal_model.py:1035-1062`, `fastvideo/models/dits/matrixgame2/causal_model.py:1104`, `fastvideo/pipelines/stages/matrixgame2_denoising.py:257-258`

---

SolitaryThinker:
Reviewed at 270de10. Substance looks right: the block-mask cache now keys on `effective_num_frame_per_block` (matches what blocks actually run with; identical output in the shipped stage flow), `freqs_cis` is confirmed dead in causal attention even after #1454, and the caches are plain attrs FSDP won't touch.

Two asks before merge:
1. pre-commit is red and it's this PR — yapf wants the new `freqs` line in `fastvideo/layers/visual_embedding.py` joined to a single line. Please run `pre-commit run --all-files` and push.
2. That same hunk moves `exp` onto GPU, so the freq table is ~1-ulp different, not bit-identical — and `TimestepEmbedder` is shared by every DiT in both the inference and training stacks. Please soften the "numerically identical" claim in the body to "within 1 ulp; feeds a bf16 MLP".

A small cache unit test like #1442's (same args → same object; different `num_frame_per_block` → new entry) would be a nice bonus but isn't blocking.

PR #1372, opened 2026-05-19
[perf] Add Adaptive Guidance (CFG gating) for stale-uncond reuse

rich7420:
## Purpose

Adds an inference-only CFG acceleration that skips the unconditional forward pass after a
fraction of the denoising schedule has completed and reconstructs the guidance prediction
from a cached `delta = noise_pred_cond − noise_pred_uncond`:

```
noise_pred = noise_pred_cond + (guidance_scale − 1) * delta_cached
```

This is the LinearAG variant of **Adaptive Guidance** ([Castillo et al. 2023, arXiv:2312.12487](https://arxiv.org/abs/2312.12487),
AAAI 2025) — described in the paper as "simple affine transformations of past score
estimates" replacing entire neural function evaluations in the post-convergence phase.

The same "split denoising at a gate step and optimize the post-gate phase" framing is
also explored in [TGATE](https://arxiv.org/abs/2404.02747) (Liu et al. 2024), but TGATE
caches attention outputs and keeps both CFG forwards — a different mechanism. This PR
implements the Adaptive Guidance mechanism (skip the uncond forward); the env var name
`FASTVIDEO_CFG_GATE_STEP` describes the action rather than naming either upstream method.

Gated by a single env var `FASTVIDEO_CFG_GATE_STEP` (float `[0, 1]`, default `1.0` =
no-op). Existing users see zero behavioural change unless they opt in.

Measured **−22.0 % end-to-end on 4× L40S** and **−23.7 % on 1× H100** for Wan T2V 1.3B at
720×1280 / 77 frames / 30 inference steps, with VBench quality preserved within noise on
4 of 5 dimensions. On a separate (out-of-scope-for-this-PR) FA3 attention path on H100,
the win composes multiplicatively with the attention speedup for **−50.8 % end-to-end**.

## Changes

- **`fastvideo/envs.py`** (+26 lines): introduce `FASTVIDEO_CFG_GATE_STEP: float`
  declaration in the `TYPE_CHECKING` block plus the dispatching lambda in
  `environment_variables`, with full inline docstring covering semantics, edge
  cases (0.0 / 0.5 / 1.0), and known caveats.
- **`fastvideo/pipelines/stages/denoising.py`** (+127 / −18 lines, behaviour-additive):
  - Setup block before the denoising loop: validate the env value (raises on
    out-of-range), pre-compute `gate_step = int(num_steps * fraction)`, and
    initialise `delta_cached` + `delta_cached_model_id` + telemetry counters.
[...]

PR #1416, opened 2026-05-30
[perf]: batched CFG for Wan/Cosmos + diffusers T5 recipe alignment

Mister-Raggs:
## Summary

Adds an opt-in batched classifier-free-guidance path in the main `DenoisingStage`: when CFG is on, the cond and uncond passes run as a single `batch=2` DiT forward per denoise step instead of two sequential `batch=1` forwards. Bit-equivalent to sequential on H100 FA3 eager (SSIM = 1.000000).

Adds the prerequisite: align Wan / Cosmos text encoding to diffusers' canonical recipe (`tokenize(padding="max_length")` → trim T5 output to real length → re-pad with explicit zeros to max_length). HunyuanVideo and all other DiTs are unaffected — diffusers uses a different recipe for those, so they're gated off.

Default flag value is `True`; sequential-CFG fallback preserved for V2V/I2V/TI2V/action/camera paths (auto-detected).

## Problem

Two issues, addressed together because they're coupled:

1. **Sequential CFG wastes launch overhead and prevents kernel-level batching.** Every denoise step runs two `current_model(...)` calls — once for the conditional prompt, once for the unconditional. On modern GPUs, batch=2 forward is typically 1.4-1.6× the wall of batch=1 (vs 2× sequential), so the mechanism leaves perf on the table.

2. **FV's Wan/Cosmos T5 output deviates from their canonical diffusers pipeline.** Diffusers' [`WanPipeline._get_t5_prompt_embeds`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/wan/pipeline_wan.py#L173-L190) and [`CosmosTextToWorldPipeline._get_t5_prompt_embeds`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/cosmos/pipeline_cosmos_text2world.py#L197-L237) both tokenize with \`padding=\"max_length\"\`, then trim the T5 output to each sample's real length, then re-pad with explicit zeros to \`max_length\`. That's the recipe these models were trained against; FV's variable-length path is a deviation. Aligning makes the model's input match training distribution, AND gives the matched-shape contract that batched-CFG needs to cat pos + neg along the batch dim.

## Solution

**\`FastVideoArgs\`:** new \`use_batched_cfg: bool = True\` field with \`--use-batched-cfg\` CLI flag. Added to \`_FROM_PRETRAINED_CONVENIENCE_KWARGS\`.

**\`TextEncodingStage\`:** when CFG is on AND the active \`dit_config\` MRO contains \`WanVideoConfig\`, \`CosmosVideoConfig\`, or \`Cosmos25VideoConfig\`, applies the diffusers-canonical trim+pad-with-zeros recipe to both pos and neg encodings. Other DiTs (HunyuanVideo, etc.) bypass the gate entirely — their behaviour is unchanged.

**\`DenoisingStage.forward\` (main one):** new gated branch when \`use_batched_cfg=True\` and CFG is on. Cats \`[neg, pos]\` along the batch dim, single forward at batch=2, \`chunk(2)\` → existing CFG combine + \`guidance_rescale\` math reused unchanged. Defensive shape-mismatch fallback to sequential if the batch was constructed without going through \`TextEncodingStage\`.

Autodetect-off when V2V/I2V/TI2V/action/camera conditioning is present in the batch — those carry batch=1 conditioning tensors that aren't covered by this PR.

## Validation

5 prompts, seed-pinned, single-container two-pass A/B (sequential pass + batched pass through the same \`VideoGenerator\`). Pairwise SSIM via \`pytorch_msssim.ssim\` (matches \`fastvideo/tests/utils.py:compute_video_ssim_torchvision\`).

| Config | Wall delta | SSIM mean | SSIM worst | Note |
|---|---|---|---|---|
| **H100 14B eager** (FA3, SP=2) | **-1.96%** | **1.000000** | **1.000000** | bit-equivalent — clean mechanism proof |
| H100 14B compile (FA3, SP=2) | -1.78% | 0.85 | 0.68 | compile-induced kernel drift; videos visually identical (verified by frame inspection) |
| L40S 1.3B eager (FA2) | +2.0% | 0.96 | 0.95 | small model + FA2 batched-numerics noise + recipe-alignment cost > batched savings on this scale |

### Visual evidence — H100 14B compile, prompt 4 (worst-SSIM prompt)

Frame-by-frame inspection confirms the SSIM 0.68-worst delta in compile mode is bf16 / Inductor kernel-drift noise, not semantic divergence.

**Baseline (sequential CFG):**

https://github.com/user-attachments/assets/6b3fba94-d669-4f87-bb54-5881d27c96ea

**Patched (batched CFG):**

https://github.com/user-attachments/assets/6188fa62-5edf-4e7b-aedb-ef8c65ee2567

### Visual evidence — H100 14B eager, prompt 4 (bit-equivalent reference)

For comparison: SSIM = 1.000000 exactly, no kernel-drift contribution. Demonstrates that the batched-CFG mechanism itself is bit-equivalent to sequential.

**Baseline (sequential CFG):**

https://github.com/user-attachments/assets/9c416096-4864-495b-aecb-f2a5aca1a386

**Patched (batched CFG):**

https://github.com/user-attachments/assets/fbb22c26-9f0e-40cb-9c4d-20c87f6263ec

## Trade-offs

**Recipe alignment changes Wan/Cosmos output character** vs FV's previous variable-length default. The new output matches diffusers' canonical pipeline (= Wan's training distribution). Existing Wan/Cosmos SSIM regression test reference videos in CI may need regeneration if the alignment shifts pixel values beyond the SSIM gate. HunyuanVideo and other DiTs are unaffected.

**Compile path shows SSIM 0.85 mean / 0.68 worst** on Wan 14B H100. The cause is Inductor picking different fused kernels for batch=1 vs batch=2 paths (not a bug in the batched-CFG mechanism itself, which is bit-equivalent on the eager path). Frame-by-frame visual inspection of the worst-SSIM prompt confirmed the videos are perceptually identical. This is consistent with bf16 numerics drift compounding across 30 layers × 30 steps, well within the perceptual-equivalence range.

**L40S 1.3B shows +2% wall regression.** The diffusers-canonical recipe pads every cross-attention K/V to \`max_length=512\` (vs FV's previous variable-length 55-98 tokens). On a small model, this per-step attention overhead exceeds the batched-CFG launch savings. The win shows up at larger models / faster GPUs / more amortizable launch overhead — which is exactly what we see on H100 14B (-1.96% eager, -1.78% compile).

## Test plan

- [x] **Unit test:** \`fastvideo/tests/stages/test_denoising_batched_cfg.py\` — stub-transformer equivalence, CFG-off equivalence, autodetect-off on 7 conditioning fields.
- [x] **Math unit test:** \`fastvideo/tests/stages/test_wan_cross_attn_mask_math.py\` — kept in tree as forward-coverage for a model-side-mask approach a future contributor might attempt for HunyuanVideo. Passes 10/11 cases (fp32 exact + bf16 within 1 ULP).
- [x] **Modal A/B harness:** \`fastvideo/tests/modal/batched_cfg_ab.py\` + \`_batched_cfg_ab_inner.py\`. Single-container two-pass A/B; subprocesses the inner script via \`/opt/venv/bin/python\` so FastVideo runs in the image's venv. Reusable for future single-flag perf PRs.
- [ ] **CI SSIM regression suite:** Wan/Cosmos may need reference-video regeneration; HunyuanVideo expected unchanged. Will iterate on review.

## Reproduce these numbers yourself

All A/B runs above are reproducible end-to-end via the included Modal harness. ~\$10 total compute for the full matrix.

**One-time setup:**
\`\`\`bash
pip install modal && modal token new
modal secret create huggingface-token HF_TOKEN=hf_xxxxxxxxxxxx
\`\`\`

**Run the same A/B legs as the validation table:**
\`\`\`bash
# H100 14B eager (the SSIM=1.000000 / -1.96% wall result)
modal run --detach fastvideo/tests/modal/batched_cfg_ab.py \\
  --gpu H100 --model wan2_2-t2v-14b

# H100 14B compile (deployment perf number)
modal run --detach fastvideo/tests/modal/batched_cfg_ab.py \\
  --gpu H100 --model wan2_2-t2v-14b --enable-compile

# L40S 1.3B eager (correctness validation)
modal run --detach fastvideo/tests/modal/batched_cfg_ab.py \\
  --gpu L40S --model wan2_1-t2v-1.3b
\`\`\`

Each leg writes mp4s to a per-config directory in the \`hf-model-weights\` Volume (\`/root/data/ab_out/<model>/<mode_tag>/{baseline,patched}/prompt_NN/\`). The harness prints the wall + SSIM table at the end of each run. If your terminal disconnects, recover via:

\`\`\`bash
modal run --detach fastvideo/tests/modal/batched_cfg_ab.py::recover_ssim \\
  --git-repo https://github.com/Mister-Raggs/FastVideo.git \\
  --git-ref perf/wan-batched-cfg \\
  --model-preset wan2_2-t2v-14b --mode-tag eager
\`\`\`

**Use the harness for your own PRs.** The single-container two-pass A/B pattern is generic (modeled on Kuan's \`nccl_stream_ab.py\` for #1395) — fork the harness, swap the \`MODEL_PRESETS\` entry and the flag toggle, and you have a reproducible perf+SSIM gate for any single-flag PR.

## Follow-ups (not blocking this PR)

- **HunyuanVideo batched-CFG:** different recipe required (encoder output used as-is per diffusers). The model-side SDPA + bool attn_mask path explored and reverted in this PR is actually the right tool for HV; the local unit test left in tree provides a head start.
- **Padding-length optimization:** pad to \`max(pos_real, neg_real)\` instead of fixed \`max_length=512\`. Could recover most of the per-cross-attn cost the recipe alignment introduces; trade-off is deviation from training distribution. Worth measuring SSIM at variable padding.
- **Multi-prompt batching:** the batched=2 framework here extends naturally to batched=N for serving workloads.
- **Recipe-alignment quality measurement:** worth comparing (a) FV current variable-length Wan vs (b) this PR's recipe-aligned Wan vs (c) diffusers reference. If alignment improves output character independently of batched-CFG, that's a separately-citable quality win.

## Related work

Composes with [#1372 Adaptive Guidance](https://github.com/hao-ai-lab/FastVideo/pull/1372): when AG runs both passes, batched=2 fires; when AG skips uncond, batched-CFG gracefully falls through to sequential.

Mirrors the upstream pattern in \`fastvideo/pipelines/stages/longcat_denoising.py:97-149\` which already does batched-CFG natively (LongCat's tokenizer is configured with \`padding=\"max_length\"\` upstream so the same shape contract holds).

### DIFF
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -39,6 +39,7 @@ surfaces:
       torch_compile_kwargs_audio_vae: generator.engine.compile.audio_vae_kwargs
       transformer_quant: generator.engine.quantization.transformer_quant
       disable_autocast: generator.engine.disable_autocast
+      use_batched_cfg: generator.engine.use_batched_cfg
       enable_stage_verification: generator.engine.enable_stage_verification
       prompt_txt: request.inputs.prompt_path
       override_text_encoder_safetensors: generator.pipeline.components.text_encoder_weights
--- a/fastvideo/api/compat.py
+++ b/fastvideo/api/compat.py
@@ -157,7 +157,7 @@ def legacy_from_pretrained_to_config(
             preset_refine["num_inference_steps"] = value
         elif key == "ltx2_refine_guidance_scale":
             preset_refine["guidance_scale"] = value
-        elif key in {"enable_stage_verification", "use_fsdp_inference", "disable_autocast"}:
+        elif key in {"enable_stage_verification", "use_fsdp_inference", "disable_autocast", "use_batched_cfg"}:
             engine[key] = value
         elif key == "override_text_encoder_quant":
             quantization["text_encoder_quant"] = value
@@ -244,6 +244,7 @@ def generator_config_to_fastvideo_args(config: GeneratorConfig | Mapping[str, An
         "enable_stage_verification": engine.enable_stage_verification,
         "use_fsdp_inference": engine.use_fsdp_inference,
         "disable_autocast": engine.disable_autocast,
+        "use_batched_cfg": engine.use_batched_cfg,
     }
     if normalized.pipeline.workload_type is not None:
         kwargs["workload_type"] = normalized.pipeline.workload_type
--- a/fastvideo/api/schema.py
+++ b/fastvideo/api/schema.py
@@ -81,6 +81,7 @@ class EngineConfig:
     enable_stage_verification: bool = True
     use_fsdp_inference: bool = False
     disable_autocast: bool = False
+    use_batched_cfg: bool = False
     quantization: QuantizationConfig | None = None
 
 
--- a/fastvideo/entrypoints/video_generator.py
+++ b/fastvideo/entrypoints/video_generator.py
@@ -94,6 +94,7 @@ _FROM_PRETRAINED_CONVENIENCE_KWARGS = frozenset({
     "pin_cpu_memory",
     "enable_torch_compile",
     "torch_compile_kwargs",
+    "use_batched_cfg",
     "output_type",
     "nvfp4_fa4",
 })
--- a/fastvideo/fastvideo_args.py
+++ b/fastvideo/fastvideo_args.py
@@ -153,6 +153,17 @@ class FastVideoArgs:
 
     disable_autocast: bool = False
 
+    # Opt-in batched classifier-free guidance: run cond + uncond as a
+    # single batch=2 DiT forward per denoise step instead of two
+    # sequential batch=1 forwards. Bit-equivalent to sequential CFG on
+    # H100 FA3 eager (SSIM=1.000000). On other configs (compile,
+    # FA2, smaller models) numerics drift slightly (~0.04 SSIM mean,
+    # visually imperceptible) due to Inductor kernel selection and
+    # batched flash-attn numerics; perf delta is run-to-run variable.
+    # Default OFF so this PR has no behaviour change for any user who
+    # doesn't opt in; recipe alignment + batched path fire together.
+    use_batched_cfg: bool = False
+
     # VSA parameters
     VSA_sparsity: float = 0.0  # inference/validation sparsity
 
@@ -543,6 +554,17 @@ class FastVideoArgs:
             default=FastVideoArgs.enable_torch_compile,
             help="Use torch.compile to speed up DiT inference." +
             "However, will likely cause precision drifts. See (https://github.com/pytorch/pytorch/issues/145213)",
+        )
+        parser.add_argument(
+            "--use-batched-cfg",
+            action=StoreBoolean,
+            default=FastVideoArgs.use_batched_cfg,
+            help="Run classifier-free guidance as a single batch=2 DiT "
+            "forward per step (cond+uncond stacked) instead of two "
+            "sequential batch=1 forwards. Output-identical at SSIM=1.0; "
+            "reduces per-step launch + memory overhead. Falls back to "
+            "the sequential path when V2V/I2V/TI2V/action/camera "
+            "conditioning is present.",
         )
         parser.add_argument(
             "--torch-compile-kwargs",
--- a/fastvideo/pipelines/stages/denoising.py
+++ b/fastvideo/pipelines/stages/denoising.py
@@ -331,22 +331,11 @@ class DenoisingStage(PipelineStage):
         _cfg_gate_active = _cfg_gate_fraction < 1.0 and batch.do_classifier_free_guidance
         _is_rank0 = get_world_group().local_rank == 0
         if _cfg_gate_active:
-            # Use len(timesteps), not num_inference_steps: the loop iterates
-            # over timesteps directly, and for schedulers with order > 1
-            # (e.g. DPM-Solver++ 2M, Heun) len(timesteps) is a multiple of
-            # num_inference_steps. Using num_inference_steps would cause the
-            # gate to fire at fraction/order of the loop instead of fraction.
             _cfg_gate_step_idx = int(len(timesteps) * _cfg_gate_fraction)
             if _is_rank0:
                 logger.info("CFG gating enabled: fraction=%.3f, gate_step=%d/%d", _cfg_gate_fraction,
                             _cfg_gate_step_idx, len(timesteps))
             if batch.guidance_rescale > 0.0 and _is_rank0:
-                # guidance_rescale rescales CFG output stats to match cond
-                # stats (Lin et al. §3.4).  When `delta_cached` goes stale,
-                # the rescaling still computes but is no longer guaranteed
-                # to preserve the original quality semantics.  Warn so the
-                # caller knows this combo is unvalidated; tighten or
-                # fallback once VBench data lands.
                 logger.warning(
                     "CFG gating (fraction=%.3f) combined with guidance_rescale=%.3f is unvalidated; "
                     "quality may degrade beyond CFG-gating-alone expectations.", _cfg_gate_fraction,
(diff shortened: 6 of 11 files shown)


Mister-Raggs:
**Rebased onto current main (post-#1372 AG merge) — composition note**

Two changes in the force-push:

1. **`docs/design/inference_schema_parity_inventory.yaml`**: added `use_batched_cfg: generator.engine.use_batched_cfg` under `surfaces.fastvideo_args.moved`, mirroring `disable_autocast`. Fixes the `test_fastvideo_args_fields_are_classified` parity test that flagged the new field.

2. **`denoising.py` merge with #1372 Adaptive Guidance**: AG and batched-CFG are now mutually exclusive at the gate:
   ```python
   use_batched_cfg = (fastvideo_args.use_batched_cfg
                      and batch.do_classifier_free_guidance
                      and not _cfg_conditioning_present
                      and not _cfg_gate_active)
   ```
   **Rationale:** AG selectively skips the uncond forward to save wall (its design); batched-CFG forces both passes in one forward (its design). Running them together would defeat AG's perf win — every step would compute uncond regardless of the gate fraction. When AG is inactive (`FASTVIDEO_CFG_GATE_STEP=1.0`, the default), batched-CFG fires normally. When AG is enabled, batched-CFG falls back to sequential and AG's delta-cache reuse runs unchanged.

   The PR body's "composes with #1372" line is now precise: they compose by **deferring to whichever is active**, not by stacking. If a future caller wants both wins, that's a separate scope (the batched path could maintain its own delta cache across batched-2 forwards — out of scope here).

History stays 4 logical commits — fixup squashed into commit 1 via `--autosquash`, conflict resolved cleanly in commit 1.

rich7420:
@Mister-Raggs thanks for the patch!
I've tried some config in modal

| leg | wall | SSIM worst |
|---|---|---|
| H100 14B eager (FA3) | -1.46% | 1.0 ✅ |
| L40S 1.3B eager (FA2) | +1.95% | 0.95 |
| H100 14B compile (FA3) | +0.13% | **0.52** |

So bit-equiv is real but **FA3+eager only** — FA2 already drifts at 0.96, and compile (what we actually deploy) gave me ~0 speedup + 0.52 worst. The win's basically just that one H100-eager cell.

Dug into why: Wan cross-attn has no text masking — `context_lens` is dead, hardcoded `None` (`wanvideo.py:383/533`), and the `encoder_attention_mask` you cat gets dropped by the sig filter. So the recipe's load-bearing (no mask → need equal-length pad), not gratuitous. Fair enough.

Couple nits:
- **default `False`** — it changes output for everyone who isn't on H100-eager, and "output-identical" only holds there
- recipe gate doesn't check `use_batched_cfg`, so `--use-batched-cfg=False` doesn't actually restore old Wan output. one-liner: `and fastvideo_args.use_batched_cfg` on the `_cfg_padding` gate
- drop the `target_dtype`→`dit_precision` bit (unrelated, hits stable_audio fp16), the `ltx2_3_base` test (rebase drift), and the reverted mask-math test
- heads up: `with_options` blows up on modal 1.3.5, had to patch the harness to run it

If you wanna go default-on later, the real fix is wiring that dead `context_lens` into the mask — then you pad to `max(real)`, get true bit-equiv with zero output change, and the whole text_encoding recipe disappears. Your mask-math test's a head start there.

SolitaryThinker:
Heads-up before you chase CI on the rebased head: `microscope-unit-tests` currently fails on **every** branch based on current main, and it's unrelated to this PR. #1388's change of `_resolve_flash_attn_varlen_func()` to return `(func, fa_version)` semantically conflicts with the resolver unit tests from #1540 (`test_flash_attn_no_pad_resolver.py` still asserts `__name__`/`__module__` on a bare function → `AttributeError: 'tuple' object has no attribute '__name__'`). The two never textually overlapped, so the merge was clean and the breakage only shows on heads containing both.

A 2-line test fix is staged on `maint/fix-resolver-test-post-1388` (unpack the tuple, assert the version leg) awaiting maintainer sign-off. Once it lands you'll need one more rebase/re-run to clear the lane. Everything else on your rebase looks unaffected — expected reds are exactly:
`test_resolver_skips_cute_without_opt_in` and `test_resolver_returns_flash_attn_when_interface_unavailable`.

PR #1422, opened 2026-06-01
[bugfix] LTX2: honor `video_position_offset_sec` in the DiT

H1yori233:
`video_position_offset_sec` was plumbed through the pipeline but swallowed by `LTX2Transformer3DModel.forward`'s `**kwargs` and never applied, so the audio continuation pre-roll had no matching video shift. In multi-segment continuation this accumulated into multi-second audio/video desync.

Fix: offset the video temporal RoPE positions (in seconds) by `video_position_offset_sec` — audio positions are already in seconds, so video aligns with the matching later portion of audio.


alexzms (approved):
lgtm

PR #1424, opened 2026-06-02
[feat] dreamverse: sequence parallelism for serving

shaoxiongduan:
Add DREAMVERSE_SP_SIZE to run a single generation across multiple GPUs via sequence parallelism. Defaults to 1, which is byte-for-byte the existing single-GPU-per-session behavior.

- config: DREAMVERSE_SP_SIZE env knob
- gpu_pool: group visible GPUs into SP slots, expose the whole group to the worker (CUDA_VISIBLE_DEVICES="g0,g1,..."), keyed by the first GPU
- video_generation: drive engine num_gpus from DREAMVERSE_SP_SIZE
- ltx2: drop @torch.compiler.disable on LTXDistributedAttention so the all-to-all distributed attention compiles into the fullgraph at sp>1

PR #1425, opened 2026-06-02
[feat]: decouple Dreamverse fMP4 streaming from generation

Davids048:
## Summary

- move Dreamverse fMP4 packaging out of the synchronous USER_STEP path into one serial worker-side background stream thread
- add MediaError for post-generation fMP4 failures and route media events through a SessionController media relay task
- keep ltx2_segment_complete and ltx2_stream_complete tied to media completion while StepComplete remains the generation/continuation-state boundary
- update session logging expectations for the split generation/media lifecycle

## Tests

- git diff --cached --check
- /home/hal-jundas/codes/FastVideo/.venv/bin/python -m py_compile apps/dreamverse/dreamverse/session/controller.py apps/dreamverse/dreamverse/tests/test_session_logging.py
- /home/hal-jundas/codes/FastVideo/.venv/bin/python -m pytest apps/dreamverse/dreamverse/tests/test_session_logging.py::test_websocket_flow_emits_required_session_events -q *(blocked during collection: Triton reports `RuntimeError: 0 active drivers ([]). There should only be one.`)*

### DIFF
--- a/apps/dreamverse/dreamverse/gpu_pool.py
+++ b/apps/dreamverse/dreamverse/gpu_pool.py
@@ -5,6 +5,7 @@ import asyncio
 import multiprocessing as mp
 import os
 import subprocess
+import threading
 import time
 import traceback
 from dataclasses import dataclass
@@ -35,6 +36,7 @@ from dreamverse.worker_ipc import (
     LeaveAck,
     MediaChunk,
     MediaComplete,
+    MediaError,
     MediaInit,
     ReloadAck,
     ReloadModelPayload,
@@ -46,6 +48,8 @@ from dreamverse.worker_ipc import (
     WorkerError,
     WorkerEvent,
 )
+
+ACTIVE_STREAM_JOIN_TIMEOUT_SECONDS = 15.0
 
 
 def _parse_requested_gpu_limit() -> int | None:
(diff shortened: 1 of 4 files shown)


SolitaryThinker:
**Hi @Davids048** — automated review from Gob, one of @SolitaryThinker's AI reviewers. I read the diff at `b4be1de7` against `origin/main` (PR #1424 just landed as `570607945`; merges cleanly, no rebase needed). Verdict: **ship-with-fixes**. Counts: 1 S1 / 2 S2 / 3 S3. All three prior Gemini findings are genuinely addressed at this HEAD — nice work on the timings snapshot, the 15s join timeout, and the `stop_event.set()` on `MediaError`.

The one thing that jumped out is the natural extension of Gemini's first finding: `step_result.timings` is correctly snapshotted with `dict(...)`, but `step_result.frames` and `step_result.audio` are still passed by reference into the background stream thread. The whole point of the refactor is to let `worker.generate_step` start the next segment while the previous one is still encoding — which is great — but that means the producer can overwrite frame/audio storage while the encoder is still reading it, if `VideoGenerationWorker.generate_step` reuses any pre-allocated output buffer. Worth either (a) cloning the frames/audio at the top of `start_stream_thread`, (b) asserting in a comment + tests that `generate_step` always returns freshly-allocated storage, or (c) calling `wait_for_active_stream()` before the next `generate_step` instead of before the next `start_stream_thread` (option c partially defeats the decoupling, so (a) or (b) is probably right).

The second-biggest gap is test coverage. The fake slot in `test_session_logging.py` still emits `MediaInit`/`Chunk`/`Complete` inline inside `user_step()`, which means none of the new async-relay timing is actually exercised by a test. Adding one test that (i) emits media events *after* `user_step` returns and asserts `ltx2_stream_complete` waits for the tail, and (ii) emits a `MediaError` and asserts the session terminates, would pin all the new invariants. The author already flagged the e2e test was blocked by a Triton driver error during local collection — worth getting that running too before merge.

A few smaller items in the review: `segment_total_hints` grows unbounded over a long session (just `pop` instead of `get` on `MediaComplete`); the `media_relay_loop` quietly removed the old out-of-order event guard and now depends on `wait_for_active_stream()` upstream — worth a one-line comment near the relay; the `USER_LEAVE` path now treats a 15s join timeout as a `WorkerError` that brings down the worker process, which is a behavior change worth confirming is what you want vs. just logging and acking the leave; and the `media_relay_loop` has no top-level `try/except` so an uncaught exception there would silently kill the relay while generation continues.

Full findings with file:line evidence in the attached review.

— Gob
(automated review; human review still required to land)

SolitaryThinker:
**Hi @Davids048** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

## TL;DR

Decouples Dreamverse fMP4 streaming from segment generation via a background encoder thread + async `media_relay_loop`. All three of gemini's prior flags are genuinely addressed at this HEAD (timings dict-snapshot, 15s join timeout, `stop_event.set()` on `MediaError`). The natural extension of gemini's first finding is the one S1 here: `step_result.frames` / `step_result.audio` are still passed by reference into the background thread while the next segment is being generated.

- S0 (blockers): 0
- S1 (must-fix): 1
- S2 (should-fix; surfaced if `persistent` or `important`): 2

---

### [S1] `step_result.frames` / `step_result.audio` passed by reference into the encoder thread

**What:** `step_result.timings` is correctly snapshotted via `dict(step_result.timings)` before the thread starts (good — that's gemini #1 addressed). But `step_result.frames` and `step_result.audio` are still passed by reference into the background stream thread, while the main loop is now free to call `worker.generate_step()` for the next segment without waiting.

**Why it matters:** The whole point of the refactor is to let `generate_step` start the next segment while the previous one is still encoding. If `VideoGenerationWorker.generate_step` ever reuses any pre-allocated output buffer for frames/audio (a common optimization), the producer overwrites the encoder's input mid-encode → corrupt segments. The window is exactly the speedup the refactor unlocks, so the data race scales with the speedup.

**Suggested fix:** Pick one:
(a) Clone frames/audio at the top of `start_stream_thread` (cheap, no behavior change required from worker).
(b) Assert in a comment + test that `generate_step` always returns freshly-allocated storage, and pin that with a regression test that mutates the original after handoff and confirms the encoder output is unaffected.
(c) Call `wait_for_active_stream()` *before* the next `generate_step` instead of before the next `start_stream_thread` — but this partially defeats the decoupling, so (a) or (b) is probably right.

**Evidence:** `apps/dreamverse/dreamverse/worker_ipc.py` `start_stream_thread` (around line 208 where `dict(step_result.timings)` snapshots; frames/audio remain by-reference).

---

### [S2] No test exercises the new async-relay timing

**What:** The fake slot in `apps/dreamverse/dreamverse/tests/test_session_logging.py` still emits `MediaInit` / `Chunk` / `Complete` inline inside `user_step()`. None of the new async-relay timing — segments emitting *after* `user_step` returns — is actually exercised by a test. Same gap on the `MediaError` path: no test asserts the session terminates and the GPU slot releases.

**Why it matters:** The refactor introduces a class of bugs (join-timeout deadlocks, out-of-order events, missed `MediaError` cleanup, frames/audio races per S1 above) that only manifest when generation and streaming actually overlap in time. The current test only proves the happy synchronous path still works — which is also what the old code did. The author's PR description already flags an e2e test as blocked by a Triton driver error during local collection.

**Suggested fix:** Add two tests in `test_session_logging.py`:
1. Have the fake slot emit `MediaInit` / `Chunk` / `Complete` *after* `user_step` returns (e.g., from a separate task/thread) and assert `ltx2_stream_complete` waits for the tail and the relay forwards every event in order.
2. Have the fake slot emit `MediaError` and assert `stop_event.set()` fires, the session reaches a terminal state, and the GPU slot is released.

Resolving the local-e2e Triton blocker before merge would also retire significant risk.

**Evidence:** `apps/dreamverse/dreamverse/tests/test_session_logging.py` (fake slot's inline event emission); `apps/dreamverse/dreamverse/session/controller.py` `media_relay_loop` + `MediaError` handling.

---

### [S2] `segment_total_hints` grows unbounded over the session lifetime

**What:** `segment_total_hints` is populated when a segment starts but never pruned. A long session (many segments) walks it indefinitely; over a long-lived worker it's a slow leak.

**Why it matters:** Not a crash, but the relay path is now the long-lived hot path of the worker — anything in it that grows with session length will surface as either memory growth or O(N) lookup cost over time. Cheap to fix.

**Suggested fix:** Replace the `get` on `MediaComplete` with `pop`, so each hint is consumed exactly once when its terminal event fires.

**Evidence:** `apps/dreamverse/dreamverse/session/controller.py` (`segment_total_hints` populated near `MediaInit` handling; consumed at `MediaComplete`).

---

PR #1426, opened 2026-06-03
[feat] Optional cache-dit step caching for the Wan DiT

Mister-Raggs:
### Optional step caching for the Wan DiT via cache-dit

**What & why.** Adds opt-in, default-off step caching for Wan inference using the [cache-dit](https://github.com/vipshop/cache-dit) library. Diffusion runs the DiT once per denoising step, and adjacent steps produce nearly-identical features — cache-dit's DBCache skips the middle DiT blocks on steps whose leading-block residual barely changes and reuses a cached result. An optional **TaylorSeer** calibrator extrapolates that residual (Taylor expansion) instead of holding it constant, giving higher fidelity at the same skip rate. Lossy (SSIM < 1.0), so default-off and gated behind `--use-cachedit`.

**How.** cache-dit attaches through a transformer-only `BlockAdapter` (no diffusers pipeline needed). `DenoisingStage._enable_or_refresh_cachedit()` enables the cache once per transformer and refreshes the cache context each generation (`num_inference_steps` + `refresh_context`) so state never leaks across prompts. Wan's separate cond/uncond forwards are handled via `enable_separate_cfg=True` / `cfg_compute_first=False`.

**Results** — Wan2.1-T2V-1.3B, 1×L40S, 720×1280 / 77f / 30 steps, 5 prompts, eager. Two operating points via `--cachedit-residual-threshold` (both F8B0 + TaylorSeer):

| Preset | threshold | wall | SSIM mean / worst |
|---|---|---|---|
| Quality (default) | 0.08 | −18% | 0.957 / 0.941 |
| Performance | 0.15 | **−32%** | 0.943 / 0.920 |

TaylorSeer is the key to the aggressive preset: spot-checked frames including a high-detail space-station cupola are visually clean at **both** thresholds. Without TaylorSeer (plain DBCache constant-residual reuse), the same −32% operating point shows undersolved-noise artifacts on that high-detail content — extrapolating the residual is what keeps the aggressive setting clean.

**Caveats.**
- Incompatible with DiT offloading — caching skips blocks, the offload hook assumes every block runs each step; guarded with a clear error (set `--dit-cpu-offload False --dit-layerwise-offload False`).
- Eager only; the `torch.compile` path is rejected with a clear error (caching adds data-dependent control flow). Follow-up.
- Wan DiT only for now.
- cache-dit is added as an **optional** `[cache]` extra, lazy-imported — not a runtime dependency. (Flagging the new optional dependency explicitly for maintainer visibility.)

**Test evidence.** Modal A/B harness `fastvideo/tests/modal/cachedit_ab.py` (baseline vs cached, per-prompt wall + SSIM) — produced the numbers above. Reproduce: `modal run fastvideo/tests/modal/cachedit_ab.py --gpu L40S --taylorseer` (quality) or `--taylorseer --threshold 0.15` (performance).

### DIFF
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -95,6 +95,13 @@ surfaces:
       moba_config: "Derived runtime config loaded from moba_config_path."
       model_paths: "Runtime bookkeeping."
       model_loaded: "Runtime bookkeeping."
+      use_cachedit: "Opt-in cache-dit step-caching toggle (Wan DiT); runtime plumbing, not part of the typed API yet."
+      cachedit_fn_compute_blocks: "cache-dit leading-block count; runtime plumbing for use_cachedit."
+      cachedit_bn_compute_blocks: "cache-dit trailing-block count; runtime plumbing for use_cachedit."
+      cachedit_residual_threshold: "cache-dit residual-diff skip threshold; runtime plumbing for use_cachedit."
+      cachedit_max_warmup_steps: "cache-dit warmup-step count; runtime plumbing for use_cachedit."
+      cachedit_taylorseer: "cache-dit TaylorSeer calibrator toggle; runtime plumbing for use_cachedit."
+      cachedit_taylorseer_order: "cache-dit TaylorSeer expansion order; runtime plumbing for use_cachedit."
 
   pipeline_config_base:
     moved:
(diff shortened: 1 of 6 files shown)


Mister-Raggs:
The cachedit schema-parity failure is fixed in the latest push. The other CI failure (test_ltx2_presets_registered) is pre-existing on main - ltx2_3_base is registered and is the LTX2 default, but the test still expects the old 3-preset set; this PR touches no LTX2/preset code. Filed separately as #1427.

Re: the hf auth login suggestion - keeping it. hf is the current huggingface_hub CLI entrypoint (huggingface-cli is the legacy alias), and this bootstrap line is validated working on the fastvideo-dev image in our A/B runs.

PR #1440, opened 2026-06-06
[bugfix] EMA in distillation pipeline

H1yori233:
## Purpose

In `fastvideo/training/distillation_pipeline.py` (used by non–self-forcing DMD pipelines), EMA had two bugs that made `--use_ema True` produce blurry results and never export EMA weights.

**Bug 1 — EMA shadow anchored to the cold init model (with `ema_start_step > 0`).**
The EMA was **eager-constructed in `__init__` from the (untrained) init weights** and updated from step 0. The lazy-construction block in the train loop was dead code.
```python
if (step >= self.training_args.ema_start_step) and (self.generator_ema is None) and (self.training_args.ema_decay > 0):
    self.generator_ema = EMA_FSDP(self.transformer, decay=...)
```
because the eager construct already set `generator_ema` non-`None`. So `ema_start_step` only gated whether EMA is used (`is_ema_ready`), not when the shadow starts accumulating. The shadow therefore averaged in the base/init model.

**Bug 2 — `save_ema_weights` crashes under FSDP (EMA never exported).**
`save_ema_weights` → `get_ema_model_copy()` did `copy.deepcopy(self.transformer)`, which FSDP does not support. Every save raised and was caught:
```
ERROR Failed to save EMA weights: FSDP does not support deepcopy. Please use state dict for serialization
```
As a result no EMA weights were ever written; the exported `generator_inference_transformer/*.safetensors` were the raw (non-EMA) weights.

## Changes

- **Bug 1:** only eager-construct the EMA in `__init__` when `ema_start_step <= 0`; otherwise defer so the existing lazy block builds it at `ema_start_step` from the already-trained weights (the per-step update is already guarded by `if self.generator_ema is not None`, so it correctly does nothing until then). This matches the pattern already used in `self_forcing_distillation_pipeline.py`.
- **Bug 2:** remove `get_ema_model_copy`/`get_ema_2_model_copy`; rewrite `save_ema_weights` to swap EMA weights into the live FSDP module **in place** via `generator_ema.apply_to_model(self.transformer)` and gather with `gather_state_dict_on_cpu_rank0` inside that context (no deepcopy; live weights restored on exit). The gather is collective-safe — all ranks call it; only rank 0 writes the file.

## Test Plan

Ran a 4-step DMD distillation on Wan2.1-T2V-1.3B with EMA starting mid-run:

```bash
torchrun --nnodes 1 --nproc_per_node 4 \
  fastvideo/training/wan_distillation_pipeline.py \
  --num_gpus 4 --sp_size 1 --tp_size 1 --hsdp_replicate_dim 4 --hsdp_shard_dim 1 \
  --model_path Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --pretrained_model_name_or_path Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --real_score_model_path Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --fake_score_model_path Wan-AI/Wan2.1-T2V-1.3B-Diffusers \
  --data_path <small_parquet_dir> --dataloader_num_workers 2 \
  --output_dir scratch/ema_dryrun_ckpt \
  --max_train_steps 4 --train_batch_size 1 --gradient_accumulation_steps 1 \
  --num_latent_t 20 --num_height 480 --num_width 832 --num_frames 77 \
  --enable_gradient_checkpointing_type full --mixed_precision bf16 \
  --training_state_checkpointing_steps 2 --weight_only_checkpointing_steps 2 \
  --use_ema True --ema_start_step 2 --ema_decay 0.99 \
  --flow_shift 8 --dmd_denoising_steps '1000,757,522' \
  --generator_update_interval 5 --real_score_guidance_scale 2.0
```

## Test Results

- `Generator EMA deferred: built lazily at ema_start_step=2 from trained weights` — deferred init.
- `ema_checkpoint-{2,4}/diffusion_pytorch_model.safetensors` written on disk; run completes 4/4 cleanly.

## Checklist

- [X] I ran `pre-commit run --all-files` and fixed all issues
- [X] I updated documentation if needed

### DIFF
--- a/fastvideo/training/distillation_pipeline.py
+++ b/fastvideo/training/distillation_pipeline.py
@@ -371,7 +371,8 @@ class DistillationPipeline(TrainingPipeline):
                                                                    self.transformer.reverse_param_names_mapping)
                     save_file(diffusers_state_dict, weight_path)
 
-                    config_dict = self.transformer.hf_config
+                    # deepcopy so deleting "dtype" doesn't mutate the live model's hf_config
+                    config_dict = copy.deepcopy(self.transformer.hf_config)
                     if "dtype" in config_dict:
                         del config_dict["dtype"]
                     config_path = os.path.join(ema_save_dir, "config.json")
@@ -398,7 +399,8 @@ class DistillationPipeline(TrainingPipeline):
                                                                      self.transformer_2.reverse_param_names_mapping)
                     save_file(diffusers_state_dict_2, weight_path_2)
 
-                    config_dict_2 = self.transformer_2.hf_config
+                    # deepcopy so deleting "dtype" doesn't mutate the live model's hf_config
+                    config_dict_2 = copy.deepcopy(self.transformer_2.hf_config)
                     if "dtype" in config_dict_2:
                         del config_dict_2["dtype"]
                     config_path_2 = os.path.join(ema_2_save_dir, "config.json")
--- a/fastvideo/training/distillation_pipeline.py
+++ b/fastvideo/training/distillation_pipeline.py
@@ -245,7 +245,9 @@ class DistillationPipeline(TrainingPipeline):
 
         self.generator_ema: EMA_FSDP | None = None
         self.generator_ema_2: EMA_FSDP | None = None
-        if (self.training_args.ema_decay is not None) and (self.training_args.ema_decay > 0.0):
+        ema_enabled = (self.training_args.ema_decay is not None) and (self.training_args.ema_decay > 0.0)
+        if ema_enabled and (self.training_args.ema_start_step <= 0):
+            # Only eager-construct from the cold init weights when averaging starts at step 0.
             self.generator_ema = EMA_FSDP(self.transformer, decay=self.training_args.ema_decay)
             logger.info("Initialized generator EMA with decay=%s", self.training_args.ema_decay)
 
@@ -253,6 +255,12 @@ class DistillationPipeline(TrainingPipeline):
             if self.transformer_2 is not None:
                 self.generator_ema_2 = EMA_FSDP(self.transformer_2, decay=self.training_args.ema_decay)
                 logger.info("Initialized generator EMA_2 with decay=%s", self.training_args.ema_decay)
+        elif ema_enabled:
+            # Defer construction to the lazy block in the train loop, which builds the EMA AT
+            # ema_start_step from the already-trained weights. Eager-constructing here would anchor
+            # the shadow to the cold init and leave it base-contaminated (blurry) on short runs.
+            logger.info("Generator EMA deferred: built lazily at ema_start_step=%s from trained weights",
+                        self.training_args.ema_start_step)
         else:
             logger.info("Generator EMA disabled (ema_decay <= 0.0)")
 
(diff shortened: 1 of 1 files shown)


SolitaryThinker (commented):

## TL;DR
Verified bugfix — both EMA bugs are real and the fixes are correct. Deferred EMA construction matches the existing self-forcing pipeline pattern, and the FSDP save now swaps EMA shards into the live module via a restoring context manager (no deepcopy). **Verdict: APPROVE** (1 minor follow-up, 1 nit). 0 blockers / 0 major.

## Verified against the code
- **`apply_to_model`** (`training_utils.py` `_ApplyEMACtx`) swaps EMA shards on `__enter__` and **restores live weights on `__exit__`** — confirmed; no FSDP deepcopy.
- **Collective-safe save**: `gather_state_dict_on_cpu_rank0` is called by all ranks inside the `with`; only rank 0 writes. No rank-divergence/hang.
- **Lazy build now reachable**: the deferred path lets the `generator_ema is None` block (`distillation_pipeline.py:1336–1342`) build the EMA at `ema_start_step` from trained weights — structurally matches `self_forcing_distillation_pipeline.py:856–862`.
- **EMA mode contract** holds: `apply_to_model` requires `mode="local_shard"`; `EMA_FSDP` defaults to it.
- The `hf_config` deepcopy now guards against mutating the live model — addresses the earlier automated-review comment.

## Minor (follow-up, non-blocking)
`save_ema_weights` still wraps its whole body in a broad `try/except Exception` that logs and swallows — that's exactly how Bug 2 stayed silent. Since this PR's point is "EMA was silently never saved," consider narrowing the catch or re-raising after logging so a future regression fails loudly.

## Nit
The transformer / transformer_2 save branches are near-identical ~25-line blocks; a shared `_save_one_ema(...)` helper would halve the surface next time (not needed for this PR).

SolitaryThinker (approved):

SolitaryThinker (commented):
**Hi @H1yori233** — follow-up automated review from Gob, one of @SolitaryThinker's AI reviewers. This PR is already merged; flagging a **post-merge regression** worth a quick fix. Ping @SolitaryThinker if anything looks off.

## TL;DR
The two EMA bugfixes here are correct and verified. But the Bug-1 fix (deferring EMA construction when `ema_start_step > 0`) introduces a **new regression on resume**: a checkpointed EMA shadow is now silently discarded when resuming any run with `ema_start_step > 0`. Worth a follow-up PR — the happy-path bugs this fixed are worse than the resume gap it opens, so the merge was reasonable, but this shouldn't linger.

## MAJOR — EMA shadow silently dropped on resume (`ema_start_step > 0`)

**What:** With the deferral fix, `initialize_training_pipeline()` leaves `self.generator_ema = None` when `ema_start_step > 0` (`distillation_pipeline.py:258-263`, the `elif ema_enabled:` defer branch). In `train()`, resume runs **before** the lazy EMA build:

- `distillation_pipeline.py:1279` — `_resume_from_checkpoint()` runs, which calls `load_distillation_checkpoint(..., self.generator_ema, ...)` with `generator_ema=None` (passed at `_resume_from_checkpoint`, ~`:920`).
- `training_utils.py:617` — the EMA-shard reload is guarded by `if generator_ema is not None:`, so with `None` the saved `ema_local_shard/generator_ema_rank*.pt` is **skipped**.
- `distillation_pipeline.py:1321-1322` — the lazy block then builds a **fresh** EMA from the resumed live weights, throwing away all checkpointed EMA history.

**Why it matters:** Pre-PR, the unconditional eager construct kept `generator_ema` non-`None` at resume, so the shard *was* loaded. This is a behavior regression specifically for **resumed long runs** — which are exactly the runs that set `ema_start_step > 0`. The EMA you resume with is silently reset to a fresh average from the current weights, losing the accumulated shadow (and `save_distillation_checkpoint` *does* save it, at `:1411/1437/1473`, so the data is on disk — just not reloaded).

**Suggested fix:** construct the EMA before the resume load when a resume is requested and `ema_start_step > 0` (so the shard reloads into a real object), or have `_resume_from_checkpoint` build-then-load the EMA, or pass a flag so `load_distillation_checkpoint` lazily instantiates it when a shard exists. Any of the three closes the gap; the first is smallest.

**Evidence (verified against merged `a931efe3`):** `train()` order init→resume(`:1279`)→lazy(`:1321`); save writes the shadow (`save_distillation_checkpoint(..., self.generator_ema, ...)`); load guard at `training_utils.py:617`; the lazy block contains no shard-reload.

## On the two original bugfixes (both correct — for the record)
- **Bug 1 (deferred construction):** correct — stops the shadow anchoring to cold-init weights; matches the `self_forcing_distillation_pipeline.py` pattern.
- **Bug 2 (FSDP-safe save):** correct — `apply_to_model` is a restoring context manager (swaps EMA shards in, restores on exit, no deepcopy); `gather_state_dict_on_cpu_rank0` is collective (all ranks call it, rank 0 writes). The `hf_config` deepcopy fix is also correct.

## Verdict
**COMMENT** (post-merge) — both fixes land correctly; one MAJOR follow-up (EMA-on-resume) recommended. No revert warranted.

PR #1441, opened 2026-06-07
[bugfix] EMA shadow on resume and EMA under MoE path

H1yori233:
## Purpose

This is a follow-up to the EMA distillation fix (#1440). That PR made EMA construction deferred when `ema_start_step > 0`. That deferral introduced a resume regression, and while fixing it I found a second, pre-existing EMA bug in the MoE path. Both live in `fastvideo/training/distillation_pipeline.py` / `fastvideo/training/training_utils.py`.

**Bug 1 — EMA shadow silently dropped on resume (`ema_start_step > 0`).**
With deferral, `generator_ema` is `None` at resume time, so `_resume_from_checkpoint` passes `None` into `load_distillation_checkpoint`, and the EMA-load is guarded by:

```python
if generator_ema is not None:   # training_utils.py
    ... load ema_local_shard/generator_ema_rank{rank}.pt ...
```

So the saved shard is skipped, and the train-loop lazy block then rebuilds a fresh EMA from the resumed weights — silently discarding the accumulated shadow. Pre-#1440 the eager construct kept `generator_ema` non-`None` at resume, so the shard was loaded; this is a behavior regression specifically for resumed long runs (exactly the runs that set `ema_start_step > 0`).

**Bug 2 — EMA_2 (MoE) never restored on resume.**
`save_distillation_checkpoint` writes the second-expert EMA to its own shard, but `load_distillation_checkpoint` read it from a distcp dict key that the save side never populates:

```python
# save  -> ema_local_shard/generator_ema_2_rank{rank}.pt   (training_utils.py:402)
# load  -> ema_2_state = generator_2_states.get("ema")      (training_utils.py:672, always None)
```

So `generator_ema_2` was never restored on resume.

## Changes

- **Bug 1:** add `_build_deferred_ema_for_resume()` and call it in `_resume_from_checkpoint` before the load.
- **Bug 2:** load EMA_2 from `ema_local_shard/generator_ema_2_rank{rank}.pt`, symmetric with the main expert, and delete the dead `generator_2_states.get("ema")` read.
- **Cleanup:** centralize the 6 scattered `EMA_FSDP(...)` construction sites into one idempotent, **per-expert** helper `_build_generator_emas()`.

## Test Plan
Use mock scripts to verify that the bugs actually exist and that the fixes are correctly applied.

## Test Results

```
# ema_sim_test.py
PASS A1 both EMAs built when decay>0 & transformer_2 present
PASS A2 nothing built when decay<=0
PASS A3 only main EMA built when no transformer_2
PASS A4 idempotent: existing main EMA untouched (same object)
PASS A4 decoupled: EMA_2 still built despite main present
PASS A5 resume: main EMA NOT pre-built (no main shard)
PASS A5 resume: EMA_2 pre-built (ema2 shard exists)
PASS A6 resume before start (no shards): both stay None
PASS B0 fresh shadows differ from saved (pre-load)
PASS B1 (Bug1) main EMA shadow restored from shard on resume
PASS B2 (Bug2) EMA_2 shadow restored from shard on resume
11/11 passed

# ema_oldnew_check.py  (Bug 2 existence + fix, identical checkpoint)
OLD (origin/main) load: EMA_2 restored from shard? False
NEW (my fix)       load: EMA_2 restored from shard? True
Bug 2 existence+fix demonstrated: True
```

## Checklist

- [X] I ran pre-commit run --all-files and fixed all issues
- [X] I updated documentation if needed

### DIFF
--- a/fastvideo/training/distillation_pipeline.py
+++ b/fastvideo/training/distillation_pipeline.py
@@ -248,13 +248,7 @@ class DistillationPipeline(TrainingPipeline):
         ema_enabled = (self.training_args.ema_decay is not None) and (self.training_args.ema_decay > 0.0)
         if ema_enabled and (self.training_args.ema_start_step <= 0):
             # Only eager-construct from the cold init weights when averaging starts at step 0.
-            self.generator_ema = EMA_FSDP(self.transformer, decay=self.training_args.ema_decay)
-            logger.info("Initialized generator EMA with decay=%s", self.training_args.ema_decay)
-
-            # Initialize EMA for transformer_2 if it exists
-            if self.transformer_2 is not None:
-                self.generator_ema_2 = EMA_FSDP(self.transformer_2, decay=self.training_args.ema_decay)
-                logger.info("Initialized generator EMA_2 with decay=%s", self.training_args.ema_decay)
+            self._build_generator_emas(context="eager init, ema_start_step<=0")
         elif ema_enabled:
             # Defer construction to the lazy block in the train loop, which builds the EMA AT
             # ema_start_step from the already-trained weights. Eager-constructing here would anchor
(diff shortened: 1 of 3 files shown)


SolitaryThinker (approved):

## Verdict: APPROVE

Open PR, training scope. Both fixes verified correct end-to-end against head `86f9298d`. No blockers or major concerns. CI green on `pre-commit` + `fastcheck-passed`; still needs the required `full-suite-passed` + 1 approval.

This is the right follow-up to the #1440 resume regression. Traced both paths:

- **Bug 1 (resume):** `_build_deferred_ema_for_resume()` runs at `distillation_pipeline.py:938` *before* `load_distillation_checkpoint` at `:940`, builds the shadow only when the shard exists, and the freshly-built `EMA_FSDP` defaults to `mode="local_shard"` (`training_utils.py:1592`) so it matches the load branch (`:619`). `load_state_dict` fully replaces `self.shadow` (`:1662`), so the throwaway cold shadow is overwritten — no cold-init contamination. The idempotent lazy block then no-ops. ✅
- **Bug 2 (MoE EMA_2):** load now reads `ema_local_shard/generator_ema_2_rank{rank}.pt` (`:637`), an exact match to the save path (`:402`); the dead distcp `.get("ema")` read is removed. ✅
- **self_forcing** inherits `_resume_from_checkpoint` (not overridden), so the Bug 1 fix covers it too.
- **rank0_full** load-branch removal is safe: both pipelines only ever construct EMAs in the default `local_shard` mode, and `rank0_full` is export-only (saved as `.safetensors`, which `torch.load` couldn't read anyway) — now handled by the explicit "not supported for resume; skipping" `else`.

### Minor (non-blocking)
- No committed regression test. The body's local `ema_sim_test.py` (11/11) + `ema_oldnew_check.py` evidence is credible and I verified the logic, but nothing lands in `fastvideo/tests/training/`. Both EMA bugs were silent and survived a merge; the save/load path is pure-CPU `torch.save`/`torch.load` on small dicts, so a small unit test (deferred build only when a shard exists; EMA_2 restores from its shard) is cheap and would lock in the invariant. Worth a follow-up.

### Nit
- `distillation_pipeline.py:921`: an earlier automated review flagged the missing `resume_from_checkpoint` None-guard before `os.path.join`. Not reachable today — the sole caller (`_resume_from_checkpoint`) only runs under `if self.training_args.resume_from_checkpoint:` — so optional defense-in-depth, not required.

Out of scope: the pre-existing MoE *save*-path rank-0-write-between-collectives concern from the #1440 review is untouched here (this PR's `training_utils.py` hunks are confined to `load_distillation_checkpoint`); track it separately if still wanted.

PR #1434, opened 2026-06-05
[bugfix]: release VSA tile cache during training

aryan5v:
## Summary
- Fixes #1423.
- Disable the persistent VSA tiled Q/K/V/gate metadata cache for training so activation checkpointing can release the scratch buffer after attention.
- Keep the existing cache behavior enabled by default for inference.

## Repro and validation
- Reproduced the OOM on 8x NVIDIA A100-SXM4-80GB with Wan2.1-T2V-1.3B, torch 2.11.0, fastvideo-kernel 0.2.6, batch size 2, num_latent_t=81, 480x832, and VSA sparsity 0.9.
- Verified the patched run completes the same 8x A100 training step and saves a checkpoint: loss=0.0477, step_time=94.29s, grad_norm=0.508.
- Ran py_compile on the touched Python files.

Issue #1423, opened 2026-06-02
[Bug] Enabling VSA (USE_VSA=1) OOMs on 8×A100-80GB for Wan2.1-1.3B full finetune, while dense attention fits


chenj02:
## Summary
Running the step-time test for **Wan2.1-T2V-1.3B full finetune** on **8×A100-80GB**, enabling **Video Sparse Attention** (`USE_VSA=1`, `FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN`) **OOMs at step 0**, whereas the **dense (FlashAttention) path runs fine**. Counterintuitively, VSA — which is meant to reduce attention cost — **increases** peak memory here, and the increase does **not** shrink when batch size is reduced.

## Repro
```bash
# scripts/finetune/benchmark_wan1.3b.sh, key settings:
#   num_latent_t=81 (~20s), 480x832, train_batch_size=2
#   sp_size=1, tp_size=1, hsdp_shard_dim=8, hsdp_replicate_dim=1
#   dit_precision=fp32, mixed_precision=bf16, enable_gradient_checkpointing_type=full
USE_VSA=1 VSA_SPARSITY=0.9 bash scripts/finetune/benchmark_wan1.3b.sh   # -> OOM at step 0
USE_VSA=0                  bash scripts/finetune/benchmark_wan1.3b.sh   # -> runs fine
```

OOM (representative):
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 4.22 GiB.
GPU 0 ... 70.79 GiB allocated by PyTorch ... (at FFN GELU, in activation-checkpoint recompute)
```

## Measured peak memory (GPU0, same env, GPU empty at start ~700MB)
| Config | Peak | Result |
| --- | --- | --- |
| dense (flash), batch=2 | **73.5 GB** | ✅ runs (~165 s/step) |
| dense (flash), batch=1 | ~40 GB (ran) | ✅ runs (88 s/step) |
| **VSA 0.9, batch=2** | **79.7 GB** | ❌ OOM @ step 0 |
| **VSA 0.9, batch=1** | **80.7 GB** | ❌ OOM @ step 0 |
| VSA 0.9, batch=1 + `expandable_segments:True` | still ~80 GB (frag 4.7→0.3 GB) | ❌ OOM @ step 0 |

## Environment
- GPU: 8× NVIDIA A100-SXM4-80GB, driver 535.261.03 (CUDA 12.4)
[...]

PR #1444, opened 2026-06-09
[feat]: make VSA tile cache configurable for training

SolitaryThinker:
## What

Make the VSA per-step tile-buffer cache (`cache_tile_buf`) **configurable for training**, instead of forcing it off.

## Why

#1434 fixed the [#1423](https://github.com/hao-ai-lab/FastVideo/issues/1423) activation-checkpointing OOM by hard-coding `cache_tile_buf=False` at both VSA training build sites. The cached padded buffer was pinned on the per-step `attn_metadata`, so under full activation checkpointing it survived into the backward recompute and inflated peak memory.

But the cache is purely a per-step speed optimization (reuse the padded QKVG tile scratch across VSA layers). On memory-rich setups — FSDP/SP-sharded large clusters, or runs that don't use full activation checkpointing — the OOM doesn't bite, and keeping the cache is a free speedup. Hard-coding `False` takes that off the table.

This exposes it as a config knob, **defaulting to `False`** to preserve #1434's OOM-safe training behavior. Opt in with `True` to keep the buffer-reuse speedup.

## Changes

- **New framework**: `TrainingConfig.vsa_cache_tile_buf` (YAML `training.vsa.cache_tile_buf`), parsed in `config.py`, wired into `WanModel._build_attention_metadata` (`wan.py`). Inherited by cosmos/hunyuan/matrixgame2/wan_causal.
- **Legacy**: `TrainingArgs.VSA_cache_tile_buf` (`--VSA-cache-tile-buf`), wired into `TrainingPipeline._build_attention_metadata` (`training_pipeline.py`). Inherited by distillation + self-forcing.
- **Inference unchanged** — the LTX2 build still omits the kwarg → default `True`.
- **Test**: added a value-equivalence test asserting the cached and non-cached `tile()` paths produce **bit-identical** tilings, on a padded shape (`raw_latent_shape=(5,4,4)` → padded T=8) that actually exercises the zero-pad-position scatter. This pins the property that the flag is a pure performance knob, never a correctness change — and addresses the two MINOR test-depth suggestions from the #1434 review.
- **Docs**: documented `cache_tile_buf` in `examples/train/configs/example.yaml`.

## Stacked on #1434

This is **stacked on #1434** (which introduces the `cache_tile_buf` plumbing — `main` has no such field yet). The net-new change here is one commit; the diff against `main` shows #1434's commit too until it merges. Once #1434 lands, I'll rebase and this collapses to just the configurability commit. Please review/merge #1434 first.

## Test plan

- `ruff` / `yapf` / `codespell` clean on all changed files.
- New equivalence test logic verified standalone (cached == uncached; padding present at `(5,4,4)`).
- Existing `cache_tile_buf` plumbing tests from #1434 still pass (structure unchanged).

### DIFF
--- a/examples/train/configs/example.yaml
+++ b/examples/train/configs/example.yaml
@@ -129,6 +129,11 @@ training:
   # --- training.vsa [TYPED] -> VSAConfig ---
   vsa:
     sparsity: 0.0                     # default: 0.0 (0.0 = disabled)
+    cache_tile_buf: false             # default: false. Reuse the per-step padded
+                                      # VSA tile buffer across attention layers.
+                                      # Off by default to avoid the activation-
+                                      # checkpointing OOM (#1423); set true on
+                                      # memory-rich setups for the buffer-reuse speedup.
 
   # --- training.model [TYPED] -> ModelTrainingConfig ---
   model:
--- a/fastvideo/fastvideo_args.py
+++ b/fastvideo/fastvideo_args.py
@@ -931,6 +931,12 @@ class TrainingArgs(FastVideoArgs):
     # VSA training decay parameters
     VSA_decay_rate: float = 0.01  # decay rate -> 0.02
     VSA_decay_interval_steps: int = 1  # decay interval steps -> 50
+    # Reuse the per-step padded VSA tile buffer across attention layers during
+    # training. Defaults to False: under full activation checkpointing the
+    # cached buffer survives into the backward recompute and inflates peak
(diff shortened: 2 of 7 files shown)


alexzms (approved):
LGTM, clean and minimal. Verified the new equivalence test passes locally (cached vs uncached `tile()` bit-identical), so the flag stays a pure perf knob, and default `false` preserves the #1434 OOM-safe behavior.

PR #1447, opened 2026-06-10
[infra] Add DGX Spark and multi-architecture CUDA support

SolitaryThinker:
## Summary

Adds first-class NVIDIA DGX Spark / Linux ARM64 support and aligns FastVideo's
installation, CUDA dependencies, Docker images, CI, and Dreamverse integrations
around the supported PyTorch 2.12 CUDA lanes.

The supported NVIDIA configurations are now:

- CUDA 12.6 with `UV_TORCH_BACKEND=cu126`
- CUDA 13.0 with `UV_TORCH_BACKEND=cu130`
- PyTorch pinned to `torch==2.12.0`

CUDA 12.8 / cu128 is intentionally omitted because PyTorch 2.12 does not publish
that backend. Installation commands use explicit backends rather than `auto`, so
resolution does not depend on the host driver's nearest CUDA index.

## DGX Spark installation

- Adds a dedicated DGX Spark (GB10, `aarch64`, CUDA 13, `sm_121`) guide.
- Uses a uv-managed Python 3.12 environment with development headers.
- Builds the in-tree `fastvideo-kernel` source because PyPI has no ARM64 kernel
  wheel, including an explicit `TORCH_CUDA_ARCH_LIST=12.1` path for GPU-less
  builders.
- Uses the prebuilt FlashAttention 2.8.3 ARM64 wheel for Python 3.12, PyTorch
  2.12, and cu130 instead of compiling FlashAttention from source.
- Includes import, CLI, and compiled-kernel execution checks plus troubleshooting.

## Docker images

- Replaces the per-Python CUDA Dockerfiles with one parameterized
  `docker/Dockerfile`.
- Builds the primary `fastvideo-dev` matrix natively for both `amd64` and
  `arm64` on CUDA 12.6.3 and CUDA 13.0.0.
- Selects architecture-specific prebuilt FlashAttention wheels:
  - amd64: wheel release `v0.9.17`, platform `linux_x86_64`
  - arm64: wheel release `v0.9.22`, platform `linux_aarch64`
- Pushes architecture images by digest and creates the shared multi-platform
  tags only after both builds complete, avoiding tag races.
- Makes CUDA 12.6/cu126 the default published image for broader L40S driver
  compatibility; CUDA 13/cu130 remains available under an explicit tag.

| Tag | CUDA / torch backend | Platforms |
|---|---|---|
| `latest`, `py3.12-latest`, `py3.12-cuda12.6.3-latest` | CUDA 12.6 / cu126 | linux/amd64, linux/arm64 (GH200 class) |
| `py3.12-cuda13.0.0-latest` | CUDA 13.0 / cu130 | linux/amd64, linux/arm64 (DGX Spark) |

Buildkite continues to use `py3.12-latest`, which now selects cu126. Modal
launchers inherit the backend baked into mutable registry tags, while honoring
caller overrides and providing fallbacks for explicit CUDA-version tags. This
keeps the tag transition safe before and after the release workflow runs.

Dreamverse images remain amd64-only for now because their FA4 dependency stack
is not yet validated on ARM64. Their existing CUDA 12.6/13 backend and UI tags
are unchanged.

## Additional integration fixes

- Copies the ARM kernel source before Docker dependency resolution.
- Translates every entry in multi-value `CMAKE_CUDA_ARCHITECTURES` lists.
- Keeps Dreamverse CUDA build overrides and Modal image tags aligned with the
  new CUDA matrix.
- Updates all Buildkite Dockerfile triggers, including the eval lane.
- Keeps the cutlass-4.5-compatible FA4 overlay on validated x86 builds and
  falls back cleanly to FA2/FA3 when FA4 is unavailable or incompatible.
- Updates installation, contributor, RunPod, ComfyUI, inference, and CI docs for
  the unified CUDA support policy.

## Validation

- `pre-commit run --files <changed files>` (including actionlint, PyMarkdown,
  codespell, ruff/yapf where applicable, and mypy)
- `python scripts/check_docs_links.py`
- `git diff --check`
- cu126 and cu130 dependency resolution with pinned PyTorch 2.12
- Static Docker wheel-selection checks for amd64/arm64 and cu126/cu130
- Verified all four referenced FlashAttention wheel assets in their GitHub
  releases
- Executed the manifest tag-generation script for both CUDA lanes and asserted
  the complete resulting tag sets
- Verified Modal backend behavior for mutable tags, explicit CUDA tags, custom
  images, and caller overrides
- Previously verified the source installation and compiled int8 kernel on an
  actual DGX Spark (GB10, CUDA 13.0, driver 580)

A full Docker image build was not run locally because Docker is unavailable in
this development environment; the manual image workflow performs the native
architecture builds.

## Update (2026-07-03): CI environment fixes + SSIM reference reseed

Running the SSIM suite against this branch surfaced several environment bugs
introduced by the torch/CUDA migration; all are fixed on this branch:

- `7e8028ba` — FA4 cute `_flash_attn_fwd` returns 4 values at the pinned ref;
  the 2-tuple unpack crashed every FLASH_ATTN inference (12/16 SSIM tasks).
- `1ae72ed1` + `d7156796` — Modal caches mutable registry tags and served the
  stale pre-transition image; launchers now resolve `tag@sha256:digest` at
  launch (`modal_image_utils.py`), so image publishes are picked up with no
  manual tag bumps.
- `13b23511` — cap `tokenizers<0.23`: 0.23's rebuilt pyo3 bindings reject the
  `cls=` kwarg transformers passes for CLIP-style tokenizers, breaking worker
  init for GameCraft/SD3.5/LTX2/Matrix-Game-3. Published images should be
  rebuilt to bake the capped resolve.
- `c3161945` — FA4 cute JIT can fail at runtime for some shape families (GQA
  on sm_89); fall back to FA2 once per process instead of crashing.
- `1a672b8c` — FLASH_ATTN backend casts non-fp16/bf16 inputs through bf16 with
  a one-time warning (safety net; TORCH_SDPA already tolerated fp32).
- `c1435a7f` — LTX2: torch 2.12 added `rms_norm` to autocast's float32 cast
  policy (`aten/src/ATen/autocast_mode.h`), so the q/k norms began returning
  fp32 into the autocast-blind flash-attention op and LTX2 died with SIGQUIT.
  `StageAwareRMSNorm` now restores the input dtype.
- `25d82762` — route grad-enabled attention to FA2: FA4 cute's backward
  asserts sm90+ (all L40S training lanes died at `loss.backward()`) and is
  unvalidated even on sm90, so training now uses FA2 on every device (the
  pre-FA4 behavior) while inference keeps FA4. Verified: the LoRA training
  test passes with train_loss matching its historical reference within 1e-3.

### SSIM reference reseed (L40S, `default` tier)

The torch 2.11+cu128 → 2.12+cu126 migration (including the `rms_norm`
autocast change) shifts numerics for few-step-distilled and single-image
models. Every model was validated against its existing refs on the new stack
first; only genuinely drifted models were reseeded, with the previous refs
backed up locally until CI is green.

| Reseeded model | before metric vs old ref |
|---|---|
| stabilityai__stable-diffusion-3.5-medium | SSIM 0.101 |
| HunyuanGameCraft-T2V | SSIM 0.139 |
| LongCat-Video-T2V | SSIM 0.200 |
| SFWan2.1-T2V-1.3B (causal) | SSIM 0.204 |
| Matrix-Game-2.0 | SSIM 0.309 |
| HunyuanGameCraft-I2V | SSIM 0.324 |
| Matrix-Game-3.0 | SSIM 0.429 |
| TurboWan2.1-T2V (SLA_ATTN) | SSIM 0.784 |
| LTX2-Distilled (latent) | cosine dist 0.800/0.816 |

Refs regenerated on Modal L40S via `fastvideo/tests/modal/ssim_test.py` with
CI-identical env (`IMAGE_VERSION=py3.12-latest`, `UV_TORCH_BACKEND=cu126`) at
`c3161945` (8 models) and `c1435a7f` (LTX2, includes its fix). Kept unchanged
after validating against existing refs: Wan2.1-T2V (0.978), Wan2.1-I2V
(0.990), LingBot (0.758), FLUX.2-klein-4B/9B, stable-audio, TurboWan2.2,
LongCat-I2V (0.969), LongCat-VC (0.986).

### Full-suite green sweep (post-reseed)

The remaining red lanes were root-caused from Buildkite logs and fixed:

- `119ff106` / `17aa7ef6` — eval-metrics: floor `ptlflow>=0.4` (unpinned, the
  in-container resolve backtracked it to 0.2.5, predating the `ckpt_path`
  API), unwrap transformers-5 CLAP `ModelOutput.pooler_output` in
  `clap_score`, and shim the removed `get_head_mask` in the vendored
  synchformer AST. Verified: 17 passed, 7 skipped (designed skips).
- `d5679133` — self-forcing full-suite lane 15m -> 30m: its budget was being
  consumed by the in-container kernel build (killed mid-nvcc, zero tests).
- `e1b12e71` — **pr_test lanes installed deps AFTER the kernel build**, so the
  pinned PyPI `fastvideo-kernel==0.2.6` wheel silently replaced the
  just-built in-tree kernel in every lane. This broke 3 of 4 kernel-test
  collection errors (post-0.2.6 symbols) and was the actual cause of the
  VSA training TMA/SIGABRT crash (0.2.6-era compiled kernels under
  torch 2.12/cu126). Install now runs before the build, mirroring
  ssim_test.py. Verified: VSA training passes.
- `e1b12e71` / `cf4b7280` — kernel tests: skip `test_attn_qat_infer`
  collection when `fp4attn_cuda` is absent (sm_120a-only build), and port
  the vsa256 CuTe fwd from the FA4 fork API (`m_block_size`/`n_block_size`)
  to the pinned upstream's `tile_mn` (+ 4-tuple return slice). Verified:
  210 passed, 2 skipped.

Follow-up worth a separate PR: bump the `fastvideo-kernel==0.2.6` pin to
the released 0.3.1 so PyPI installs of fastvideo get a kernel matching the
current python code (CI now tests the in-tree kernel either way).

Known follow-ups (out of scope here): `ssim_test.py` per-partition volume
sync overwrites sibling partitions' outputs; audit other
`nn.RMSNorm`-into-custom-op consumers (VSA/STA) for the same torch 2.12 fp32
exposure; orphaned refs (LongCat-T2V `TORCH_SDPA` mp4, LTX2 mp4) can be
deleted.


Satyam-53 (commented):
LGTM

Satyam-53 (approved):

Mister-Raggs:
Small addition for the install steps: `--prerelease=allow` (needed for the cu130 torch on aarch64) also pulls `transformers` 5.x over the repo's `transformers==4.57.3` pin. Wan models tolerate it, but Qwen2.5-VL-based models (e.g. Cosmos-Predict2.5) then fail at load with `AttributeError: 'Qwen2_5_VLConfig' object has no attribute 'pad_token_id'` (5.x nests it under `text_config`). Pinning it back after install fixes it:

```bash
uv pip install "transformers==4.57.3"
```

Verified on a GB10. Happy to push the doc line if helpful.

PR #1449, opened 2026-06-11
[feat]: add Hugging Face Kernel Hub packaging

### LINKED ISSUE #1318: [Feature] distribute fastvideo kernels through the `kernels` project
### Motivation

Would there be any interest in packaging, building, and distributing the FastVideo kernels using the [Kernels project](https://huggingface.co/docs/kernels/index)?

We recently worked with SGLang to build their fork of Flash Attention and provide an interface like the following:

```py
from kernels import get_kernel

fa3 = get_kernel("sgl-project/sgl-flash-attn3", version=1) 

fa3.flash_attn_varlen_func(q, k, v, cu_seqlens_q, cu_seqlens_k, causal=True)
fa3.flash_attn_with_kvcache(q, k_cache, v_cache, cache_seqlens=cache_seqlens, causal=True)
fa3.is_fa3_supported()  # True on H100/H200
```

The Kernels project is two-fold:

* Provide utilities to build kernels, enforcing structure and a support matrix. Kernels built this way become automatically compatible with the above `get_kernel()` API.
* The client-side utilities containing stuff like `get_kernel()`. 

I think the FastVideo kernels could benefit from this setup in the sense that more developers could easily integrate those kernels into their regular modeling code (where applicable). 

### Related resources

* https://huggingface.co/docs/kernels/index
* https://huggingface.co/kernels

aryan5v:
## Summary
- add Hugging Face kernel-builder metadata, flake, Hub card, and example for fastvideo-kernel
- add a torch-ext package exposing native FastVideo CUDA entrypoints through Kernel Hub-compatible Torch ops
- add a GitHub Action that builds and uploads hao-ai-lab/fastvideo-kernel via huggingface/kernel-builder-job
- document the Hub loading path and leave an exploration note for future SOP standardization

Closes #1318

## Testing
- git diff --check
- python3 -m py_compile fastvideo-kernel/torch-ext/fastvideo_kernel/__init__.py fastvideo-kernel/example.py
- python3 TOML parse for fastvideo-kernel/build.toml and fastvideo-kernel/pyproject.toml
- Ruby YAML parse for .github/workflows/publish-fastvideo-kernel-hub.yml
- .venv/bin/pre-commit run --files <changed paths>

## Notes
- Local full kernel-builder/Nix validation was not run because this workstation does not have nix or kernel-builder installed. The new workflow performs the remote HF Jobs build/upload path on main or manual dispatch.

### DIFF
--- a/.github/workflows/publish-fastvideo-kernel-hub.yml
+++ b/.github/workflows/publish-fastvideo-kernel-hub.yml
@@ -1,0 +1,57 @@
+name: Publish FastVideo Kernel to Hugging Face Hub
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/publish-fastvideo-kernel-hub.yml"
+      - "fastvideo-kernel/build.toml"
+      - "fastvideo-kernel/flake.nix"
+      - "fastvideo-kernel/CARD.md"
+      - "fastvideo-kernel/example.py"
+      - "fastvideo-kernel/torch-ext/**"
+      - "fastvideo-kernel/csrc/**"
+      - "fastvideo-kernel/include/**"
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Hub branch to upload the build"
+        default: ""
+        required: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build kernel via HF Jobs and upload to hao-ai-lab/fastvideo-kernel
+        uses: huggingface/kernel-builder-job@main
+        with:
+          token: ${{ secrets.HF_TOKEN }}
+          namespace: hao-ai-lab
+          flavor: cpu-xl
+          timeout: "21600"
+          script: |
+            set +x
+            export HF_TOKEN="${{ secrets.HF_TOKEN }}"
+            export GIT_LFS_SKIP_SMUDGE=1
+
+            git clone "${{ github.server_url }}/${{ github.repository }}" FastVideo
+            cd FastVideo
+            git checkout "${{ github.sha }}"
+            git submodule update --init --recursive fastvideo-kernel/include/cutlass fastvideo-kernel/include/tk
+            cd fastvideo-kernel
+
+            UPLOAD_BRANCH="${{ github.event.inputs.branch }}"
+            BRANCH_ARGS=()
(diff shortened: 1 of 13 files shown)


sayakpaul:
> Local full kernel-builder/Nix validation was not run because this workstation does not have nix or kernel-builder installed. The new workflow performs the remote HF Jobs build/upload path on main or manual dispatch.

Without this we cannot be sure if this is the right port. We need to see the built kernels and then verify if the yield expected results.

SolitaryThinker:
Maintainer review done. Good news first: the changes to the existing build are safe — the pybind guard macro is a no-op locally, the layernorm optional tweak is contained, and kernel CI compiled + tested the modified .cu files on H100. Runtime import path for fastvideo/ is untouched. A bunch of scarier-looking things were checked and dismissed (flake refs are correct post-monorepo-merge, build.toml keys are valid v4, sta_fwd param order matches the real definition).

What blocks merge:
1. **The actual deliverable has never been built.** @sayakpaul's objection stands — no kernel-builder/nix run has ever executed for this repo layout, and the workflow auto-publishes to `hao-ai-lab/fastvideo-kernel` on every main push touching `csrc/**` with no test or parity gate in between. First-ever build should not be a post-merge surprise: one validated run (workflow_dispatch to a scratch repo, or local nix) with a numerical spot-check vs `fastvideo_kernel._C` would settle it.
2. `secrets.HF_TOKEN` isn't provisioned in this repo (no other workflow uses it) — as-is the workflow fails on every qualifying merge.
3. Mechanical items for whenever the above is sorted: pin `kernel-builder-job` to a SHA (it receives the token), add a `concurrency:` group (parallel merges → racing 6h jobs), fix the torch op schemas (`output` args written in-place need `Tensor!?`, and `sta_fwd`/`gemm_cuda` return aliases undeclared in the schema — silent wrong results under torch.compile), and rebase (68 behind; the kernel tree moved to 0.3.x multi-arch, so the Hopper-only build.toml subset should at least say so on purpose).

There's also a policy question for the maintainers: whether Hub publishing should run ungated on every push or be version-gated like the PyPI flow — given the recent 0.3.0 release churn, gated seems safer. Raising that internally.

sayakpaul:
@SolitaryThinker I actually provided an update here: https://github.com/hao-ai-lab/FastVideo/issues/1318#issuecomment-4691359138. Could you perhaps review it?

PR #1453, opened 2026-06-12
[feat] Add opt-in dynamic generation batching

macthecadillac:
## Purpose

  Adds an opt-in dynamic batching path for compatible text-only generation requests. This lets prompt-file generation and the OpenAI-compatible video server coalesce compatible requests
  into a single forward pass while preserving the historical single-request path by default.

  Fixes # N/A

  ## Changes

  - Add typed batching config and CLI/API compatibility wiring for `batching_mode`, `batching_max_size`, `batching_delay_ms`, admission config, and batching metrics.
  - Add batching primitives for request compatibility signatures, conservative text-only admission, max batch caps, and model/resolution/cost-based JSON rules.
  - Add `VideoGenerator.generate_video_batch()` with merge/split behavior, per-request seed/output preservation, prompt-file dynamic batching, and sequential fallback for incompatible
  requests.
  - Make standard input-validation, text-encoding, and denoising stages batch-aware for the conservative text-only path.
  - Add an async OpenAI video `VideoBatchScheduler` that starts in server lifespan and routes eligible requests through dynamic batching.
  - Add focused unit tests, a Modal L40S parity/benchmark helper, SSIM validation notes, and final validation docs.
  - Added follow-up validation commit: `da55c18b [docs]: record SSIM validation follow-up`.
  - Added checklist closure commit: `95dc7d2b [docs]: record checklist closure (#1453)`.

  ## Test Plan

  ```bash
  pytest fastvideo/tests/batching \
    fastvideo/tests/api/test_compat_translation.py \
    fastvideo/tests/entrypoints/test_video_generator.py \
    fastvideo/tests/entrypoints/test_openai_api.py \
    fastvideo/tests/stages/test_input_validation_batching.py \
    fastvideo/tests/stages/test_text_encoding.py -q
```

  SSIM:

 ` FASTVIDEO_SSIM_MODEL_ID=Wan2.1-T2V-1.3B-Diffusers \
    pytest fastvideo/tests/ssim/test_wan_t2v_similarity.py -vs`

  Also ran changed-file pre-commit, full `pre-commit run --all-files`, and GPU validation through `fastvideo/tests/modal/launch_l40s_job.py` from the `interleavethinker` branch.

  ## Test Results
  <details>
  <summary>Focused validation</summary>

  Final changed-file suite on Modal L40S:
  - Modal app: ap-1mFqrE5eCwPkEKnffQcQou
  - Code validation commit: 2304837e
  - Result: 119 passed, 14 warnings
  - pre-commit: passed

  Focused suites:
  - batching + compat: 27 passed
  - generator batching: 51 passed
  - OpenAI scheduler: 61 passed
  - batch compat fix: 23 passed
  - text padding fix: 28 passed
  - single-text-encode fix: 29 passed
  </details>
  <details>
  <summary>SSIM validation</summary>

  H100 SSIM attempt:
  - Modal app: ap-KaJr2loSTefvmj8ijYwWOK
  - GPU: H100:2
  - Result: generated both Wan T2V videos, but failed before SSIM comparison
  - Reason: missing H100 reference folders for both FLASH_ATTN and TORCH_SDPA

  L40S SSIM fallback:
  - Modal app: ap-iWP6PA1IyZbXHDKtIE1LQH
  - GPU: L40S:2
  - Command:
    FASTVIDEO_SSIM_MODEL_ID=Wan2.1-T2V-1.3B-Diffusers pytest fastvideo/tests/ssim/test_wan_t2v_similarity.py -vs
  - Result: 2 passed, 6 warnings
  - FLASH_ATTN mean SSIM: 0.9786614696
  - TORCH_SDPA mean SSIM: 0.9743387236
  </details>
  <details>
  <summary>Pre-commit and checklist validation</summary>

  Changed-file pre-commit:
  - Passed in prior validation.

  Final full pre-commit at PR head:
  - Modal app: ap-XLXARFbxkJa40Yom5VxxfX
  - Commit: 71fe451b
  - Command: pre-commit run --all-files
  - Result: passed yapf, ruff, codespell, PyMarkdown, actionlint, mypy, filename spaces, and suggestion hooks.

  Final docs-patch validation:
  - Modal app: ap-9zeeQDxLjSzuxlYss4dIL8
  - Commit: 71fe451b plus local checklist docs patch, later committed as 95dc7d2b
  - Command: pre-commit run --all-files
  - Result: passed yapf, ruff, codespell, PyMarkdown, actionlint, mypy, filename spaces, and suggestion hooks.

  Support matrix:
  - Reviewed docs/inference/support_matrix.md.
  - No update is required because this PR adds opt-in batching behavior and does not add a model ID or registry entry.
  </details>

  ## Known Limitations

  Dynamic batching remains disabled by default and is intentionally limited to compatible text-only requests. Image, video, audio, refine, continuation, and other conditioning-heavy
  requests fall back to sequential execution.

  Wan latent parity completed successfully but is not near bit-identical to sequential denoising in the current implementation:

  - aggregate max abs diff: 0.1457520127
  - aggregate mean abs diff: 0.0122398026
  - allclose(atol=1e-4, rtol=1e-4): false

  Benchmarks on small latent-output Wan2.1 T2V 1.3B workloads showed:

  - batch 2, 2 steps: +1.6%
  - batch 4, 2 steps: +0.1%
  - batch 2, 8 steps: +10.7%

  ## Checklist

  - [x] I verified targeted Wan T2V SSIM regression tests pass on L40S
  - [x] I updated the support matrix if adding a new model (N/A: no model ID or registry entry added; support matrix reviewed)


SolitaryThinker:
Maintainer update: rebased onto current main (was 89 behind — clean, and I checked the batch-size-1 assert deletion still makes sense in the reworked denoising stage; it does, the loop is batch-aware now) and pushed 6 fix commits on top. Head is 82b6d2db.

What I found and fixed:
- duplicate prompts in a batch silently overwrote each other's output files — paths were all resolved before any file was written, so the exists-dedupe never fired. Paths are now reserved at resolution time, suffixed like the sequential path.
- with the default `batching_delay_ms=0`, the server scheduler was a silent no-op — `wait_for(timeout=0)` timed out before draining the queue, so every batch was size 1. delay=0 now drains greedily.
- `submit()` after `stop()` hung the caller forever; it now fails fast.
- one failed group in a prompt-file batch discarded all completed results; failures now produce per-prompt error entries like the legacy loop.
- `fastvideo/tests/batching/` wasn't collected by any CI lane, so the green build never ran your new tests; added it to the unit lane.
- dropped the 3 `.agents/exploration` journal files (repo policy — no experiment journals under .agents/).

Validation: Modal L40S run at the new head — 467 passed (batching, api, entrypoints suites), including your scheduler and work-item merge tests, now actually executing.

Deliberately NOT touched, flagged for discussion: (1) `_generate_single_video` is duplicated ~350 lines into the new helpers — future save/mux fixes would need to land twice; (2) batching currently only serves the legacy kwargs surface — typed `generate(request)` prompt lists still expand to sequential singles. Both feel like they should be resolved before or right after merge.

Seed handling, conditioning deny-lists, and batch-boundary state were all reviewed and came back clean — nice work on those.

SolitaryThinker:
Rebased onto current main to pick up the pre-commit environment fix (#1572) that was failing the lint lane with exit 127 on pre-#1572 branches; no content changes (range-diff clean, all 31 commits =).

SolitaryThinker (changes requested):
# Review: dynamic generation batching

**Overall**: the design is genuinely conservative and well-structured — identity-based signature freezing for unhashable values, explicit per-pipeline capability opt-in with a DMD/causal veto, per-request seed/generator preservation (verified end-to-end: `batch.seeds` → per-seed CPU generators → per-element `randn`), and a scheduler with careful crash-drain logic. But there are two high-severity correctness regressions and a cluster of failure-isolation problems that should be fixed before merge. The PR is also currently conflicting with main and needs a rebase.

Every finding below was independently verified against the PR head (`b8abf6f`) before being included; several plausible-sounding issues (e.g. "Wan variants inherit `supports_dynamic_batching` unsafely") were checked and found *not* to be real, and are excluded.

## High severity

### 1. Deleted batch-size-1 assert turns `num_videos_per_prompt>1` into silent output corruption

`fastvideo/pipelines/stages/denoising.py:250` — the PR removes `assert latent_model_input.shape[0] == 1` with no replacement shape guard. With `num_videos_per_prompt=2` (a plain user parameter, reachable with batching **disabled**), latents have batch dim 2 but prompt embeds stay `[1, 512, D]`; Wan cross-attention does `.view(b, -1, n, d)`, which silently reshapes the single context into `[2, 256, n, d]` — video 0 attends to half the prompt tokens, video 1 attends mostly to zero padding. Pre-PR this crashed with the assert; post-PR it produces garbage with no warning.

**Fix**: assert `latents.shape[0] == prompt_embeds[0].shape[0]` in the stage, or reject `num_videos_per_prompt > 1` upstream.

### 2. Batched output ≠ sequential output (0.146 max abs latent divergence) silently breaks seed determinism

Traced by elimination: noise (per-request generators), text embeddings (Wan's `t5_postprocess_text` already fixes every prompt to `[1,512,D]`, so `_cat_tensors` takes the plain-concat path), timesteps, and scheduler state are all bitwise identical to solo runs — the divergence comes from batch-size-dependent kernel numerics under bf16, amplified over denoising steps. That's inherent to batching, but it means enabling it makes the same seeded request return different videos depending on server load. It's disclosed in the PR body but not in user-facing docs, and nothing bounds it (see 3).

### 3. No automated parity gate

`fastvideo/tests/batching/run_dynamic_batching_parity.py` computes allclose/diff metrics but never asserts and exits 0 regardless; it has no `test_` prefix and zero references from CI or `pr_test.py`. Every CI batching test stubs `generate_video_batch`/`_run_forward_batch`, so the real numerical path (merge → forward → split) is executed by **zero** tests. A future regression that turns batched output into garbage keeps CI green.

**Fix**: convert the parity script into an asserting gate (even a loose threshold) so the current 0.146 divergence can't silently grow.

## Medium severity

### 4. `prompt_path` batching is a silent no-op

`fastvideo/entrypoints/video_generator.py:666` — the per-item `sampling_param` retains `prompt_path`, which `_request_is_batchable` rejects (`fastvideo/batching/signature.py:121`), so every group has size 1. The feature only works via `--prompt-txt`. **Fix**: clear `prompt_path` on the per-item params (it's already consumed). Untested because tests only exercise the `prompt_txt` route.

### 5. One failure in a merged group kills every innocent request in it

- Prompt-file mode (`video_generator.py:1183`): a merged-group failure returns error entries for *all* items with no per-item retry — a regression vs. the old per-prompt try/except isolation, and more likely to trigger since a batch-of-4 forward can OOM where batch-of-1 succeeds (there is no VRAM-derived cap unless a `--batching-config` is supplied).
- Server mode (`fastvideo/entrypoints/openai/batching.py:177`): `_dispatch` fans one request's per-request failure (e.g. an unwritable `output_path`, which is deliberately excluded from the signature) out to every co-batched client as the same error.

**Fix**: fall back to per-item sequential execution when a merged group fails; covers both variants.

### 6. `--batching-config` is validated lazily and re-read from disk on every batch

A typo'd path passes startup (the capability flag *is* checked at startup — inconsistently), the server reports healthy, then 100% of generations fail at dispatch. In prompt-file mode the `BatchAdmissionController` is constructed outside the `tolerate_failures` try, so a bad config aborts the whole run. **Fix**: validate once at startup, cache the parsed rules.

### 7. Merge/split tests never assert tensor values

`fastvideo/tests/entrypoints/test_video_generator.py:210` fakes the forward with a value-distinguishable `arange` and then only checks *shapes*. An indexing bug in `_split_output_batch` that hands every client request 0's video would pass the entire suite — which matters doubly because its heuristic (any list of `len == batch_size` gets index-split) is fragile. One-line fix per test: assert the slice values.

### 8. `prompt_txt` batching branch and `tolerate_failures` machinery have zero test coverage

`grep -rn tolerate_failures fastvideo/tests/` → nothing, yet it's the path every `--prompt-txt` CLI user hits, and it ends in a `zip(..., strict=True)` that would crash mid-run if group error accounting ever drifts.

## Low severity

9. `_config_entries` silently reinterprets `{"rules": {...}}` (object instead of array) as a model-keyed mapping, yielding a never-matching rule — the user's cap is dropped without error (`fastvideo/batching/admission.py:242`).
10. `calibration` is an accepted rule key that no code reads — validates, then silently does nothing (`admission.py:26`).
11. The refactor made the forward-thread pointless: `_run_forward_batch` now does `thread.start(); thread.join()` back-to-back, and the pinned ~50MB host buffer that used to allocate *concurrently* with the forward is now allocated after it — a small serial latency add on **every** generation, including the default non-batched path (`video_generator.py:909`).
12. If the scheduler task previously died, `lifespan` shutdown re-raises from `stop()`'s `await self._task` and skips `generator.shutdown()`/`clear_state()` (`fastvideo/entrypoints/openai/api_server.py:85`).
13. `batching_mode="dynamic"` with the default `batching_max_size=1` is a silent no-op on every surface — worth a startup warning.
14. `test_admission.py` covers only methods the production merge path never calls; the one it does call (`reject_reason_for_candidate`) is untested.

## Advisory (quality/perf, no verification pass)

- `_sampling_param_from_kwargs` calls `SamplingParam.from_pretrained` (disk I/O) twice per compatibility probe **on the event loop**, re-probing the base each time (`openai/batching.py:212`) — cache it.
- Over-built for day one: the config file accepts three input formats plus fuzzy-key suggestions; `BatchAdmissionController` ships four public methods used only by tests; `enable_batching_metrics` is six surfaces of plumbing for one `logger.info`; `dynamic_batch_size`/`dynamic_batch_output_paths` extras are written but never read; `_merge_work_items` has a no-op seed assignment (`video_generator.py:1112`); scheduler test scaffolding is duplicated verbatim across two test directories.
- The `padding=True` injection in `encode_text` (`fastvideo/pipelines/stages/text_encoding.py:350`) is dead for this feature (the list path encodes prompts individually) but silently changes behavior for any pre-existing caller passing multiple preprocessed texts — remove or scope it.
- Grouping is adjacent-only, so one incompatible request between compatible ones splits a batch — fine as a FIFO-fairness choice, but worth a code comment.

## Requested before merge

(a) restore the shape guard or reject `num_videos_per_prompt>1`; (b) fix or document `prompt_path` batching; (c) validate `batching_config` at startup; (d) per-item fallback on merged-group failure; (e) value assertions in the merge/split tests; (f) an asserting parity gate; (g) determinism caveat in user-facing docs, not just the PR body; (h) rebase onto main.

macthecadillac:
Thanks for the detailed review. I worked through the requested items and pushed the fixes; the current head is `bc3bda6b`.

1. Restored an explicit batch-dimension guard between latents and prompt embeddings, preventing `num_videos_per_prompt > 1` from silently corrupting attention inputs.
2. `prompt_path` is now cleared after expansion into per-item requests, with coverage for both prompt-file routes.
3. Batching configuration is validated and loaded once during startup, rather than lazily re-read during dispatch.
4. Merged-forward failures now fall back to individual execution. Postprocessing failures are isolated per item, and the API records the actual successfully published output path.
5. Merge/split tests now verify tensor values and request mapping, not only shapes. The `prompt_txt` and `tolerate_failures` paths also have direct coverage.
6. Added an asserting parity gate to both Fastcheck and the full suite. It exercises a real merged forward on two L40S GPUs with SP2, CFG, negative prompting, pixel decoding, and media publication. The gate checks mapping, shapes, finiteness, per-request mean error, and a calibrated worst-0.1% tail metric. Raw maximum error remains a diagnostic because isolated bf16 extremes were not stable enough to use as a reliable gate.
7. User documentation now explicitly states that dynamic batching preserves request seeds but does not promise bitwise identity with sequential execution. The recurring gate bounds that numerical difference instead of leaving it unmonitored.
8. The branch was rebased. Although `main` has advanced again, `git merge-tree` against the current head of `main` is clean and GitHub reports the PR as mergeable.

I also addressed the lower-severity items: malformed `rules` objects and unused `calibration` fields are rejected, forward/output-buffer overlap was restored, scheduler failures cannot skip generator cleanup, dynamic mode with max size 1 emits a warning, the production admission path is tested, and batching metrics now report utilization and rejection reasons. Atomic output publication and permission preservation were also hardened during follow-up review.

Final two-L40S validation produced per-request mean differences of `0.012266` and `0.018214`, with worst-0.1% tail means of `0.173599` and `0.263686`, all within the configured `0.02`/`0.30` bounds. Both hosted dynamic-batching parity lanes, pre-commit, unit, API, LoRA, and performance checks pass.

Two final-head training jobs exited with status `124`; the available GitHub check output does not expose enough detail to classify those timeouts. The final-head SSIM lane is still pending.

I also ran the independent review/adjudication cycle against the final commit; the last review found no remaining actionable issues.

PR #1457, opened 2026-06-12
[feat] QAD 5090: Wire the Attn-QAT inference attention backend (8/12)

alexzms:
## Purpose

Continues the Attn-QAT upstreaming stack (#1225). Wires in the
`AttnQatInferBackend` that #1358 landed as **deadcode**, now that its modified
SageAttention3 FP4 kernels land in #1455. This makes the 4-bit attention path
selectable end-to-end for Wan-2.1 inference.

## Changes

- `platforms/interface.py`: add `ATTN_QAT_INFER` to `AttentionBackendEnum`. The
  name matches `AttnQatInferBackend.get_name()`, so `backend_name_to_enum()`
  resolves the `FASTVIDEO_ATTENTION_BACKEND=ATTN_QAT_INFER` override.
- `platforms/cuda.py`: dispatch `ATTN_QAT_INFER` → `AttnQatInferBackend`, guarded
  by `is_attn_qat_infer_available()`; falls back to Flash Attention when the
  kernel is not built (non-Blackwell, or before #1455 merges).
- `configs/models/dits/base.py`: add `ATTN_QAT_INFER` to the default
  `_supported_attention_backends` so Wan can select it.

The training-side `ATTN_QAT_TRAIN` backend depends on a separate Triton kernel
and is deferred to the training upstream.

## Test Plan / Test Results

- ✅ `name ↔ enum` closed: `AttnQatInferBackend.get_name() == "ATTN_QAT_INFER"`
  is a valid `AttentionBackendEnum` member.
- ✅ **Dispatch verified on a Blackwell node** (using the #1455-built kernel):
  - kernel present → `get_attn_backend_cls(ATTN_QAT_INFER)` returns
    `fastvideo.attention.backends.attn_qat_infer.AttnQatInferBackend`
  - kernel absent → logs the fallback and returns `FlashAttentionBackend`
- ✅ pre-commit (yapf / ruff / mypy / codespell) passes.

### Notes
- **Depends on #1455** (the kernel). Until that merges, `is_attn_qat_infer_available()`
  returns False on `main`, so this safely no-ops to Flash Attention.
- Full runtime/numeric parity is `sm_120` (RTX 5090)–specific by kernel design.

Part of #1225.

PR #1358, opened 2026-05-16
[feat]: Attn-QAT inference + training backends (deadcode) (Attn-QAT 4/12)

SolitaryThinker:
Re-opens replacement for #1356 — original PR was auto-closed when its base branch `pr1225_s3` was deleted upstream after slice 3 (#1350) squash-merged.

Slice 4 of 12 in the PR #1225 decomposition. Deadcode — adds the two QAT attention backends but does NOT register them in the selector. Activation happens at slice 12.

## What this adds

- `fastvideo/attention/backends/attn_qat_infer.py` (+121)
- `fastvideo/attention/backends/attn_qat_train.py` (+145)

Both files extracted from PR #1225 (`3f818d0fc532ec6494b465967d5f485150917d0c`), audited for newest-API compat against `origin/main`, and retargeted from `fastvideo/training/` to `fastvideo/train/` where applicable.

## Review feedback from #1356 addressed in this PR

Six gemini-code-assist inline comments on #1356 are addressed here (commits squashed into the slice-4 commit):

- `attn_qat_train` accepts optional `sm_scale` parameter; falls back to `1/sqrt(d)` when `None`.
- `attn_qat_infer` passes `sm_scale=self.softmax_scale` to the kernel.
- `AttnQatTrainImpl.forward()` plumbs `self.softmax_scale` through to `attn_qat_train`.
- Both `AttnQatInferImpl` and `AttnQatTrainImpl` raise `NotImplementedError` if `dropout_p > 0` (kernels do not support dropout — surface the limitation instead of silent misbehavior).

## Out of scope (in later slices)

- Selector registration → slice 12 (activation)
- Shared attention infra additions for QAT-compat → slice 5
- Linear/MLP FP4 path additions → slice 6

Attn-QAT-Stack: 4/12

Co-Authored-By: jzhang38 <42993249+jzhang38@users.noreply.github.com>
Co-Authored-By: RandNMR73 <99706358+RandNMR73@users.noreply.github.com>

Closes #1356

PR #1225, opened 2026-04-09
[feat] Upstream Attn-QAT Video Diffusion Code

RandNMR73:
## Summary

- Add Attn-QAT training kernels
- Add flashinfer NVFP4 linear layers (currently hardcoded for Wan-2.1 arch)
- Add Modified SageAttention3 kernels in the fastvideo-kernel package
- Add Attn-QAT video model training scripts

PR #1459, opened 2026-06-12
[feat] QAD 5090: Wire the Attn-QAT training attention backend (10/12)

alexzms:
## Purpose

Continues the Attn-QAT upstreaming stack (#1225). First training-side slice:
makes the `AttnQatTrainBackend` (landed as deadcode in #1358) selectable for
**quantization-aware finetuning**, mirroring the inference wiring of #1457. The
fake-quant-in-backward attention path becomes reachable via
`FASTVIDEO_ATTENTION_BACKEND=ATTN_QAT_TRAIN`.

This is **config-driven** and intentionally avoids the research fork's
monkey-patch module-swapping (`traverse_swap_module` / `swap_*`): main already
selects the attention backend from config in both training and inference, so no
training-pipeline change is needed.

## Changes

- `platforms/interface.py`: add `ATTN_QAT_TRAIN` to `AttentionBackendEnum`.
- `platforms/cuda.py`: dispatch `ATTN_QAT_TRAIN` → `AttnQatTrainBackend`, guarded
  by `is_attn_qat_train_available()`.
- `attention/backends/attn_qat_train.py`: add `is_attn_qat_train_available()`
  helper (mirrors `is_attn_qat_infer_available()`).
- `configs/models/dits/base.py`: add `ATTN_QAT_TRAIN` to the default supported set.

## Test Plan / Test Results

- ✅ `name ↔ enum` closed: `AttnQatTrainBackend.get_name() == "ATTN_QAT_TRAIN"` is
  a valid enum member.
- ✅ Dispatch verified: `get_attn_backend_cls(ATTN_QAT_TRAIN)` returns the backend
  when the kernel is available; otherwise warns and falls back to Flash Attention.
- ✅ pre-commit (yapf / ruff / mypy / codespell) passes.

### Notes
- The training Triton kernel `fastvideo_kernel.triton_kernels.attn_qat_train` is
  **not upstreamed yet**, so this is deadcode for now (selecting it falls back to
  Flash with a warning). A follow-up will land the kernel.
- **Overlaps #1457** on `interface.py` / `cuda.py` / `base.py` (adjacent enum /
  dispatch / supported-list lines). Whichever merges second needs a trivial
  rebase.
- Open design question: for training, should a missing kernel **hard-error**
  instead of silently falling back (to avoid "fake QAT")? Currently warn+fallback
  to match inference.

Part of #1225.

### DIFF
--- a/fastvideo/attention/backends/attn_qat_train.py
+++ b/fastvideo/attention/backends/attn_qat_train.py
@@ -49,6 +49,10 @@ def _get_attn_qat_train_attention() -> Callable[..., torch.Tensor] | None:
     return _attn_qat_train_attention
 
 
+def is_attn_qat_train_available() -> bool:
+    return _get_attn_qat_train_attention() is not None
+
+
 def attn_qat_train(q_BLHD: torch.Tensor,
                    k_BLHD: torch.Tensor,
                    v_BLHD: torch.Tensor,
--- a/fastvideo/configs/models/dits/base.py
+++ b/fastvideo/configs/models/dits/base.py
@@ -24,7 +24,8 @@ class DiTArchConfig(ArchConfig):
                                                  AttentionBackendEnum.TORCH_SDPA,
                                                  AttentionBackendEnum.VIDEO_SPARSE_ATTN,
                                                  AttentionBackendEnum.VMOBA_ATTN, AttentionBackendEnum.SAGE_ATTN_THREE,
-                                                 AttentionBackendEnum.ATTN_QAT_INFER, AttentionBackendEnum.SLA_ATTN,
+                                                 AttentionBackendEnum.ATTN_QAT_INFER,
+                                                 AttentionBackendEnum.ATTN_QAT_TRAIN, AttentionBackendEnum.SLA_ATTN,
                                                  AttentionBackendEnum.SAGE_SLA_ATTN)
 
     hidden_size: int = 0
--- a/fastvideo/platforms/cuda.py
+++ b/fastvideo/platforms/cuda.py
@@ -153,8 +153,10 @@ class CudaPlatformBase(Platform):
             if is_attn_qat_train_available():
                 logger.info("Using Attn-QAT training (fake-quantized attention) backend.")
                 return "fastvideo.attention.backends.attn_qat_train.AttnQatTrainBackend"
-            logger.warning("Attn-QAT training kernel is not built; falling back to Flash Attention "
-                           "(NO fake-quant in the attention path).")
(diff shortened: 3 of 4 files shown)


SolitaryThinker:

## TL;DR

I recommend changing the training-side missing-kernel path to hard-error rather than warning and silently falling back, because otherwise a user can think they are QAT-training while actually training the attention path without fake quantization. Separately, this PR currently needs a mechanical rebase against current `main` after #1457 merged; that is a pre-merge metadata gate, not a code-quality concern.

## Verdict: **request-changes**

- S0 (blockers): 1 pre-merge rebase gate
- S1 (must-fix): 0
- S2 (should-fix): 1
- S3 (discussion): not shown here; see review.md

---

## Pre-merge gate: rebase required

PR is currently CONFLICTING with main (the `needs-rebase` label is applied). After #1457 merged at 20:12Z, `interface.py`, `cuda.py`, and `base.py` all gained the `ATTN_QAT_INFER` lines this PR also touches. Rebase is mechanical (adjacent enum/list additions, no semantic conflict) but blocks merge.

## Prior gemini-code-assist concerns (status at `4e8ee637`)

| Prior concern | Status at HEAD | Severity now |
|---|---|---|
| Hard-error vs warn+fallback for missing training kernel | ⏸️ PERSISTENT (same question, my recommendation: hard-error) | S2 |

### [S2] Training-side fallback should hard-error, not warn+silent-fallback

**What:** `CudaPlatform.get_attn_backend_cls()` accepts `FASTVIDEO_ATTENTION_BACKEND=ATTN_QAT_TRAIN`, checks `is_attn_qat_train_available()`, and when the Triton kernel is missing logs a warning before dropping through to the normal Flash/SDPA fallback path.

**Why it matters:** Silent fallback in training means the user thinks they're doing QAT but is actually full-precision-training the attention path. The model weights get shaped by mismatched distributions; the resulting "QAT" checkpoint is a lie — at inference time it gets quantized into a state it was never trained for. Inference (#1457) can defensibly warn+fallback because the user gets a quality regression, not a fundamentally different training run.

**Suggested fix:** For the `ATTN_QAT_TRAIN` branch only, raise an `ImportError` or `RuntimeError` when `is_attn_qat_train_available()` is false. Keep the inference branch's warn+fallback behavior unchanged if that remains the intended inference UX.

**Evidence:** `fastvideo/platforms/cuda.py:143-150`

---

*— Gob (@SolitaryThinker's AI reviewer). Full review (including S3 items + verification log) is archived locally.*

SolitaryThinker (approved):

PR #1463, opened 2026-06-16
[feat] QAD 5090: FP4 QAT linear STE for training (13/12)

alexzms:
## Purpose

Continues the QAD 5090 stack (#1225). Wires the **FP4 linear straight-through
estimator** into training, so the DiT linear layers are quantized during QAT
(forward FP4, backward full precision) — "4-bit everything" training alongside
the Attn-QAT attention path. This was the remaining gap: finetune/distill so far
ran linear in bf16.

## Changes

- `layers/quantization/nvfp4_qat_train_config.py`: a training quant method that
  bridges the otherwise-orphaned `_LinearFWD4BWD16Fn` STE (`fp4linear.py`) into
  the standard `quant_config` path. The weight stays a **trainable** bf16/fp32
  master, fake-quantized to FP4 each forward (no `convert_model_to_fp4`). Matches
  Wan's `to_q/k/v/out` + `ffn` layers; registered as `nvfp4_qat_train`.
- `fastvideo_args.py`: add a `--transformer-quant` training CLI arg, and resolve a
  registry-name **string** to a `QuantizationConfig` instance in
  `_apply_transformer_quant` (a bare string has no `get_quant_method`).

## Test Results (Blackwell GB200 / sm_100)

- ✅ **Unit**: a `nvfp4_qat_train` `ReplicatedLinear` runs forward in FP4 and
  backward produces a **nonzero `weight.grad`** (+ `x.grad`) — the STE works.
- ✅ **Integration**: a finetune smoke with `ATTN_QAT_TRAIN` attention +
  `--transformer-quant nvfp4_qat_train` linear trains 10 steps with healthy
  loss/grad (loss 0.45 → 0.04, grad_norm ~0.25–3.6) — 4-bit attention and 4-bit
  linear together.

## Notes
- Uses `ATTN_QAT_TRAIN`, so depends on #1459 (wiring) + #1460 (kernel).
- With this, the recipe scripts in #1462 can add `--transformer-quant
  nvfp4_qat_train` to quantize linear during finetune/distill.

Part of #1225.


SolitaryThinker (commented):

## Verdict: COMMENT

The STE and wiring are **correct** — I traced the gradient path and the integration end-to-end. Forward genuinely FP4-fake-quants; `_LinearFWD4BWD16Fn.backward` (`fastvideo/layers/fp4linear.py:92-108`) passes full-precision identity gradients through the saved bf16/fp32 master (`grad_x = grad_y·W`, `grad_w = grad_yᵀ·x`), and `create_weights` sets `requires_grad=True` so the master trains. The config/method satisfy the `QuantizationConfig`/`QuantizeMethodBase` ABCs, the `create_weights`/`apply` signatures match `LinearBase` (`linear.py:258-266`, `:296`), registry + `Literal` + CLI are all wired, and the loader hooks (`_maybe_convert_model_to_nvfp4`, `convert_model_to_fp4`) correctly leave the trainable master untouched. The `isinstance(tq, str)` branch in `_apply_transformer_quant` is backward-compatible — the existing typed/instance path is unchanged, so `tests/api/test_typed_quant_flow.py` still holds.

Not approving outright for one reason:

### Major
- **No committed test for the STE.** `fastvideo/layers/quantization/nvfp4_qat_train_config.py` — the PR body claims a unit test (nonzero `weight.grad`) and a 10-step integration smoke, but the diff adds no test file and there's no existing test for `_LinearFWD4BWD16Fn` / this method anywhere in the tree. The STE is the whole correctness contract here and it's sm_100-gated (`get_min_capability` → 100), so neither `fastcheck` nor `full-suite` exercises it — the manual Blackwell result isn't reproducible from the tree. Please add an sm_100-guarded unit test asserting forward runs and `weight.grad`/`x.grad` are non-None and finite after a backward, so a future `fp4linear.py` refactor can't silently break the gradient path. (MAJOR not BLOCKER: the logic is verified-correct by reading and the path is opt-in/dormant without `--transformer-quant`.)

### Minor
- `fastvideo/fastvideo_args.py:278-281` — the resolved instance lives in local `tq`, not written back to `self.transformer_quant`. Harmless today (called once from `__post_init__`), but `self.transformer_quant = tq` after resolution would make the attribute self-consistent and avoid a re-resolve if ever called twice.
- `fastvideo/layers/quantization/nvfp4_qat_train_config.py:46-49` — the `fp4_layers` list + `get_quant_method` body are duplicated verbatim from `nvfp4_qat_config.py:114-117`. Consider a shared helper so the inference and train methods can't drift on which layers they quantize.

### Notes
- `pre-commit` is red on **yapf only** (`mypy`/`ruff`/`codespell` pass) — style; run `pre-commit run -a` to clear the required check.
- `buildkite/ci/microscope-transformer-tests` is red while the `pr-fastcheck` mirror of the same job and the required `fastcheck-passed` gate are green; the new path is opt-in/dormant without the CLI flag, so this looks like a pre-existing/env CI-leg flake rather than a regression here — but I couldn't open the (private) buildkite build to confirm. Suggest `/test training` or a targeted sm_100 run before merge.
- Dependency check: `_LinearFWD4BWD16Fn` (#1460) + the quant base/registry are on `main`; PR is 6 ahead / 1 behind, MERGEABLE — no unmerged-part blocker.

SolitaryThinker (approved):

PR #1464, opened 2026-06-16
[feat] QAD 5090: FP8 QAT linear training (14/12)

alexzms:
## Purpose

Continues the QAD 5090 stack (#1225). Adds **FP8 (e4m3) quantization-aware
training** for the DiT linear layers — the FP8 sibling of the FP4 linear STE
(#1463), for FP8-class GPUs (the blog's `XXX-FP8-1.3B` path).

## Changes (a couple of files, mirroring #1463)

- `layers/fp8linear.py`: `_LinearFWD8BWD16Fn` — FP8 forward (`torch._scaled_mm`
  on sm89+, **bf16 fake-quant fallback on older GPUs**) + full-precision backward
  (straight-through estimator). Absmax tensor/row-wise scaling, `FP8_MAX=448`.
  Adapted from the fork's FP8 training commit (96011d29) to the config path.
- `layers/quantization/fp8_qat_train_config.py`: a training quant method that
  keeps a **trainable** bf16 master and applies the STE; registered as
  `fp8_qat_train` (matches Wan `to_q/k/v/out` + `ffn`).
- register in `layers/quantization/__init__.py`.

No `flashinfer` and no monkey-patch flags. Enable with `--transformer-quant
fp8_qat_train` (the CLI arg + string→config resolve from #1463). Runs on any
sm89+ GPU (and older via the fallback), not just Blackwell.

## Test Results (Blackwell GB200 / sm_100)

- ✅ **Unit**: an `fp8_qat_train` `ReplicatedLinear` runs forward via
  `torch._scaled_mm` (FP8) and backward gives a **nonzero `weight.grad`** — STE works.
- ✅ **Integration**: a finetune smoke with `--transformer-quant fp8_qat_train`
  trains 10 steps with healthy decreasing loss (0.11 → 0.076) and grad (~0.4)
  — and faster steps than FP4 (no flashinfer JIT).

Depends on #1463 (the `--transformer-quant` CLI arg). Part of #1225.

### DIFF
--- a/fastvideo/layers/fp8linear.py
+++ b/fastvideo/layers/fp8linear.py
@@ -1,0 +1,135 @@
+# SPDX-License-Identifier: Apache-2.0
+"""FP8 quantization-aware training for linear layers.
+
+Mirror of ``fp4linear.py`` but for FP8 (e4m3). The forward pass quantizes both
+activations and weights to FP8 and runs ``torch._scaled_mm``; the backward pass
+is a bf16 straight-through estimator so the high-precision master weights stay
+trainable. Falls back to a bf16 fake-quant forward on GPUs older than sm89.
+"""
+import torch
+
+FP8_DTYPE = torch.float8_e4m3fn
+FP8_MAX = float(torch.finfo(FP8_DTYPE).max)  # 448.0
+FP8_MIN_SCALE = 1.0 / (FP8_MAX * 512.0)
+
+
+def _supports_fp8_compute() -> bool:
+    """Whether the active device supports FP8 ``_scaled_mm`` (sm89+)."""
+    if not torch.cuda.is_available():
+        return False
+    cap = torch.cuda.get_device_capability()
+    return cap[0] > 8 or (cap[0] == 8 and cap[1] >= 9)
+
+
+def _quantize_tensorwise(
+    x_2d: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Returns ``(x_fp8 [M, K], x_scale [1] float32)``."""
+    x_absmax = x_2d.abs().amax().float()
+    x_scale = (x_absmax / FP8_MAX).clamp(min=FP8_MIN_SCALE)
+    x_fp8 = (x_2d / x_scale.to(x_2d.dtype)).clamp(-FP8_MAX, FP8_MAX).to(FP8_DTYPE)
+    return x_fp8, x_scale.view(1)
+
+
+def _quantize_rowwise(
+    x_2d: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Returns ``(x_fp8 [M, K], x_scale [M, 1] float32)``."""
+    x_absmax = x_2d.abs().amax(dim=-1, keepdim=True).float()
+    x_scale = (x_absmax / FP8_MAX).clamp(min=FP8_MIN_SCALE)
+    x_fp8 = (x_2d / x_scale.to(x_2d.dtype)).clamp(-FP8_MAX, FP8_MAX).to(FP8_DTYPE)
+    return x_fp8, x_scale
+
+
+def _fake_quant(x_2d: torch.Tensor, granularity: str) -> torch.Tensor:
+    """bf16 fake-quant (quantize then dequantize) for pre-sm89 fallback."""
+    if granularity == "channel":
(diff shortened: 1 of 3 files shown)


SolitaryThinker (commented):

## Summary
FP8 (e4m3) sibling of the FP4 QAT linear STE (#1463): a new `fp8linear.py` (`_LinearFWD8BWD16Fn` — FP8 forward via `torch._scaled_mm` on sm89+, bf16 fake-quant fallback below sm89, full-precision STE backward) + `fp8_qat_train_config.py` (trainable bf16/fp32 master bridged into the `quant_config` path as `fp8_qat_train`) + the registry wiring. The STE math and the autograd/registry/contract wiring are **correct** (verified by reading + tracing). The rebase onto current `main` is clean — `MERGEABLE`, registry conflict resolved. One substantive ask remains: like #1463, no committed STE test.

**Verdict: COMMENT** — logic is correct and the rebase is clean; the path is opt-in/dormant without `--transformer-quant fp8_qat_train`, so nothing lands broken. No blockers.

## Major
- **No committed test for the FP8 STE** (`fastvideo/layers/fp8linear.py`). The body claims a unit test (forward via `_scaled_mm` + nonzero `weight.grad`) and a 10-step finetune smoke (loss 0.11→0.076, grad ~0.4 on GB200/sm_100), but the diff adds 3 non-test files and zero tests — nothing under `fastvideo/tests/` or `tests/` references `fp8_qat`/`fp8linear`/`_LinearFWD8BWD16Fn`. The STE is the whole correctness contract here, and the active `_scaled_mm` path is sm89-gated, so the claimed result isn't reproducible from the tree (same gap as #1463). Please add a small unit test asserting (a) forward runs and (b) `weight.grad is not None` and nonzero after a backward — routed through the bf16 **fallback** path (see the device fix below) so it runs on CPU / older GPUs in CI rather than Blackwell-only.

## Minor
- `fp8linear.py:16-21,69` — `_supports_fp8_compute()` ignores `x.device`: it keys off `torch.cuda.is_available()` + the **default** device capability. If `x` is a CPU tensor while a CUDA device exists (CPU init / offload / a CPU unit test), it returns `True` and runs `torch._scaled_mm` on CPU FP8 → crash instead of taking the bf16 fallback. Gate on `x.is_cuda` and pass `x.device` into the check. (This is also what lets the fallback back a portable CI test for the MAJOR.)
- `fp8linear.py:82-99` — the `"channel"` (row-wise) branch passes 2-D scales (`x_scale [M,1]`, `scale_b [1,N]`) to `torch._scaled_mm`, which wants 1-D `[M]`/`[N]` → would error. It's currently **unreachable** (every caller passes `"tensor"`; the `forward` default is `"tensor"`), so it's dormant dead code with a latent bug. Either wire row-wise scaling properly (`.squeeze(-1)` + a `granularity` knob) or drop the `"channel"` paths.
- `fp8linear.py:131-135` — `fp8_linear_forward` is never called (`apply()` invokes `_LinearFWD8BWD16Fn.apply` directly), and its `-> torch.Tensor` annotation is wrong (it returns `(tensor, None)`). The FP4 sibling's `fp4_linear_forward` is also unused, so this is a pre-existing vestigial pattern — fine to drop both in a follow-up.
- `fp8linear.py:16-21` — `_supports_fp8_compute()` calls `torch.cuda.get_device_capability()` on every forward of every quantized linear. Capability is static per device — cache it (`functools.lru_cache` keyed on device). Pairs naturally with the device-aware refactor above.

## Notes (not blocking)
- **Rebase is done.** `MERGEABLE`; 1-ahead/0-behind `main`; the registry conflict with merged #1463 is resolved correctly — `quantization/__init__.py` carries **both** `nvfp4_qat_train` and `fp8_qat_train` in the `Literal` (`:5`), the imports (`:56-57`), and `method_to_config` (`:63-64`). The stale mergify "merge conflicts" comment predates the rebase.
- `get_min_capability()` returning `89` is moot — the method has no callsite anywhere in `fastvideo/` (the merged FP4-train config's `100` is also never checked), so the value gates nothing today.
- `pre-commit` is red, but it's yapf-only on `fastvideo/api/errors.py`, which this PR doesn't touch — a pre-existing base issue, not this diff.
- Buildkite legs are re-running post-rebase; suggest `/test training` (or `/test full`) before merge once they settle.

SolitaryThinker (approved):

PR #1465, opened 2026-06-16
[feat] QAD 5090: FP8 linear layer inference

kevin314:
## Purpose

Continues the QAD 5090 stack (#1464). Adds a generic FP8 inference quantization path for DiT linear layers (attention projections + MLP), so models can be served in FP8 after training.

## Changes

- `layers/quantization/fp8_config.py`: a new FP8Config and FP8QuantizeMethod that matches Wan's to_q/k/v/out + ffn.fc_in/fc_out layers by suffix
- Supports per-tensor (fast, default) and per-channel (higher accuracy) granularity via _scaled_mm
- Falls back to bf16 dequant on pre-sm89 GPUs
- `layers/quantization/__init__.py`: registers FP8 in QuantizationMethods and wires FP8Config into the quant registry.
## Test Results (sm89 / RTX 4090) and RTX 5090

<details>

```
End-to-end video generation completes successfully with fp8 quant
```

</details>

## Checklist

- [ ] I added or updated tests for my changes
- [ ] I considered GPU memory impact of my changes

### DIFF
--- a/docs/inference/optimizations.md
+++ b/docs/inference/optimizations.md
@@ -11,6 +11,9 @@ This page describes the various options for speeding up generation times in Fast
   - [Sliding Tile Attention (Archived)](#sliding-tile-attention-archived)
   - [Sage Attention](#sage-attention)
   - [Sage Attention 3](#sage-attention-3)
+
+- [FP8 Weight Quantization](#fp8-weight-quantization)
+
 - [Adaptive Guidance (CFG gating)](#adaptive-guidance-cfg-gating)
 
 - [torch.compile](#torch-compile)
@@ -217,6 +220,50 @@ To use Sage Attention 3 in FastVideo, follow the `README.md` in the linked repos
 These backends are model-specific and require the corresponding kernels and
 dependencies. Use the support matrix and model examples to confirm compatibility
 before enabling them.
+
+## FP8 Weight Quantization
+
+**`transformer_quant="FP8"`**
+
+Quantizes DiT linear layers (attention projections and FFN) to FP8 e4m3.
+
+On GPUs older than sm89, the FP8 matmul falls back to a bf16 dequant path
+automatically.
+
+### Requirements
+
+- **GPU**: sm89+ (H100, L40S, RTX 4090, or newer) for hardware FP8 compute
+- No additional packages required beyond the base FastVideo install
+
+### Usage
+
+```python
+from fastvideo import VideoGenerator
+from fastvideo.layers.quantization import get_quantization_config
+
+gen = VideoGenerator.from_pretrained(
+    "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
+    # Pass an instance — the bare string is not resolved on the from_pretrained path.
+    transformer_quant=get_quantization_config("FP8")(),          # per-tensor (default)
+    # transformer_quant=get_quantization_config("FP8")(granularity="channel"),  # slower, higher accuracy
+)
+gen.generate(request={"prompt": "A raccoon in sunflowers", "output": {"save_video": True}})
+```
+
+Or run the example script:
+
+```bash
+python examples/inference/optimizations/fp8_wan2_1_1_3b.py
+python examples/inference/optimizations/fp8_wan2_1_1_3b.py --granularity channel
+python examples/inference/optimizations/fp8_wan2_1_1_3b.py --bf16  # baseline
+```
+
+### Granularity
+
+| Mode | Weight scales | Activation scales | Speed | Accuracy |
+|------|--------------|-------------------|-------|----------|
+| `tensor` (default) | per-tensor | per-tensor | faster | lower |
+| `channel` | per-output-channel | per-token (rowwise) | slower | higher |
 
 <a id="torch-compile"></a>
 
--- a/examples/inference/optimizations/fp8_wan2_1_1_3b.py
+++ b/examples/inference/optimizations/fp8_wan2_1_1_3b.py
@@ -1,0 +1,140 @@
+"""FP8 weight quantization inference example.
+
+Runs Wan2.1-T2V-1.3B with FP8 e4m3 quantized DiT linear layers (attention
+projections and FFN). Weights are quantized in-place after loading; activations
+are quantized dynamically at runtime. Reduces GPU memory relative to BF16 and
+can improve throughput on sm89+ GPUs.
+
+Requirements:
+    - GPU: sm89+ (H100, L40S, RTX 4090, Ada Lovelace, or newer)
+      Falls back to a bf16 dequant path on older GPUs.
+    - TAEHV (optional): Follow install instructions at https://github.com/madebyollin/taehv
+
+Usage:
+    python fp8_wan2_1_1_3b.py              # FP8 per-tensor (default)
+    python fp8_wan2_1_1_3b.py --bf16       # BF16 baseline
+    python fp8_wan2_1_1_3b.py --granularity channel  # per-channel (higher accuracy but slower)
+    python fp8_wan2_1_1_3b.py --taehv-checkpoint /path/to/taew2_1.pth
+"""
+
+import argparse
+import os
+import sys
+import time
+
+import torch
+
+OUTPUT_PATH = "video_samples"
+
+
+def load_taehv(checkpoint_path, device="cuda", dtype=torch.float16):
+    repo_dir = os.path.dirname(checkpoint_path)
+    if repo_dir not in sys.path:
+        sys.path.insert(0, repo_dir)
+    from taehv import TAEHV
+    print(f"Loading TAEHV from {checkpoint_path}...")
+    model = TAEHV(checkpoint_path=checkpoint_path).to(device, dtype)
+    print("TAEHV loaded.")
+    return model
+
+
+@torch.no_grad()  # type: ignore[misc]
+def decode_with_taehv(taehv_model, latents):
+    latents = latents.permute(0, 2, 1, 3, 4)
+    latents = latents.to(device=next(taehv_model.parameters()).device,
+                         dtype=next(taehv_model.parameters()).dtype)
+    decoded = taehv_model.decode_video(latents, parallel=False, show_progress_bar=False)
+    frames = []
+    for frame in decoded[0]:
+        frame_np = (frame.clamp(0, 1) * 255).byte().cpu().permute(1, 2, 0).numpy()
+        frames.append(frame_np)
+    return frames
+
+
+def main():
+    parser = argparse.ArgumentParser(description="FP8 video generation benchmark")
+    parser.add_argument("--bf16", action="store_true",
+                        help="BF16 baseline (no FP8 quantization)")
+    parser.add_argument("--granularity", choices=["tensor", "channel"], default="tensor",
+                        help="FP8 weight scale granularity: tensor (faster) or channel (more accurate)")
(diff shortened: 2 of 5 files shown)


SolitaryThinker (changes requested):
**Hi @kevin314** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

**Open PR** · scope: model · reviewed @ `10b4bf18`. The post-rebase diff is clean (2 files, +247/-1); the FP8 QAT-*training* path is #1464 (already on `main`) and isn't re-reviewed here.

## Verdict: REQUEST_CHANGES
The FP8 inference path as committed is unreachable — the weight-conversion pass is never called, so the first FP8 linear forward raises `AttributeError`. Also missing SSIM/quality evidence for a model-scope numerics change.

### Blocking
- **BLOCKER** `fastvideo/layers/quantization/fp8_config.py:121` (root cause `:207`) — **the FP8 path is dead: `apply()` reads a buffer nothing registers.** `FP8QuantizeMethod.apply` opens with `out_dim = layer._fp8_weight.shape[0]` (and `_apply_dequant` at `:162`), but `_fp8_weight` / `_fp8_weight_scale` are registered *only* by `convert_model_to_fp8` (`:229-230`), which **has zero callers in the whole tree**:
  - `grep -rn "convert_model_to_fp8\|FP8QuantizeMethod" fastvideo/ --include=*.py` matches only inside `fp8_config.py` (self-defs + `__all__`) — no caller, no dynamic dispatch.
  - The sibling NVFP4 / NVFP4-QAT methods wire their conversion into the loader at `fastvideo/models/loader/fsdp_load.py:31-66` (`_maybe_convert_model_to_nvfp4`, invoked `:206` from the inference DiT load), dispatching to `convert_model_to_nvfp4` / `convert_model_to_fp4`. **There is no `FP8QuantizeMethod` branch.**
  - The base seam `QuantizeMethodBase.process_weights_after_loading` (`base_config.py:41`) exists but is never auto-invoked by the loader, so it isn't the implicit hook either.
  - Wan (the target) builds `to_q/k/v/out` + `ffn` via `ReplicatedLinear(..., quant_config=...)` (`models/dits/wanvideo.py:124-127`) and calls them through `ReplicatedLinear.forward` → `quant_method.apply(self, x, bias)` (`linear.py:296`), so the missing buffer crashes on the first projection.

  Not a corner case — it's the *only* runtime path: with `FP8` selected, every targeted linear raises `AttributeError: ... has no attribute '_fp8_weight'` on the first forward. Fix: add an `FP8QuantizeMethod` branch to `_maybe_convert_model_to_nvfp4` (or a generalized post-load hook) calling `convert_model_to_fp8(model)`, mirroring the FP4 wiring, plus a runnable example like `examples/inference/optimizations/nvfp4_qat_wan2_1_1_3b.py`. (Gemini flagged the same `AttributeError` surface at `:125` as a conditional edge case — the sharper point is that `convert` is invoked *never*, so it fails 100% of the time.)

### Major
- **MAJOR** no SSIM / quality evidence for a `scope: model` precision change. The body's Test Results are smoke-only ("End-to-end video generation completes successfully with fp8 quant") and the template "I verified SSIM regression tests pass" box is unchecked. A dynamic FP8 (e4m3) activation+weight quant on Wan's linears needs an SSIM run attached. Note: that smoke claim can't have exercised *this* diff given the BLOCKER above — please confirm what was actually run (possibly an earlier local version that wired the convert pass).

### Minor
- **MINOR** `fp8_config.py:98-112,127-132` — the prequant plumbing (`quantize_input` / `wants_prequantized_input` / `pre_quantized`) is dead for this PR: Wan never calls `quantize_input`, and only LTX-2's attention forward uses that path (LTX-2 isn't wired to FP8 here). Wire a consumer or drop the surface until one exists.
- **MINOR** naming — there are now three FP8-family methods (`AbsMaxFP8`, new `FP8`, `fp8_qat_train`). They're genuinely distinct (`AbsMaxFP8` = static checkpoint scales; `FP8` = dynamic scales at load+runtime; `fp8_qat_train` = STE training), but the bare name `FP8` is easy to confuse with `AbsMaxFP8` — consider `fp8_dynamic` + a contrasting docstring.

### Nit
- **NIT** `fp8_config.py:188` — `get_min_capability()` returns 89 but nothing enforces it (the real gate is the runtime `_supports_fp8_compute()` + bf16 fallback). Consistent with the sibling configs; documentation-only.

The registry wiring (`__init__.py`: `Literal` + import + `method_to_config`) is internally consistent ✅. The pre-commit CI failure is a pure `yapf` line-wrap reflow — style is owned by pre-commit, not flagged here. Gemini's robustness nits (empty-tensor `amax()` guard, multi-GPU `get_device_capability(device)`, meta-device guard, fallback device/dtype cast) are reasonable and not re-litigated.

SolitaryThinker:
**Hi @kevin314** — follow-up from Gob, one of @SolitaryThinker's AI reviewers. This expands the BLOCKER from the prior review with a full lifecycle trace and a side-by-side against the working NVFP4 path. Reviewed @ `10b4bf18`; ping @SolitaryThinker if anything looks off.

## TL;DR

The FP8 quant method has a **three-beat lifecycle** (attach → register buffers → read in forward). Beat 1 and beat 3 are present; **beat 2 — the load-time conversion pass — is never wired**, so the first FP8 linear forward dereferences a buffer that was never registered. NVFP4 implements all three beats; FP8 is missing only the middle one. This is a ~3-line loader fix, not a redesign.

## The quant-method lifecycle (three beats)

Every quantized linear in FastVideo goes through three distinct phases:

| Beat | Who runs it | What it does |
|---|---|---|
| **1. Attach method** | `LinearBase.__init__` → `config.get_quant_method(self, prefix)` (`fastvideo/layers/linear.py:200`) | Attaches a `*QuantizeMethod` instance as `layer.quant_method` if the layer's `prefix` matches the targeted suffixes |
| **2. Register quant buffers** | a load-time conversion walk in `fsdp_load.py` | Reads the freshly-loaded bf16 `weight`, computes scales, and `register_buffer(...)`s the quantized weight/scale onto each tagged layer |
| **3. Read buffers in forward** | `*Linear.forward` → `quant_method.apply(self, x, bias)` (`linear.py:287`) | Dereferences the buffers registered in beat 2 |

## NVFP4: all three beats wired ✅

```
Beat 1 (construct):
  ReplicatedLinear(..., quant_config=NVFP4Config)
    → LinearBase.__init__ (linear.py:200)
        → NVFP4Config.get_quant_method(layer, prefix)
            → attaches NVFP4QuantizeMethod as layer.quant_method

Beat 2 (load):                          ← THE HOOK
  load_model_from_full_model_state_dict (fsdp_load.py)
    → :199  _maybe_convert_model_to_nvfp4(model)
        → :54-59  for mod in model.modules():
                    if isinstance(mod.quant_method, NVFP4QuantizeMethod):
                        convert_model_to_nvfp4(model)   # nvfp4_config.py:433
                        return                          # registers _nvfp4_weight* / _nvfp4_alpha

Beat 3 (forward):
  ReplicatedLinear.forward → quant_method.apply(self, x, bias)  (linear.py:287)
    → reads _nvfp4_* buffers   ← exist (registered in beat 2)   ✅
```

The FP4-QAT sibling does the same: `convert_model_to_fp4` (`nvfp4_qat_config.py:121`), discovered by the same loader walk.

## FP8 in this PR: beat 2 is missing ❌

```
Beat 1 (construct):
  ReplicatedLinear(..., quant_config=FP8Config)
    → LinearBase.__init__ (linear.py:200)
        → FP8Config.get_quant_method(layer, prefix)   (fp8_config.py:200-205)
            isinstance(layer, LinearBase) and any(s in prefix for s in _FP8_SUFFIXES)
            → attaches FP8QuantizeMethod as layer.quant_method      ✅

Beat 2 (load):                          ← MISSING
  fsdp_load.py:54-56 gate is hard-isinstance'd to NVFP4QuantizeMethod ONLY.
  FP8QuantizeMethod extends QuantizeMethodBase directly (fp8_config.py:63),
  NOT NVFP4QuantizeMethod, so the walk never fires for it.
  convert_model_to_fp8 (fp8_config.py:207) — which holds the only
  register_buffer("_fp8_weight", ...) calls (:229-230) — has ZERO callers.   ❌

Beat 3 (forward):
  ReplicatedLinear.forward → quant_method.apply(self, x, bias)  (linear.py:287)
    → FP8QuantizeMethod.apply opens with:
        out_dim = layer._fp8_weight.shape[0]   (fp8_config.py:121)
    → AttributeError: ... has no attribute '_fp8_weight'        💥
```

### Proof that beat 2 has no caller

```
$ git grep -n "convert_model_to_fp8\|FP8QuantizeMethod" -- 'fastvideo/**/*.py'   # at 10b4bf18
fastvideo/layers/quantization/fp8_config.py:63   class FP8QuantizeMethod(...)      (self-def)
fastvideo/layers/quantization/fp8_config.py:203  return FP8QuantizeMethod(...)     (self-construct)
fastvideo/layers/quantization/fp8_config.py:207  def convert_model_to_fp8(...)     (self-def)
fastvideo/layers/quantization/fp8_config.py:215  isinstance(qm, FP8QuantizeMethod) (inside convert)
fastvideo/layers/quantization/fp8_config.py:242  "FP8QuantizeMethod"               (__all__)
fastvideo/layers/quantization/fp8_config.py:243  "convert_model_to_fp8"            (__all__)
```

Every match is inside `fp8_config.py` itself. No loader caller, no `__init__.py` re-export into the loader, no dynamic dispatch, and `QuantizeMethodBase.process_weights_after_loading` (`base_config.py:41`) is never auto-invoked by the loader either — so it isn't an implicit hook.

### Why this fails 100%, not as an edge case

`get_quant_method` succeeding (beat 1) is exactly what makes the crash *guaranteed*: the method **is** attached, so `forward` dutifully calls `apply()`, which dereferences `_fp8_weight` on its first line. Wan builds `to_q/k/v/out` + `ffn.fc_in/fc_out` via `ReplicatedLinear(..., quant_config=...)` (`models/dits/wanvideo.py:124-127`) and routes them through `ReplicatedLinear.forward → apply()` (`linear.py:287`). So with `FP8` selected, the *first targeted projection of the first forward pass* raises — there is no path that avoids it. (Gemini flagged the same `AttributeError` surface at `:125` as a conditional edge case; the sharper framing is that the convert pass runs *never*.)

## Suggested fix (mirrors NVFP4, ~3 lines)

Generalize the loader hook to recognize the FP8 method too:

```python
# fastvideo/models/loader/fsdp_load.py — inside the same model.modules() walk
from fastvideo.layers.quantization.fp8_config import (
    FP8QuantizeMethod, convert_model_to_fp8)   # keep lazy, like the NVFP4 import
...
for mod in model.modules():
    qm = getattr(mod, "quant_method", None)
    if isinstance(qm, NVFP4QuantizeMethod):
        convert_model_to_nvfp4(model); return
    if isinstance(qm, FP8QuantizeMethod):
        convert_model_to_fp8(model); return
```

(Longer-term: route everything through `QuantizeMethodBase.process_weights_after_loading` and have each method implement it — cleaner, but bigger blast radius, so a follow-up rather than this PR.)

Plus, to clear the rest of the prior review: a runnable example (mirroring `examples/inference/optimizations/nvfp4_qat_wan2_1_1_3b.py`) and an SSIM run for the `scope: model` precision change. The design itself is sound — the name-match-at-construct / type-match-at-convert handshake is consistent with NVFP4; it's purely the missing wiring.

PR #1466, opened 2026-06-16
[Bugfix] QAD 5090: Torch.compile and other optimizations (15/12)

loaydatrain:
## Purpose

Continues the QAD 5090 stack (#1225). Inference-side companion to #1463 (FP4
QAT training): wires the FP4 linear path for Wan2.1 1.3B end-to-end under
`torch.compile`, pops the bf16 weight after FP4 packing, removes `x_global_sf`, and adds TAEHV scripts

With Sage3, running in 2it/s, 1.78s e2e.

## Changes
- `layers/quantization/nvfp4_qat_config.py` (+ `loader/fsdp_load.py`):
  inference path made `torch.compile`-friendly — precomputed `x_global_sf`
  (no per-call sync), single-level quant, and the dense bf16 weight is popped
  after FP4 pack (only `_fp4_weight` + scales survive).
- `attention/backends/sage_attn3.py`: removed data-dependent control flow on
  tensor metadata so the backend traces cleanly under `torch.compile`.
- `examples/inference/optimizations/fp4_linear_wan2_1_1_3b.py` and
  `FastWan_QAD_TAEHV.py`: end-to-end Wan2.1 1.3B inference with the FP4 linear
  path + sage_attn3 + `torch.compile`; the TAEHV variant swaps in the TAEHV
  decoder.

No new CLI flags; reuses `--transformer-quant nvfp4_qat` from #1463.

## Test results (RTX 5090 / sm_120)

- ✅ Inference: `--transformer-quant nvfp4_qat` Wan2.1 1.3B runs end-to-end
  with `torch.compile` enabled; sample output
  `video_samples/raccoon_fp4_linear_taehv_compile.mp4`.
- ✅ Step time: DiT time: 2it/s // e2e time: 1.78s for a 5s 480p video
- ✅ Memory: dense bf16 weight popping drops DiT resident memory after the
  first forward by TODO MB.

Depends on #1463 (`--transformer-quant` CLI + `nvfp4_qat` config). Part of #1225.

### DIFF
--- a/examples/inference/optimizations/FastWan_QAD_TAEHV.py
+++ b/examples/inference/optimizations/FastWan_QAD_TAEHV.py
@@ -1,0 +1,286 @@
+"""Fast NVFP4 linear inference for Wan2.1-T2V-1.3B with TAEHV decoding.
+
+This is the FP4-linear fast path from ``fp4_linear_wan2_1_1_3b.py`` with the
+heavy Wan VAE swapped out for TAEHV -- a tiny autoencoder that decodes Wan2.1
+latents directly (no denormalization) and is dramatically faster / lighter.
+
+How it works: the generator runs with ``output_type="latent"`` so the pipeline
+returns raw denoised latents instead of pixels (the Wan VAE is offloaded and
+never used). We then decode those latents with TAEHV in this script and save
+the frames ourselves. This mirrors the FastVideo-Quantization
+``quantization_example_taehv.py`` proof-of-concept, but kept clean: TAEHV is a
+pip package (no ``sys.path`` hacks), the latent->uint8 conversion is vectorized,
+and there is no dead profiler / sanitization code.
+
+Requirements:
+    - Blackwell GPU (B200/B300, sm100a/sm103a) for the FP4 linear path
+    - flashinfer (``pip install flashinfer-python``)
+    - TAEHV weights ``taew2_1.pth`` (https://github.com/madebyollin/taehv)
+
+Usage:
+    python fp4_linear_taehv_wan2_1_1_3b.py                    # FP4 + TAEHV + compile
+    python fp4_linear_taehv_wan2_1_1_3b.py --no-taehv         # FP4 + full Wan VAE
+    python fp4_linear_taehv_wan2_1_1_3b.py --no-compile       # eager
+    python fp4_linear_taehv_wan2_1_1_3b.py --baseline         # dense bf16 reference
+    python fp4_linear_taehv_wan2_1_1_3b.py --distilled_model ''  # base Wan2.1 weights
+"""
+
+import argparse
+import contextlib
+import logging
+import os
+import time
+
+import imageio
+import torch
+
+from fastvideo import VideoGenerator
+from fastvideo.configs.pipelines.base import PipelineConfig
+from fastvideo.layers.quantization.nvfp4_qat_config import NVFP4QATConfig
+
+OUTPUT_PATH = "video_samples"
+
+# Distilled, quantization-aware (QAD) transformer for Wan2.1-1.3B (3 steps,
+# guidance 1.0). Loaded on top of the base Wan2.1 pipeline; pass
+# ``--distilled_model ''`` to run the base weights instead.
+DEFAULT_DISTILLED_MODEL = "FastVideo/FastWan-QAD-1.3B"
+DISTILLED_WEIGHTS_FILE = (
+    "generator_inference_transformer/diffusion_pytorch_model.safetensors"
+)
+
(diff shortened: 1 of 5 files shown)


SolitaryThinker (commented):
**Hi @loaydatrain** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

**Scope:** training/quant + perf (the FP4-linear `torch.compile` enablement + optimizations). **Open PR, `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` + `needs-rebase`.** Verdict: **COMMENT** — the compile fix is correct and the core logic lands on files `main` hasn't touched, but rebase first and confirm two things below before merge.

## Major

- **Needs rebase — the branch is stacked on the *pre-merge* #1463 commit (`09f9b56`).** #1463 already squash-merged to `main` (`6da206e1`), and that base is also ~13k lines behind current `main` (the big deletion count GitHub shows is this artifact, not intended changes). The actual textual conflict is just `fastvideo/models/loader/fsdp_load.py` and `fastvideo/fastvideo_args.py`, and both PR versions are **semantically equivalent** to what #1463 already put on `main` (variable name / branch order / comment wording in `fsdp_load.py`; pure formatting reflow of the existing `--transformer-quant` arg in `fastvideo_args.py`). Rebase onto `main` and drop those two near-dup hunks — the substantive logic (sage_attn3, `attention/layer.py`, `nvfp4_qat_config.py`) is conflict-free.

- **Activation-quant numerics change with no quality evidence — `nvfp4_qat_config.py:84,106`.** `apply()` switched the input scale from per-call dynamic `(448*6)/x.absmax()` to static `x_global_sf = torch.tensor(1.0)`. The comment ("matches the production path") is **correct** — the LTX-2 NVFP4 method uses the same static `1.0` (`nvfp4_config.py:270,338`), and this is exactly what removes the per-call `.max()` sync blocking CUDA-graph capture. But it's still a real FP4 activation-quant change and the body ships no SSIM/quality number (the memory claim is a literal "TODO MB"). Per the repo's model-PR convention, please post an SSIM delta vs dense bf16. Also note it diverges from the **training STE**, which still quantizes activations with the dynamic scale (`fp4linear.py:50`) — confirm the train-vs-inference input-scale skew is intended for QAT.

- **Latent layout-contract break in the sage3 hook refactor.** The old impl transposed q/k/v *inside* `forward()` after the chunk; the new `preprocess_qkv` (`sage_attn3.py:58-70`) does `permute(0,2,1,3)` on the **stacked** tensor, which `layer.py:134` / `ltx2.py:1303` call **before** the replicated-QKV concat (`dim=1`) and the post-attention replicated split. After the permute dim-1 is heads, not sequence, so the replicated-QKV concat/split no longer mean what they did. **Safe for this PR's Wan target** (`replicated_q` is never passed there, and it's never populated anywhere in-tree today), but LTX-2's non-VSA path carries the replicated-QKV plumbing and can select `SAGE_ATTN_THREE`, so the contract is silently wrong for any future caller. Either preserve the per-tensor transpose for the replicated path, or assert `replicated_q is None` in the sage3 path so a break is loud, not silent.

## Minor

- **No div-by-zero guard — `nvfp4_qat_config.py:201-202`:** `weight_global_sf = (448*6) / weight_absmax` divides by an unclamped reduced absmax; an all-zero tile gives `inf`/`nan`. A `.clamp_min(eps)` is cheap insurance.
- **Hardcoded path — `FastWan_QAD_TAEHV.py:53`:** `DEFAULT_TAEHV_CHECKPOINT = "/root/taehv/taew2_1.pth"` is environment-specific; default it to a repo-relative / env-driven value so the example runs out of the box.

## Nits
- TAEHV docstring (`FastWan_QAD_TAEHV.py:3,21-25`) references a sibling `fp4_linear_wan2_1_1_3b.py` and a `python fp4_linear_taehv_wan2_1_1_3b.py` usage line that don't match the actual filename (`FastWan_QAD_TAEHV.py`).
- PR body "drops DiT resident memory ... by TODO MB" — fill in the number (it's the headline benefit of the weight-pop).

**Verified positives:** reusing the `register_fake`-backed `fastvideo_fp4::` custom ops instead of the old `@torch.compile` + raw flashinfer is the correct graph-break fix (no silently-swallowed breaks); the weight-pop memory logic is sound (`persistent=False` buffers + `_parameters.pop` + `grad=None` + `gc`/`empty_cache`); the `FASTVIDEO_DISABLE_ATTENTION_COMPILE` env-gate defaults to the historical eager behavior; title matches the merge regex; pre-commit is green. CI microscope failures are uniform `exit 124` timeouts consistent with the stale/conflicting base — rebase and re-run.

PR #1471, opened 2026-06-19
[feat] Add Kandinsky-5 T2V/I2V pipeline support

aryan5v:
## Summary

Adds first-class Kandinsky-5 T2V and I2V support through the normal FastVideo model-support path:

- Kandinsky-5 T2V and I2V pipeline configs and default presets
- `basic/kandinsky5` composed T2V and I2V pipeline wiring
- Kandinsky-specific image encoding, latent prep, denoising, normalization, and latent decode stages
- registry/default-preset wiring for Lite, Pro, and distilled Kandinsky-5 T2V/I2V checkpoint variants
- loader fixes needed by Kandinsky's CLIP text encoder layout and text-encoder CPU offload
- transformer parity fixes in the existing Kandinsky5 DiT implementation
- NABLA sparse-attention configuration and denoising support for Kandinsky-5
- basic usage examples at `examples/inference/basic/basic_kandinsky5_t2v.py` and `examples/inference/basic/basic_kandinsky5_i2v.py`

## Validation

- `pre-commit run --files <Kandinsky implementation/example files>`: passed locally
- B200 registry/import smoke: passed for LTX2, MatrixGame2/3, Flux2, Stable Audio, and all Kandinsky Lite/Pro/distilled T2V/I2V preset resolutions
- `uv run pytest tests/local_tests/kandinsky5/test_kandinsky5_lite_transformer_parity.py -q -s -rs`: passed on B200 GPU, `1 passed, 14 warnings`
- `CUDA_VISIBLE_DEVICES=0 uv run python tests/local_tests/kandinsky5/run_kandinsky5_lite_pipeline_smoke.py`: passed, one-step latent smoke generated successfully
- High-quality generation validation on B200 GPU 4: [W&B run `fuczhqid`](https://wandb.ai/aryan5v-san-jose-state-university/fastvideo-kandinsky5-validation/runs/fuczhqid), `768x512`, 121 frames, 80 inference steps, guidance scale 5.0, output `outputs/kandinsky5_validation/kandinsky5_red_motorcycle_best_quality_512x768_121f_80s.mp4`
- High-quality otter generation validation on B200 GPU 7: [W&B run `0n0w1nr4`](https://wandb.ai/aryan5v-san-jose-state-university/fastvideo-kandinsky5-validation/runs/0n0w1nr4), `768x512`, 121 frames, 100 inference steps, guidance scale 5.0, elapsed 421.21s, generation 401.71s, peak memory 10365.89 MB, output `outputs/kandinsky5_validation/kandinsky5_otter_best_quality_512x768_121f_100s.mp4`


leffff:
Hi! @aryan5v my name is Lev (Kandinsky Lab)! 
I am here to review your PR, fasten the merge, and add more features.
I have opened a PR that adds example for Kandinsky 5 T2V Lite 5s.

leffff:
@aryan5v I will also add support for Kandinsky 5 T2V Pro 5s. 
It needs a flex attention backend. I will add it

leffff:
@SolitaryThinker pls review

SolitaryThinker:
Maintainer update: this got a deep review. The Kandinsky port itself is in good shape — properly registry-integrated, stages composed the right way, official Diffusers weights load directly, and the NABLA backend with layer-level `default_backend` is a clean pattern. Several things that looked suspicious were checked and dismissed (selector ordering, I2V image conditioning matches diffusers, config pins verified against all 7 official model_index.json files).

I rebased onto current main and pushed one fix commit (9f28d34d): **reverted the global `text_encoder_cpu_offload` default flip** back to True. That flip changed the default VRAM footprint for every model family (small-GPU users would OOM on Wan/LTX2/Hunyuan loads) and was the exact cause of your red unit lane — `test_parser.py`'s golden dict still expects True; I reproduced the failure and verified the fix (202/202 api tests pass on GB200 now). Your examples pass `text_encoder_cpu_offload=True` explicitly and your new offload-aware TextEncodingStage is what makes offload work for Kandinsky's CLIP — so nothing in this PR needs the flip. If you want that default changed, it deserves its own PR.

What still needs attention before merge:
1. **Encoder + transformer lanes were red at the old head** — the rebase may clear them (fresh run coming), but if not, the prime suspects are the `qwen2_5_vl_custom.py` transformers-compat edits (the flash→SDPA silent fallback is a numerics change for HY1.5/Reason1 on transformers 4.x). Needs attribution.
2. **Zero CI-collected test coverage** — the only Kandinsky test is weight-gated in `tests/local_tests/`, which no lane collects, and this PR modifies the DiT forward. Repo convention wants a collected regression test + SSIM refs for a new pipeline. Happy to help seed SSIM refs once weights/plan are agreed.
3. The **Modal `pr_test.py` checkout rewrite** bundled in here affects every PR's CI — that should be its own PR (or explicitly blessed by the maintainers as-is).

Also: heads up that the PR is titled draft-pr but isn't a draft — if you consider it ready, the above is the path.

SolitaryThinker:
Follow-up done: the bundled Modal checkout rewrite is dropped from this PR (pr_test.py now matches main — that work should be its own PR if still wanted), and the PR now carries a reduced-size SSIM regression test (256x384, 21 frames, `test_kandinsky5_similarity.py`). The L40S reference is seeded on FastVideo/ssim-reference-videos and a full Modal verify run passes against it (1/1). One note baked into the test: `FASTVIDEO_FA4=0` — the SSIM image's FA4 pin crashes the CLIP encoder's dense-MHA path on L40S (CuTeDSL/cutlass-dsl mismatch); that image issue is being tracked separately. With the offload revert, the SSIM coverage, and the infra split-out, this is in good shape — remaining: encoder/transformer lanes green on the current head.

leffff:
@aryan5v can you please check on tests?

SolitaryThinker:
Good news on the two red lanes: attributed, and **they're not this PR's fault**. All 5 failing tests die on the same FA4 CuTeDSL `fmax` TypeError — the CI-image/cutlass-dsl-4.6 break (issue #1558) that hit every PR; #1564 fixed the pin and the image rebuild is in progress. Your compat edits are actually green at head (qwen2.5/t5/siglip encoder tests all pass), and your latest commit fixed the unit lane. Once the rebuilt image lands we'll retry the lanes — expecting green with no changes needed here.

SolitaryThinker:
Pushed 72d44fd9 on top of your head — a small follow-up on the sdpa mask handling, plus a regression test.

**What broke:** 69184185 moved the 2D → 4D mask lift into `SDPAMetadataBuilder.build()`, but that metadata is cross-backend: HYWorld and HunyuanVideo15 build `SDPAMetadata` while the layer selector picks FLASH_ATTN in CI. The FLASH_ATTN equal-length branch padded `attn_mask.shape[1]` assuming a 2D mask, so a 4D `[B,1,1,text_len]` mask (where `shape[1] == 1`) produced a garbage-length pad → out-of-bounds gather → CUDA device-side assert, which SIGABRT'd the whole pytest session in build 4112 (test_hyworld crashed; test_matrixgame2/test_wanvideo never got to run). Really subtle cross-backend contract — easy to miss.

**Your 98116070 already stops the crash** from the consumer side (normalizing whatever shape reaches the flash branch), and it stays. My commit hardens the producer side to match: the builder now stores masks exactly as passed (2D stays 2D — the convention every in-tree producer/consumer assumes), and the 2D → `[B,1,1,key_len]` lift happens only inside `_normalize_attn_mask_for_sdpa`, the torch-sdpa-only consumer. Your integer → bool coercion and the torch-sdpa padding semantics from 69184185 are fully preserved — the lift just lives in the consumer that needs it instead of the shared metadata. Also adds a CPU-only contract test (`fastvideo/tests/attention/test_sdpa_metadata_mask_contract.py`) and wires it into the CI unit lane so the contract stays guarded.

**GB200 (4×) verification:**
- `test_hyworld + test_matrixgame2 + test_wanvideo + the new contract test` in one session: no CUDA assert, no SIGABRT. test_wanvideo and both contract tests pass; test_matrixgame2 is skipped-by-design ("Not reliably reproducible").
- test_hyworld fails its latent tolerance on GB200 (FLASH 2.04%, TORCH_SDPA 4.35% vs the recorded references) — but that's pre-existing hardware variance, not this change: I ran your head 98116070 as a baseline on the same node and got **bit-identical latents** to 72d44fd9 on both backends (FLASH: -201149.9431345705, TORCH_SDPA: -201834.08767374977). The references were recorded for the L40S CI lane, where this should reproduce as before.

Notably the two backends now agree with each other to ~0.3% on GB200, while the two recorded references disagree by ~7% — consistent with the mask-semantics inconsistency both of these commits fixed.

SolitaryThinker:

SolitaryThinker:
Pure rebase onto current main to pick up #1565 (memory cleanup between train-framework tests) and #1566 (CI lane CPU/memory reservations); no content changes (range-diff all `=`). Old head 1d0235ef -> new head 66ac38ba.

SolitaryThinker:
@aryan5v rebased this branch onto current main (595f0ea6) after #1538 (DreamX-World) merged — the two branches overlapped in the pipeline-config registry exports and `pr_test.py`. Resolution was a pure union (kept both DreamX and Kandinsky entries); no content changes to any Kandinsky commit — range-diff shows all 37 commits identical apart from the `__all__` merge lines. Old head 66ac38ba -> new head 233d403d.

SolitaryThinker (approved):

PR #1472, opened 2026-06-19
[perf] Default Wan VAE decode to bf16 (lossless, faster)

Mister-Raggs:
## Summary

bf16 VAE **decode** on `AutoencoderKLWan` is effectively lossless (MS-SSIM 0.9999 vs fp32 on an identical latent) and ~1.2–1.3× faster — the same VAE already ships bf16 in Cosmos-Predict2.5 / fp16 in Cosmos. fp32 here was just the inherited `PipelineConfig` default, not a Wan-specific requirement, and decode is the majority cost on bandwidth-limited devices.

The catch: `vae_precision` is a single knob that drives both VAE **encode** and **decode**. bf16 is only lossless for *decode* (output-only). For I2V/causal Wan variants (`WanI2V480PConfig`, Matrix-Game-2.0, …) the *encode* of the conditioning image seeds the denoising trajectory, so lowering it to bf16 shifts the seed latent; in a causal autoregressive rollout that divergence compounds (Matrix-Game-2.0 SSIM **0.98 → 0.76**). A naive `vae_precision = bf16` flip on the base config cascades into exactly this.

So this PR **decouples decode precision from encode precision**:

- Adds an optional `PipelineConfig.vae_decode_precision` (default `None` → falls back to `vae_precision`), so **no behavior change for any other model**.
- `DecodingStage.decode` + `streaming_decode` honor the override.
- Wan configs keep `vae_precision = fp32` (encode untouched → trajectory bit-identical to references) and set `vae_decode_precision = bf16`. Every Wan variant derives from `WanT2V480PConfig`, so the decode win reaches **all** of them — including I2V/causal — without touching encode.
- The two legacy `wan_*.json` configs get the same split (the 14B i2v one is `load_encoder: true`, squarely the case the decouple protects).

Net: faster bf16 decode everywhere it's safe; fp32 encode preserved where it matters.

## Evidence

Measured by decoding **one real Wan latent** fp32 vs bf16 in the same process (no denoise non-determinism, no codec noise):

- **MS-SSIM(bf16, fp32) = 0.9999** on the identical latent — effectively lossless. Random-latent control gave the same 0.9999.
- **~1.2–1.3×** faster VAE decode → roughly 5–10% end-to-end on decode-bound few-step models; free (quality-wise) on full-step.

Mechanism of the prior regression is structural (encode seeds the trajectory), confirmed against the config inheritance + stage code; the SSIM suite rerun is the end-to-end proof.

## Test plan

- [x] T2V (decode-only) SSIM — unchanged, 0.9999 vs fp32, well above threshold.
- [ ] I2V/causal (Matrix-Game-2.0) — encode restored to fp32 → trajectory should match references; decode bf16 is the 0.9999-lossless path. Expected to recover above the 0.98 threshold (was 0.76 under the cascaded encode flip). Confirming via the CI SSIM rerun on this push.
- [x] No-op for non-Wan models (`vae_decode_precision` defaults to `None`).

### DIFF
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -116,6 +116,7 @@ surfaces:
       dit_precision: "Precision override pending dedicated typed component precision design."
       upsampler_precision: "Precision override pending dedicated typed component precision design."
       vae_precision: "Precision override pending dedicated typed component precision design."
+      vae_decode_precision: "Decode-only precision override pending dedicated typed component precision design."
       image_encoder_precision: "Precision override pending dedicated typed component precision design."
       text_encoder_precisions: "Precision override pending dedicated typed component precision design."
     internal_only:
--- a/fastvideo/configs/pipelines/base.py
+++ b/fastvideo/configs/pipelines/base.py
@@ -51,6 +51,12 @@ class PipelineConfig:
     # VAE configuration
     vae_config: VAEConfig = field(default_factory=VAEConfig)
     vae_precision: str = "fp32"
+    # Optional decode-only precision override. When None, the decode stage falls
+    # back to `vae_precision`. This lets a pipeline run a faster, lossless bf16
+    # decode while keeping a higher-precision *encode* (the image/video VAE
+    # encode seeds the denoising trajectory, so lowering its precision can shift
+    # the output for I2V/causal models — decode is output-only and safe to lower).
+    vae_decode_precision: str | None = None
     vae_tiling: bool = True
     vae_sp: bool = True
 
@@ -132,6 +138,15 @@ class PipelineConfig:
             default=PipelineConfig.vae_precision,
             choices=["fp32", "fp16", "bf16"],
             help="Precision for VAE",
+        )
+        parser.add_argument(
+            f"--{prefix_with_dot}vae-decode-precision",
+            type=str,
+            dest=f"{prefix_with_dot.replace('-', '_')}vae_decode_precision",
+            default=PipelineConfig.vae_decode_precision,
+            choices=["fp32", "fp16", "bf16"],
+            help="Optional decode-only VAE precision override (falls back to "
+            "--vae-precision when unset)",
         )
         parser.add_argument(
             f"--{prefix_with_dot}vae-tiling",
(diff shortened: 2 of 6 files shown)


alexzms (approved):
Verified the diff (+7/-3): `wan.py:54` is the only `vae_precision` in the entire Wan config tree, so this single default change genuinely covers every Wan variant (T2V/I2V/720P/Wan2.2/LucyEdit/SelfForcing) with no override escaping it. The two legacy JSON configs are updated for consistency.

Evidence is solid: MS-SSIM 0.9999 (bf16 vs fp32 on an identical latent) = effectively lossless, ~1.2-1.3x faster VAE decode (~5-10% e2e on decode-bound few-step models), and the same AutoencoderKLWan already ships bf16 in Cosmos-Predict2.5 / gen3c / longcat. Well-precedented, no quality concern. LGTM.

Mister-Raggs:
Heads-up on two small surface additions in this PR, since they touch areas with parity tests:

**1. `vae_decode_precision` in the schema-parity inventory** (`docs/design/inference_schema_parity_inventory.yaml`)

Adding the new `PipelineConfig.vae_decode_precision` field requires classifying it, or `test_pipeline_config_base_fields_are_classified` fails. I put it in `compatibility_only`, right next to the sibling precision overrides (`vae_precision`, `dit_precision`, `upsampler_precision`, `image_encoder_precision`, `text_encoder_precisions`) — all carrying the same "pending dedicated typed component precision design" note. It's a plain-string entry, so it's **exempt** from `test_inventory_targets_exist_in_typed_schema` (that test only walks `moved`/`preset_owned` dotted targets against the typed schema). No typed-schema home is implied — when the precision-typing design lands, this field gets handled alongside the other `*_precision` fields, same as `vae_precision` already would.

**2. The `--vae-decode-precision` CLI flag** is added to the legacy `PipelineConfig.add_cli_args`, kept for parity with `--vae-precision`. The inference subcommands (`generate`/`serve`) are config-only (`expected_dests: [config]`) and never call `add_cli_args`, so this flag does **not** touch the typed-schema CLI surface or `test_cli_dest_inventory_matches_live_parsers`.

Happy to reclassify or drop either if a maintainer prefers a different home.

PR #1473, opened 2026-06-21
[bugfix] Fix performance component timing extraction

### LINKED ISSUE #1377: [Bug] Downstream perf CI stage metric key mapping bug
### Describe the bug

https://github.com/hao-ai-lab/FastVideo/pull/1337 introduced end to end latency and throughput regression along with introducing perf CI stage metric key mapping bug

### Reproduction

/test performance on any pr

### Environment

any version.

Satyam-53:
## Purpose

  Fixes #1377.

  Restore downstream performance CI component timings after pipeline stage logging moved to pipeline-
  unique stage keys.

  ## Changes

  - Map performance stage timings using each stage entry’s `stage_class`, with fallback to legacy class-
  name stage keys.
  - Add focused tests for real `PipelineLoggingInfo`, pipeline-keyed logs, legacy logs, duplicate
  denoising accumulation, and unmapped stages.
  - Update performance benchmark docs to describe pipeline-unique stage logging and component metric
  extraction.

  ## Test Plan

  ```bash
  python -m pytest \
    fastvideo/tests/performance/test_inference_performance_component_times.py \
    fastvideo/tests/performance/test_compare_baseline_policy.py \
    fastvideo/tests/performance/test_dashboard_service.py \
    fastvideo/tests/performance/test_dashboard_api.py \
    -q
```

 ```bash
  pre-commit run --files \
    docs/contributing/performance_benchmarks.md \
    fastvideo/tests/performance/test_inference_performance.py \
    fastvideo/tests/performance/test_inference_performance_component_times.py
```

  ## Test Results
  <details>

  ........................                                                 [100%]
  24 passed in 0.96s

  yapf.................................................(no files to check)Skipped
  ruff (legacy alias)..................................(no files to check)Skipped
  PyMarkdown...............................................................Passed
  Lint GitHub Actions workflow files...................(no files to check)Skipped
  mypy.................................................(no files to check)Skipped
  Check for spaces in all filenames........................................Passed
  Suggestion...............................................................Passed
  </details>

  ## Checklist

  - [x] I ran pre-commit on changed files and fixed all issues

### DIFF
--- a/docs/contributing/performance_benchmarks.md
+++ b/docs/contributing/performance_benchmarks.md
@@ -105,11 +105,15 @@ Each benchmark records six metrics:
 
 `test_inference_performance.py` temporarily sets `FASTVIDEO_STAGE_LOGGING=1`
 while it runs so pipeline stage execution times are available in
-`generate_video(...).logging_info`. It maps `TextEncodingStage` to
-`text_encoder_time_s`, `DenoisingStage` and `DmdDenoisingStage` to
-`dit_time_s`, and `DecodingStage` to `vae_decode_time_s`. If a pipeline does
-not report one of those stages, that component metric is stored as `null` and
-is skipped by the static threshold and rolling baseline checks.
+`generate_video(...).logging_info`. Stage logs use pipeline-unique keys such as
+`prompt_encoding_stage` so duplicate stage classes do not collide; each entry's
+`stage_class` field is mapped to component metrics. The extractor maps
+`TextEncodingStage` to `text_encoder_time_s`, `DenoisingStage` and
+`DmdDenoisingStage` to `dit_time_s`, and `DecodingStage` to
+`vae_decode_time_s`, with a fallback for older logs that used the class name as
+the stage key. If a pipeline does not report one of those stages, that
+component metric is stored as `null` and is skipped by the static threshold and
+rolling baseline checks.
 
 ## The two gates
 
--- a/docs/contributing/performance_benchmarks.md
+++ b/docs/contributing/performance_benchmarks.md
@@ -106,14 +106,15 @@ Each benchmark records six metrics:
 `test_inference_performance.py` temporarily sets `FASTVIDEO_STAGE_LOGGING=1`
 while it runs so pipeline stage execution times are available in
 `generate_video(...).logging_info`. Stage logs use pipeline-unique keys such as
-`prompt_encoding_stage` so duplicate stage classes do not collide; each entry's
-`stage_class` field is mapped to component metrics. The extractor maps
-`TextEncodingStage` to `text_encoder_time_s`, `DenoisingStage` and
-`DmdDenoisingStage` to `dit_time_s`, and `DecodingStage` to
+`prompt_encoding_stage` so duplicate stage classes do not collide. For
+`PipelineStage` entries, the extractor maps the `stage_class` field:
+`TextEncodingStage` maps to `text_encoder_time_s`, `DenoisingStage` and
+`DmdDenoisingStage` map to `dit_time_s`, and `DecodingStage` maps to
 `vae_decode_time_s`, with a fallback for older logs that used the class name as
-the stage key. If a pipeline does not report one of those stages, that
-component metric is stored as `null` and is skipped by the static threshold and
-rolling baseline checks.
+the stage key. Generator-side timings such as `PostDecodeFrameProcessStage`,
+`VideoSaveStage`, and `AudioMuxStage` are intentionally ignored. If a pipeline
+does not report one of the mapped stages, that component metric is stored as
+`null` and is skipped by the static threshold and rolling baseline checks.
 
 ## The two gates
 
--- a/fastvideo/tests/performance/test_inference_performance.py
+++ b/fastvideo/tests/performance/test_inference_performance.py
@@ -98,10 +98,12 @@ def _extract_component_times(result: dict) -> dict[str, float | None]:
         return component_times
     logger.info("Discovered pipeline stages: %s", list(stages.keys()))
     for stage_name, stage_data in stages.items():
-        metric_key = STAGE_METRIC_MAP.get(stage_name)
+        stage_class = stage_data.get("stage_class", stage_name)
+        metric_key = STAGE_METRIC_MAP.get(stage_class)
         if metric_key is None:
-            logger.debug("Unmapped stage '%s' (%.3fs)",
+            logger.debug("Unmapped stage '%s' class '%s' (%.3fs)",
                          stage_name,
+                         stage_class,
                          stage_data.get("execution_time", 0))
             continue
         elapsed = stage_data.get("execution_time")
--- a/fastvideo/tests/performance/test_inference_performance.py
+++ b/fastvideo/tests/performance/test_inference_performance.py
@@ -98,6 +98,9 @@ def _extract_component_times(result: dict) -> dict[str, float | None]:
         return component_times
     logger.info("Discovered pipeline stages: %s", list(stages.keys()))
     for stage_name, stage_data in stages.items():
+        if not isinstance(stage_data, Mapping):
+            logger.debug("Skipping malformed stage '%s' data: %r", stage_name, stage_data)
+            continue
         stage_class = stage_data.get("stage_class", stage_name)
         metric_key = STAGE_METRIC_MAP.get(stage_class)
         if metric_key is None:
--- a/fastvideo/tests/performance/test_inference_performance_component_times.py
+++ b/fastvideo/tests/performance/test_inference_performance_component_times.py
@@ -17,6 +17,8 @@ def test_extract_component_times_handles_pipeline_logging_info_object():
 
 
 def test_extract_component_times_uses_stage_class_for_pipeline_stage_keys():
+    # Regression guard for #1377: pre-fix code looked up the pipeline stage key
+    # and returned all component metrics as None for this shape.
     result = {
         "logging_info": {
             "stages": {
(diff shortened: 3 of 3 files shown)


SolitaryThinker:
**Hi @Satyam-53** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

## TL;DR

This looks ship-ready: the extractor now reads `stage_class` from each pipeline-unique stage entry and falls back to the old class-name stage key shape, which addresses #1377's component-metric attribution break. The new focused tests cover the important compatibility paths, including real `PipelineLoggingInfo`, new keyed logs, legacy logs, duplicate denoising accumulation, and unmapped stages.

## Verdict: **approve**

- S0 (blockers): 0
- S1 (must-fix): 0
- S2 (should-fix; surfaced if `persistent` or `important`): 0

---

## What's good

- The fix is narrowly scoped to the performance CI extraction path: `stage_data.get("stage_class", stage_name)` preserves new pipeline-unique stage keys while keeping legacy class-name logs working.
- Duplicate component classes are handled by summing into the existing component metric instead of overwriting, so repeated denoising stages still contribute to `dit_time_s` correctly.
- Unmapped stages are skipped without affecting mapped component metrics, which keeps additional pipeline bookkeeping stages from breaking performance reports.
- The docs now describe the same `stage_class`-based extraction and legacy fallback implemented by the test helper, so future benchmark debugging has the right mental model.

One robustness nit from the bot review remains technically true: malformed `stages` entries that are not dict-like would still raise on `.get()`. I do not think that should block this PR because the producer contract uses `PipelineLoggingInfo.stages: OrderedDict[str, dict[str, Any]]`, and the legacy/new shapes covered here both use dict stage entries.

---

fastvideo/tests/performance/test_inference_performance.py
```diff
@@ -98,10 +98,12 @@ def _extract_component_times(result: dict) -> dict[str, float | None]:
         return component_times
     logger.info("Discovered pipeline stages: %s", list(stages.keys()))
     for stage_name, stage_data in stages.items():
-        metric_key = STAGE_METRIC_MAP.get(stage_name)
+        stage_class = stage_data.get("stage_class", stage_name)
+        metric_key = STAGE_METRIC_MAP.get(stage_class)
```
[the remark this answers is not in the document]
Satyam-53: addressed.

fastvideo/tests/performance/test_inference_performance.py
```diff
@@ -98,10 +98,15 @@ def _extract_component_times(result: dict) -> dict[str, float | None]:
         return component_times
     logger.info("Discovered pipeline stages: %s", list(stages.keys()))
     for stage_name, stage_data in stages.items():
-        metric_key = STAGE_METRIC_MAP.get(stage_name)
+        if not isinstance(stage_data, Mapping):
+            logger.debug("Skipping malformed stage '%s' data: %r", stage_name, stage_data)
+            continue
+        stage_class = stage_data.get("stage_class", stage_name)
+        metric_key = STAGE_METRIC_MAP.get(stage_class)
```
SolitaryThinker, line 105: **NIT (maintainability):** `STAGE_METRIC_MAP.get(stage_class)` is an exact match on `__class__.__name__`. Correct for the current Wan benchmark (base `DenoisingStage`/`TextEncodingStage`/`DecodingStage`), but a future benchmark whose pipeline uses a `DenoisingStage` subclass (e.g. `CosmosDenoisingStage`, `Gen3CDenoisingStage`, `CausalDenoisingStage`) would silently map to `dit_time_s=None` and skip the DiT gate — the same attribution gap this PR fixes. You already had to add `DmdDenoisingStage` explicitly, which is the tell. Consider matching by `isinstance`/base class (or normalizing a subclass to its component base) so a new config doesn't need a map edit. Not blocking — no current benchmark hits it.

docs/contributing/performance_benchmarks.md
```diff
@@ -105,11 +105,15 @@ Each benchmark records six metrics:
...
 `test_inference_performance.py` temporarily sets `FASTVIDEO_STAGE_LOGGING=1`
 while it runs so pipeline stage execution times are available in
-`generate_video(...).logging_info`. It maps `TextEncodingStage` to
-`text_encoder_time_s`, `DenoisingStage` and `DmdDenoisingStage` to
-`dit_time_s`, and `DecodingStage` to `vae_decode_time_s`. If a pipeline does
-not report one of those stages, that component metric is stored as `null` and
-is skipped by the static threshold and rolling baseline checks.
+`generate_video(...).logging_info`. Stage logs use pipeline-unique keys such as
+`prompt_encoding_stage` so duplicate stage classes do not collide; each entry's
+`stage_class` field is mapped to component metrics. The extractor maps
```
SolitaryThinker: **NIT (docs):** "each entry's `stage_class` field is mapped" slightly overstates it — only the `PipelineStage`-driven entries carry `stage_class`. The generator post-processing entries (`PostDecodeFrameProcessStage`, `VideoSaveStage`, `AudioMuxStage`) are logged via `add_stage_execution_time(...)` only and have no `stage_class`. Harmless (they fall through to the `stage_name` lookup, miss the map, and are skipped — exactly what the `ignores_unmapped_stages` test asserts), but the sentence could note that those entries are intentionally ignored.

fastvideo/tests/performance/test_inference_performance_component_times.py
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    assert _extract_component_times(result) == {
+        "text_encoder_time_s": 1.2,
+        "dit_time_s": 3.4,
+        "vae_decode_time_s": 0.8,
+    }
+
+
+def test_extract_component_times_keeps_legacy_class_name_keys():
```
SolitaryThinker, line 48: **NIT (tests):** Nice coverage. One note: this legacy-class-name-keys case (and `ignores_unmapped_stages`) passes identically on the pre-fix code, so it characterizes the backward-compat fallback rather than guarding the #1377 regression. The actual regression guard is `test_extract_component_times_uses_stage_class_for_pipeline_stage_keys` (it goes all-`None` on the old `stage_name` lookup) — worth a one-line comment marking which test pins the fix vs which document fallback behavior, so a future refactor knows which one must stay red on a revert.

SolitaryThinker (commented):
Hi @Satyam-53, thanks for the PR!

Traced the fix end-to-end and it correctly restores component-timing attribution for the live benchmark. Since #1337 the stages dict is keyed by pipeline-unique `stage_key`, so mapping each entry's stored `stage_class` (with the `stage_name` fallback) is the right fix. Confirmed the only benchmark config (`wan-t2v-1.3b.json` → `Wan-AI/Wan2.1-T2V-1.3B-Diffusers` → base `WanPipeline`) registers the base `TextEncodingStage`/`DenoisingStage`/`DecodingStage`, whose `__class__.__name__` match `STAGE_METRIC_MAP` exactly, so all three metrics populate. The `isinstance(stage_data, Mapping)` guard and the duplicate-class accumulation both look right, and the new tests cover the real `PipelineLoggingInfo` object plus the keyed/legacy/duplicate/unmapped/malformed shapes.

No blockers. A few small notes inline, all NIT-level.

FYI the `build` check is red on a pre-existing broken docs link (`docs/training/examples/wan_t2v_1.3B_mixkit.md:59`), unrelated to this PR.

## Dropped (considered, not flagged)
- New test only runs in the performance lane (`pytest ./fastvideo/tests/performance`), not per-PR fastcheck — dropped: `_extract_component_times` is perf-CI tooling that only executes in that lane anyway, so co-locating its guard there is correct placement, not a coverage gap.

fastvideo/tests/performance/test_inference_performance_component_times.py
Satyam-53: Addressed. Added comments to distinguish the actual #1377 regression guard from the tests that document legacy fallback and intentionally ignored generator-side timings.

docs/contributing/performance_benchmarks.md
Satyam-53: Addressed. Updated the docs to say that `stage_class` applies to `PipelineStage` entries, while
  generator-side timings like `PostDecodeFrameProcessStage`, `VideoSaveStage`, and `AudioMuxStage` are intentionally ignored.

fastvideo/tests/performance/test_inference_performance.py
Satyam-53: Left unchanged for now. This PR targets the current Wan performance benchmark, whose stages match the existing map exactly. Subclass normalization would broaden behavior beyond this bugfix, so I’d handle  that separately if/when we add a benchmark using those subclasses. - Opened an issue to track this- #1475

Satyam-53 (commented):
Addressed nits.

SolitaryThinker (approved):
Hi @Satyam-53, thanks for the PR!

The earlier nits are all handled: NIT2 (docs) and NIT3 (test comments) are resolved, and NIT1 (subclass exact-name match) is reasonably deferred to #1475 — agreed it's out of scope for this bugfix.

Re-verified the fix still holds at 52eab31b (it's unchanged since the last review): the only benchmark config (`wan-t2v-1.3b.json` → base `WanPipeline`) registers base `TextEncodingStage`/`DenoisingStage`/`DecodingStage`, whose `__class__.__name__` match `STAGE_METRIC_MAP` exactly, so all three component metrics still populate. The docs rewrite is accurate — every `PipelineStage` entry gets a `stage_class` (stages/base.py:157), and the named generator-side stages (`PostDecodeFrameProcessStage`/`VideoSaveStage`/`AudioMuxStage`) are written into the stages dict but correctly fall through as unmapped, so "intentionally ignored" is the right framing. The added test comments cleanly separate the #1377 regression guard from the legacy-fallback and generator-side-ignore cases.

LGTM.

FYI the `build` check is still red on the same pre-existing broken docs link (`docs/training/examples/wan_t2v_1.3B_mixkit.md:59`), unrelated to this PR.

## Dropped (considered, not flagged)
- Nothing new this pass. The prior note that the component-times test only runs in the performance lane (not per-PR fastcheck) still stands as a non-issue — `_extract_component_times` is perf-CI tooling that only executes in that lane, so co-locating its guard there is correct placement.

PR #1479, opened 2026-06-23
[feat]: hard-fail on missing attention backends + force VSA for FastWan

SolitaryThinker:
## Problem

Two silent footguns in attention-backend selection:

**(1) Missing backends silently fall through to FlashAttn.** When a user sets `FASTVIDEO_ATTENTION_BACKEND` to `SAGE_ATTN`, `SAGE_ATTN_THREE`, or `ATTN_QAT_INFER` but the corresponding package is not installed, the dispatch in `fastvideo/platforms/cuda.py` logs a warning and falls through to FlashAttn. The user believes they are running the requested backend but they are not — outputs differ from what was asked for, with no error surface.

**(2) FastWan requires VSA for correctness.** The FastWan model family is sparse-distilled with VSA and produces incorrect outputs at inference without it. The README tells users to set `os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "VIDEO_SPARSE_ATTN"` before `from_pretrained` — another silent footgun if the user forgets or the env var is overridden elsewhere.

### `fastvideo/platforms/cuda.py` — hard-raise on missing explicit backends

| Backend | Before | After |
|---|---|---|
| `SAGE_ATTN` | warn + fall through to FlashAttn | `raise ImportError` with `uv pip install sageattention` hint |
| `SAGE_ATTN_THREE` | warn + fall through to FlashAttn | `raise ImportError` with `uv pip install sageattn3` hint |
| `ATTN_QAT_INFER` (kernel missing) | warn + fall through to FlashAttn | `raise ImportError` with docs URL |
| `ATTN_QAT_INFER` (built but unavailable) | warn + fall through to FlashAttn | `raise ImportError` with docs URL |
| `VIDEO_SPARSE_ATTN` | already raises | unchanged |
| `ATTN_QAT_TRAIN` | already raises (PR #1459) | unchanged |

### `fastvideo/configs/pipelines/wan.py` + `selector.py` + `models/dits/wanvideo.py` — force VSA for FastWan

- `FastWan2_1_T2V_480P_Config.__post_init__` and `FastWan2_2_TI2V_5B_Config.__post_init__` (and the `Wan2_2_TI2V_5B_Config` parent both subclass) now call `force_fastwan_attention_backend()`, which sets the selector's global force to `VIDEO_SPARSE_ATTN`.
- `fastvideo/attention/selector.py:global_force_attn_backend` now clears the `lru_cache` on the selector when the force changes, so a stale FlashAttn selection cannot survive a subsequent FastWan force.
- `WanTransformer3DModel` in `fastvideo/models/dits/wanvideo.py` now reads the global-forced backend before falling back to `FASTVIDEO_ATTENTION_BACKEND` when choosing between `WanTransformerBlock_VSA` and `WanTransformerBlock`.
- If the VSA kernel (`fastvideo_kernel.video_sparse_attn`) is unavailable, the existing `VIDEO_SPARSE_ATTN` path in `cuda.py` already raises `ImportError` — that error now surfaces for FastWan loads instead of silently falling through.

**Env-var precedence**: FastWan's config-side force **overrides** `FASTVIDEO_ATTENTION_BACKEND`. Rationale: the user explicitly chose FastWan, the model REQUIRES VSA for correctness, an env-var override would silently produce incorrect outputs. Other model families are unaffected — env still wins for non-FastWan configs.

## Test plan

- `pre-commit run --files fastvideo/attention/selector.py fastvideo/configs/pipelines/wan.py fastvideo/models/dits/wanvideo.py fastvideo/platforms/cuda.py` — clean.
- `python -m py_compile` on all four changed files — clean.
- Manual smoke (worker did this in a venv without the optional packages):
  - Simulated missing `sageattention` -> `SAGE_ATTN` request raises `ImportError` with the install hint. ✅
  - Simulated missing `sageattn3` -> `SAGE_ATTN_THREE` request raises `ImportError` with the install hint. ✅
  - Simulated missing `fastvideo_kernel` -> `VIDEO_SPARSE_ATTN` request raises `ImportError` with the docs URL (existing behavior, verified unchanged). ✅
  - Simulated unavailable `attn_qat_infer` -> `ATTN_QAT_INFER` request raises `ImportError` with the docs URL. ✅
  - Instantiated `FastWan2_1_T2V_480P_Config()` and `FastWan2_2_TI2V_5B_Config()` -> `get_global_forced_attn_backend()` returns `VIDEO_SPARSE_ATTN`. ✅

## Notes

- This is a **breaking change** for any user currently relying on the silent fall-through to FlashAttn when they explicitly set `FASTVIDEO_ATTENTION_BACKEND` to one of the affected enums. The previous behavior was a footgun; the new behavior is consistent with how `VIDEO_SPARSE_ATTN`, `BSA_ATTN`, `VMOBA_ATTN`, `SLA_ATTN`, `SAGE_SLA_ATTN`, and (post #1459) `ATTN_QAT_TRAIN` already behave.
- The FastWan VSA force does not change the README's example — setting `os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "VIDEO_SPARSE_ATTN"` before `from_pretrained` is now redundant for FastWan loads (the config does it) but not harmful.

### DIFF
--- a/fastvideo/attention/selector.py
+++ b/fastvideo/attention/selector.py
@@ -69,6 +69,7 @@ def global_force_attn_backend(attn_backend: AttentionBackendEnum | None) -> None
     '''
     global forced_attn_backend
     forced_attn_backend = attn_backend
+    _cached_get_attn_backend.cache_clear()
 
 
 def get_global_forced_attn_backend() -> AttentionBackendEnum | None:
--- a/fastvideo/configs/models/dits/wanvideo.py
+++ b/fastvideo/configs/models/dits/wanvideo.py
@@ -2,6 +2,7 @@
 from dataclasses import dataclass, field
 
 from fastvideo.configs.models.dits.base import DiTArchConfig, DiTConfig
+from fastvideo.platforms.interface import AttentionBackendEnum
 
 
 def is_blocks(n: str, m) -> bool:
@@ -92,6 +93,7 @@ class WanVideoArchConfig(DiTArchConfig):
 
 @dataclass
 class WanVideoConfig(DiTConfig):
-    arch_config: DiTArchConfig = field(default_factory=WanVideoArchConfig)
+    arch_config: WanVideoArchConfig = field(default_factory=WanVideoArchConfig)
 
     prefix: str = "Wan"
+    required_attention_backend: AttentionBackendEnum | None = None
--- a/fastvideo/configs/pipelines/wan.py
+++ b/fastvideo/configs/pipelines/wan.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 
 import torch
 
-from fastvideo.configs.models import DiTConfig, EncoderConfig, VAEConfig
+from fastvideo.configs.models import EncoderConfig, VAEConfig
 from fastvideo.configs.models.dits import WanVideoConfig
 from fastvideo.configs.models.dits.wanvideo import WanVideoArchConfig
 from fastvideo.configs.models.encoders import (BaseEncoderOutput, CLIPVisionConfig, T5Config,
@@ -12,6 +12,9 @@ from fastvideo.configs.models.encoders import (BaseEncoderOutput, CLIPVisionConf
 from fastvideo.configs.models.vaes import WanVAEConfig
 from fastvideo.configs.models.vaes.wanvae import WanVAEArchConfig
 from fastvideo.configs.pipelines.base import PipelineConfig
+from fastvideo.platforms import AttentionBackendEnum
+
+FASTWAN_REQUIRED_ATTENTION_BACKEND = AttentionBackendEnum.VIDEO_SPARSE_ATTN
 
 
 def t5_postprocess_text(outputs: BaseEncoderOutput) -> torch.Tensor:
@@ -31,7 +34,7 @@ class WanT2V480PConfig(PipelineConfig):
 
     # WanConfig-specific parameters with defaults
     # DiT
-    dit_config: DiTConfig = field(default_factory=WanVideoConfig)
+    dit_config: WanVideoConfig = field(default_factory=WanVideoConfig)
     # VAE
     vae_config: VAEConfig = field(default_factory=WanVAEConfig)
     vae_tiling: bool = False
@@ -55,7 +58,7 @@ class WanT2V480PConfig(PipelineConfig):
 
     # WanConfig-specific added parameters
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         self.vae_config.load_encoder = False
         self.vae_config.load_decoder = True
 
@@ -114,6 +117,10 @@ class FastWan2_1_T2V_480P_Config(WanT2V480PConfig):
     flow_shift: float | None = 8.0
     dmd_denoising_steps: list[int] | None = field(default_factory=lambda: [1000, 757, 522])
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.dit_config.required_attention_backend = FASTWAN_REQUIRED_ATTENTION_BACKEND
+
 
 @dataclass
 class Wan2_2_TI2V_5B_Config(WanT2V480PConfig):
@@ -132,7 +139,7 @@ class Wan2_2_TI2V_5B_Config(WanT2V480PConfig):
 class LucyEditDevConfig(Wan2_2_TI2V_5B_Config):
     """Configuration for Decart Lucy Edit Dev video editing."""
 
-    dit_config: DiTConfig = field(default_factory=lambda: WanVideoConfig(arch_config=WanVideoArchConfig(
+    dit_config: WanVideoConfig = field(default_factory=lambda: WanVideoConfig(arch_config=WanVideoArchConfig(
         num_attention_heads=24,
         in_channels=96,
         out_channels=48,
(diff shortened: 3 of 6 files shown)


SolitaryThinker:
**Hi @SolitaryThinker** — automated review from Gob, one of @SolitaryThinker's AI reviewers. Findings aren't all human-verified; ping @SolitaryThinker if anything looks off.

## TL;DR

The mechanics check out: cuda.py hard-raises are structurally clean and the selector cache-clear is correct (`@cache` exposes `.cache_clear()`). The two design-shape concerns gemini surfaced are both real — independently verified — and worth addressing before merge or as immediate follow-up. No S1 blockers.

## Verdict

**approve-with-followup**

## Severity counts

`S0=0 S1=0 S2=2` (S3 not shown — see archived review)

## Findings

### S2-1 — `FastWan` `__post_init__` writes process-wide selector state with no symmetric unset

**What.** `FastWan2_1_T2V_480P_Config.__post_init__` and `FastWan2_2_TI2V_5B_Config.__post_init__` call `force_fastwan_attention_backend()`, which mutates the module-level `forced_attn_backend` in `fastvideo/attention/selector.py`. There is no symmetric `unforce`, and sibling non-FastWan Configs (`WanT2V480PConfig`, `Wan2_2_T2V_A14B_Config`, `Wan2_2_I2V_A14B_Config`, `LucyEditDevConfig`) do not clear the force in their own `__post_init__`.

**Why it matters.** The PR body says "Other model families are unaffected — env still wins for non-FastWan configs." That contract holds in a fresh process, but not when a FastWan Config is instantiated first and a non-FastWan Config second in the same process — the second one inherits the VSA force. Realistic failure modes:

- Pytest fixtures that build a FastWan Config leak VSA into sibling tests that build non-FastWan Configs (pytest does not fork by default).
- Notebooks / multi-pipeline scripts that switch model families silently route the second pipeline through VSA selection. `WanTransformer3DModel.__init__` will then pick `WanTransformerBlock_VSA` for a non-FastWan checkpoint, producing wrong outputs or a load-time failure on a checkpoint that should have worked.

**Fix.** Three options, listed cheapest to cleanest:

1. Document the sticky semantics on `force_fastwan_attention_backend()` and on `forced_attn_backend`, and have non-FastWan Wan Configs defensively call `global_force_attn_backend(None)` in their own `__post_init__`. Spreads the side-effect.
2. Move the force out of `__post_init__` to the FastWan pipeline-build site and wrap it in `global_force_attn_backend_context_manager(VIDEO_SPARSE_ATTN)` (already exists at `selector.py:127`). Scoped, no leak.
3. Store `required_attention_backend` on the Config dataclass and have `WanTransformer3DModel.__init__` consult `pipeline_config.required_attention_backend` directly, eliminating the global write. Bigger refactor.

Option 2 is the closest fit to the existing codebase since the context manager was designed for this.

**Evidence.**
- `fastvideo/attention/selector.py:56,70-72` — module-level global, no unset.
- `fastvideo/configs/pipelines/wan.py:126-128` (`FastWan2_1_T2V_480P_Config`).
- `fastvideo/configs/pipelines/wan.py:284-286` (`FastWan2_2_TI2V_5B_Config`).
- `fastvideo/configs/pipelines/wan.py:67-69, 298-309` — sibling Configs that don't clear.

### S2-2 — New `ImportError` messages hard-code `FASTVIDEO_ATTENTION_BACKEND=...` as the only entry point

**What.** The three new messages in `cuda.py` open with phrasing like *"SageAttention backend was explicitly requested via `FASTVIDEO_ATTENTION_BACKEND=SAGE_ATTN`"*. But this very PR adds a second request path: `global_force_attn_backend()` (used by `force_fastwan_attention_backend()`). A future Config that programmatically forces SAGE_ATTN without an env var will hit a misleading error message telling the user to check an env var they never set.

**Why it matters.** Today the only programmatic-force caller is FastWan→VSA, and the VSA message at `cuda.py:182-186` is already generic, so users don't hit the misleading wording in the wild. But the precedent matters — every future programmatic force will inherit the same misleading message shape.

**Fix.** Generalize the wording, mirroring the existing VSA / ATTN_QAT_TRAIN style. E.g.:

> "SageAttention backend was explicitly requested (via `FASTVIDEO_ATTENTION_BACKEND=SAGE_ATTN` or `global_force_attn_backend(AttentionBackendEnum.SAGE_ATTN)`) but the `sageattention` package is not installed. Install it with: `uv pip install sageattention`."

Same shape for `SAGE_ATTN_THREE` and the two `ATTN_QAT_INFER` paths.

**Evidence.**
- `fastvideo/platforms/cuda.py:129-132` (SAGE_ATTN).
- `fastvideo/platforms/cuda.py:143-146` (SAGE_ATTN_THREE).
- `fastvideo/platforms/cuda.py:152-155, 159-162` (ATTN_QAT_INFER, both paths).

SolitaryThinker:
**Update — S2-1 (FastWan VSA force bleed) addressed** at `20331775b915a20761ec01738864e751dab66e4f`.

Per the directive to force VSA *only* for FastWan models and never for base Wan models, the process-global force has been removed and replaced with a per-config requirement:

- **`force_fastwan_attention_backend()` global write removed** from both FastWan `__post_init__` methods. No Config instantiation mutates process-wide selector state anymore.
- **`WanVideoConfig.required_attention_backend`** new field; `FastWan2_1_T2V_480P_Config` and `FastWan2_2_TI2V_5B_Config` set it to `VIDEO_SPARSE_ATTN` on their own `dit_config`.
- **`WanTransformer3DModel.__init__`** reads `config.required_attention_backend` first (before the selector global / env-var) when choosing `WanTransformerBlock_VSA` vs `WanTransformerBlock`.
- **FastWan block construction is wrapped in `global_force_attn_backend_context_manager(VIDEO_SPARSE_ATTN)`** so `get_attn_backend` resolves VSA during the build — preserving the hard-`ImportError` on a missing VSA kernel — then reverts on exit. Base Wan models built later in the same process see no force.

Net effect: a `FastWan2.1-T2V` Config instantiated before a base `WanT2V480PConfig` in the same process no longer leaks VSA into the base model (the original bleed). Regression test added at `fastvideo/tests/attention/test_fastwan_vsa_scoping.py` (2 passing) asserting (a) FastWan carries the VSA requirement on its own dit_config, (b) base Wan does not, and (c) `get_global_forced_attn_backend()` is `None` after FastWan Config creation.

S2-2 (ImportError-message wording mentioning only the env-var entry point) is not addressed in this push — left to author judgment as a non-blocking follow-up.

PR #1488, opened 2026-06-24
[bugfix] nvfp4: cast fp32 inputs to bf16 instead of asserting

Mister-Raggs:
**Problem.** `NVFP4QuantizeMethod.quantize_input()` and `apply()` assert bf16/fp16 inputs and crash on the fp32 pre-attention norm output when running **without `torch.compile`** (eager mode), where the norm isn't fused back to bf16:

```
AssertionError: only allow bf16/fp16 inputs to fp4 linear, got torch.float32
```

Repro: any `NVFP4Config` inference run in eager mode — e.g. LTX2-distilled on a DGX Spark (sm_121), no compile — fails on the first denoising step.

**Fix.** Cast fp32 inputs to bf16 instead of asserting, in both code paths. This is safe and consistent: the FP4 linear already emits bf16 unconditionally (`_mm_fp4` out dtype), and the sibling `fastvideo/layers/fp4linear.py` already casts non-bf16/fp16 inputs the same way. No change to the compiled path (inputs are already bf16 there).

**Test.** With the fix, LTX2-distilled runs end-to-end under `NVFP4Config` in eager mode on sm_121; output is visually unchanged vs bf16 and the FP4 linears measurably speed up denoising. No new behavior on the existing compiled path. `pre-commit` (yapf/ruff/codespell) passes on the changed file.

### DIFF
--- a/fastvideo/layers/quantization/nvfp4_config.py
+++ b/fastvideo/layers/quantization/nvfp4_config.py
@@ -285,8 +285,12 @@ class NVFP4QuantizeMethod(QuantizeMethodBase):
 
     def quantize_input(self, x: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
         SfLayout, _, _ = _require_flashinfer()
-        assert x.dtype == torch.bfloat16 or x.dtype == torch.float16, (
-            f"only allow bf16/fp16 inputs to fp4 linear, got {x.dtype}")
+        # The pre-attention norm can emit fp32 (e.g. in eager mode, without the
+        # torch.compile fusion that keeps it bf16). FP4 linear emits bf16 regardless
+        # (see apply(): _mm_fp4 out dtype is bf16), so cast fp32 inputs to bf16
+        # instead of failing — matching the sibling fastvideo/layers/fp4linear.py.
+        if x.dtype not in (torch.bfloat16, torch.float16):
+            x = x.to(torch.bfloat16)
         x_2d = x.view(-1, x.shape[-1])
         x_fp4, x_scale = _nvfp4_quantize(
             x_2d,
@@ -332,8 +336,8 @@ class NVFP4QuantizeMethod(QuantizeMethodBase):
             if x_scale.dim() > 2:
                 x_scale = x_scale.view(-1, x_scale.shape[-1])
         else:
-            assert x.dtype == torch.bfloat16 or x.dtype == torch.float16, (
-                f"only allow bf16/fp16 inputs to fp4 linear, got {x.dtype}")
+            if x.dtype not in (torch.bfloat16, torch.float16):
+                x = x.to(torch.bfloat16)
             x = x.view(-1, x.shape[-1])
             x_global_sf = self.x_global_sf
             x_fp4, x_scale = _nvfp4_quantize(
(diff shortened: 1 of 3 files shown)


alexzms (approved):
Thanks for contribution, LGTM

SolitaryThinker:
Maintainer update: rebased onto current main (was 19 behind — the two red full-suite lanes don't touch this diff and should clear on the rebase) and added one commit on top: the root cause of the fp32 you were seeing.

torch 2.12 added `rms_norm` to autocast's fp32 cast policy, so `_rms_norm_dispatch` in `ltx2.py` started returning fp32 under autocast(bf16) — #1447 fixed this exact thing for the q/k norms one function over, but missed this dispatch. That fp32 flows through the ada modulation into your `quantize_input` call sites, which is why the assert fired. Added the same `.to(x.dtype)` there (5c138673).

Your coercion helper stays — it's good defense-in-depth and matches what `fp4linear.py` already does. But without the norm fix, eager runs would compute the norm+modulation in fp32 while the QuACK path uses bf16, which is a parity drift for every consumer, not just FP4.

@alexzms heads-up: your LGTM + /merge predates this commit — it's a one-liner mirroring b36d0ef0, but yell if you want it split out.

Mister-Raggs:
The pre-commit red isn't from this PR. #1545 landed `fastvideo/performance/hf_store.py` and `metric_policy.py` with a yapf 0.43.0 violation, so every open PR is failing `pre-commit --all-files` on it right now. #1551 clears it by moving those files under `tests/` (excluded from the hook), so once that or a quick `yapf -i` on main lands, this should go green on a rebase. Nothing in this diff's own files is malformed.

SolitaryThinker:
Rebased onto main to pick up the FA4/cutlass pin fix (and, once merged, the perf-lane CPU reservation) — no content changes (range-diff clean).

SolitaryThinker:

SolitaryThinker:
Pure rebase onto current main to pick up #1565 (memory cleanup between train-framework tests) and #1566 (CI lane CPU/memory reservations); no content changes (range-diff all `=`). Old head 63400778 -> new head 87737d10.

SolitaryThinker:
Reverted the maintainer-added ltx2 dtype hunk (87737d1): it changes LTX2 numerics vs the seeded SSIM references on CI's torch 2.12 — build 4189's LTX2-Distilled SSIM failure with sibling PRs green on the same test pins it to that commit. The nvfp4 fix stands alone and is lane-clean. The dtype restore will return in its own PR together with a decision on re-seeding the references.

Mister-Raggs:
@SolitaryThinker — the perf failure here looks like the torch 2.12 `rms_norm` autocast upcast from your revert note, not this PR.

Signal that points that way:
- Only the RMSNorm+attention stages regress — text encoder +77%, DiT +79% — while VAE decode (conv) got *faster*. Consistent with fp32 leaking out of the norms into attention.
- It's identical across commits: text encoder was 3.595s on my Jun-29 run (branch had only the nvfp4 commits then) and 3.596s now. Wan never instantiates `NVFP4Config`, so nothing in this diff runs in that benchmark.
- Baseline (2.032s text enc / 5.235s DiT) was recorded before the 2.12 bump.

Does the baseline need re-recording under 2.12, or is fixing the upcast at the norm first?

Issue #1379, opened 2026-05-21
RMSNorm.forward_native doesn't unwrap DTensor weight under FSDP CPU offload


SolitaryThinker:
## Background

This issue tracks an FSDP CPU-offload edge case in `fastvideo/layers/layernorm.py`'s `RMSNorm.forward_native` that was originally bundled into PR #572 by @Edenzzzz as a workaround, but was dropped from that PR's scope to keep it focused on distributed weight loading. The underlying bug still exists; this issue tracks it so it doesn't get lost.

## The bug

When a model uses FSDP with CPU offload, `RMSNorm.forward_native`'s `self.weight` can remain as a `DTensor` rather than being unwrapped to a local tensor, which then breaks downstream ops. The author's own inline note (from PR #572's original branch state `c2b66087d`) flagged it as a known FSDP workaround that needs upstream investigation:

```python
# TODO(wenxuan): When using CPU offload, FSDP has a bug that
# doesn't unwrap DTensor in final_layer_norm. Report this
weight = self.weight.to_local().to(x.device) if isinstance(self.weight, DTensor) else self.weight
```

## Workaround (saved on the original PR #572 branch)

A 5-line addition to `RMSNorm.forward_native` immediately before `x = x * weight`:

```python
weight = self.weight.to_local().to(x.device) if isinstance(self.weight, DTensor) else self.weight
```

(See PR #572's pre-drop commit `c2b66087d` for the original FastVideo patch; commit `b7c338210` is the scope-drop that removed it before merge.)

## What this issue is asking for

Before re-landing the workaround as a separate PR, we need:

- [ ] **A reliable reproducer**: what FSDP version, model, hardware, and code path triggers this? The original PR didn't ship a regression test.
- [ ] **An upstream issue link** (pytorch/pytorch?): is this a known FSDP-side bug, or specific to how FastVideo wires FSDP + RMSNorm? If it's pytorch-side, the workaround should be retired when upstream lands a fix.
- [ ] **A unit test** that exercises the CPU-offload + DTensor weight path on `RMSNorm.forward_native` so the workaround can ship with a regression gate.
- [ ] **A clear behavioral spec**: does the bug surface only on `final_layer_norm` (per the comment) or anywhere `RMSNorm.forward_native` runs with `has_weight=True` under FSDP CPU offload? The comment says the former; needs verification.

## What this issue is NOT

- This is **not** about `RMSNorm.forward_cuda` — only the `forward_native` path.
[...]

PR #1494, opened 2026-06-25
[feat]: hard-fail on missing/mismatched attention backends; require VSA for FastWan

SolitaryThinker:
Re-opens #1479 from a branch on the main repo (the original was a cross-fork PR). Same content + the design revision discussed on that PR.

## Problem

Two silent footguns in attention-backend selection:

1. **Missing backends silently fall through to FlashAttn.** Setting `FASTVIDEO_ATTENTION_BACKEND` to `SAGE_ATTN`, `SAGE_ATTN_THREE`, or `ATTN_QAT_INFER` when the package isn't installed logged a warning and fell back to FlashAttn — the user runs a different backend than they asked for, with no error surface.
2. **FastWan requires VSA for correctness.** FastWan is sparse-distilled with VSA and produces incorrect output without it, but nothing enforced the backend at load time.

## Change

- **`platforms/cuda.py`** — explicitly-requested-but-unavailable backends (`SAGE_ATTN`, `SAGE_ATTN_THREE`, `ATTN_QAT_INFER`) now `raise ImportError` with an install hint instead of falling through to FlashAttn.
- **`envs.py`** — `FASTVIDEO_ATTENTION_BACKEND` is validated on read; an unknown value raises with the list of valid backends instead of silently auto-selecting.
- **`DiTConfig.required_attention_backend`** (new base field, `None` default) — a model config can declare a backend it is only numerically correct with. Lives next to `_supported_attention_backends` ("what it *can* use" vs "what it *must* use").
- **`attention/selector.py`** — `check_attn_backend_requirement()` resolves the effective backend (global force > env var) and **fails loudly** when it doesn't match a model's requirement.
- **`models/dits/wanvideo.py`** — `WanTransformer3DModel` enforces the requirement via that check instead of forcing the backend during block construction. A transient force only reached block construction and left the denoising stage resolving a different backend (VSA blocks + no VSA metadata → crash at forward); failing loud makes every resolution site read the same env var consistently.
- **`configs/pipelines/wan.py` + `registry.py`** — FastWan configs set `required_attention_backend = VIDEO_SPARSE_ATTN`; a new `FastWan2_2_TI2V_5B_FullAttn_Config` (no requirement) is routed the dense `FastWan2.2-TI2V-5B-FullAttn-Diffusers` checkpoint so it is no longer forced onto VSA.

**Breaking change**: explicitly selecting an uninstalled backend, or loading a FastWan model without `FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN`, now raises instead of silently running FlashAttn. This is consistent with how `VIDEO_SPARSE_ATTN` / `ATTN_QAT_TRAIN` already behave.

## Test plan
- `pre-commit run --files ...` — clean (yapf / ruff / mypy / codespell).
- New tests under `fastvideo/tests/attention/` cover the env validation, the FastWan VSA requirement (incl. the FullAttn config carrying no requirement), and every branch of `check_attn_backend_requirement`.

### DIFF
--- a/docs/inference/support_matrix.md
+++ b/docs/inference/support_matrix.md
@@ -56,7 +56,7 @@ pipeline initialization and sampling.
 | Model Name | HuggingFace Model ID | Resolutions | TeaCache | Sliding Tile Attn (Legacy Branch) | Sage Attn | VSA | BSA |
 |------------|---------------------|-------------|----------|-------------------|-----------|-----|-----|
 | FastWan2.1 T2V 1.3B | `FastVideo/FastWan2.1-T2V-1.3B-Diffusers` | 480P | ⭕ | ⭕ | ⭕ | ✅ | ⭕ |
-| FastWan2.2 TI2V 5B Full Attn* | `FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers` | 720P | ⭕ | ⭕ | ⭕ | ✅ | ⭕ |
+| FastWan2.2 TI2V 5B Full Attn* | `FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers` | 720P | ⭕ | ⭕ | ⭕ | ❌ | ⭕ |
 | Wan2.2 TI2V 5B | `Wan-AI/Wan2.2-TI2V-5B-Diffusers` | 720P | ⭕ | ⭕ | ✅ | ⭕ | ⭕ |
 | DreamX-World 5B Cam | `FastVideo/DreamX-World-5B-Cam-Diffusers` | 480P | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ |
 | DreamX-World 5B AR | `FastVideo/DreamX-World-5B-Diffusers` | 704px1280p | ⭕ | ⭕ | ⭕ | ⭕ | ⭕ |
--- a/examples/inference/gradio/serving/ray_serve_backend.py
+++ b/examples/inference/gradio/serving/ray_serve_backend.py
@@ -127,11 +127,12 @@ def save_image_from_base64(image_data: str, output_dir: str) -> Optional[str]:
 
 
 def setup_model_environment(model_path: str) -> None:
-    # if "fullattn" in model_path.lower():
-    #     os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "FLASH_ATTN"
-    # else:
-    #     os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "VIDEO_SPARSE_ATTN"
-    os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "FLASH_ATTN"
+    from fastvideo.platforms import AttentionBackendEnum
+    from fastvideo.registry import get_pipeline_config_cls_from_name
+
+    pipeline_config = get_pipeline_config_cls_from_name(model_path)()
+    attention_backend = pipeline_config.dit_config.required_attention_backend or AttentionBackendEnum.FLASH_ATTN
+    os.environ["FASTVIDEO_ATTENTION_BACKEND"] = attention_backend.name
     os.environ["FASTVIDEO_STAGE_LOGGING"] = "1"
 
 
--- a/examples/train/configs/fine_tuning/wan/fast_ti2v_fullattn_lora.yaml
+++ b/examples/train/configs/fine_tuning/wan/fast_ti2v_fullattn_lora.yaml
@@ -1,6 +1,8 @@
-# FastWan 2.2 TI2V 5B FullAttn LoRA finetune with VSA.
+# FastWan 2.2 TI2V 5B FullAttn LoRA finetune with dense attention.
 #
 # Notes:
+# - The FullAttn checkpoint has no VSA gate-compression weights. Train its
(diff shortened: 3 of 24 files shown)


SolitaryThinker:
Rebased onto main to pick up the FA4/cutlass pin fix (and, once merged, the perf-lane CPU reservation) — no content changes (range-diff clean).

SolitaryThinker:

SolitaryThinker:
Pure rebase onto current main to pick up #1565 (memory cleanup between train-framework tests) and #1566 (CI lane CPU/memory reservations); no content changes (range-diff all `=`). Old head f41d4a61 -> new head 54def7ae.

SolitaryThinker:
Pushed 7bed2765 fixing a load failure this PR introduced: in the ssim lane's pinned `FASTVIDEO_ATTENTION_BACKEND=FLASH_ATTN`, matrixgame2's CLIP vision encoder (head_size=80) died with "Attention backend FLASH_ATTN was explicitly selected but is incompatible with this layer: SDPA does not support head_size=80". The platform had already substituted SDPA for the FLASH_ATTN pin (FA does not support head 80), but the new selector validation still treated the resolved fallback's incompatibility as a user error against the pin and raised, making any global pin unusable with models carrying aux encoders.

The fix draws the distinction the validation intended: hard-fail only when the explicit selection was actually honored and that very backend's capabilities reject the layer (real user error); when resolution fell back to a different backend (pin outside the layer's supported set, or a platform-level substitution like the head-size SDPA fallback), the layer never participated in the pin and we emit the one-per-cached-resolution warning instead, as with auto-selection.

Tests: updated `test_selector_capability_validation.py` so the hard-fail case models an honored pin, and added two regression tests — pinned FLASH_ATTN with a platform fallback warns instead of raising (the matrixgame2 shape), and a pin outside the layer's supported set falls back with a warning. All 57 CPU attention tests pass locally; pre-commit clean.

SolitaryThinker:
Rebased onto main to resolve conflicts with the default_backend threading (#1379 chain) and the unit-test lane list; no content changes (range-diff clean).

SolitaryThinker:
Pure rebase onto current main to pick up #1597’s FlashAttention resolver-test fix and #1596’s already-merged SDPA head-size correction; no contributor content changes (range-diff clean after accounting for #1596’s absorbed main-side hunk).

SolitaryThinker:
Fixed the exact-head unit-lane integration failure in 1e38fe47e: #1596 correctly made SDPA head sizes unrestricted, so this PR's new capability test now expects None instead of the pre-#1596 finite list. Reproduced before the fix (1 failed, 61 passed, 37 skipped) and verified after it (62 passed, 37 skipped) on Torch 2.12; the branch was fast-forwarded, not force-pushed. Ready-label CI has restarted automatically; no duplicate /merge or /test full was posted.

SolitaryThinker:
Maintainer follow-up pushed at 6828eacdfe715a95c78ea0761828886d84c9ae04:

- Detect renamed/local FullAttn checkpoints from published model-index metadata (`_class_name=WanDMDPipeline` plus `expand_timesteps=true`) while preserving sparse FastWan routing when that marker is absent.
- Compare an explicitly selected backend against its canonical enum name, so an incompatible `TORCH_SDPA` pin now hard-fails while a platform-substituted SDPA fallback still only warns.

Local evidence on the exact head:

- `pytest fastvideo/tests/attention -q`: 65 passed, 37 skipped.
- `pytest tests/local_tests/ltx2/test_ltx2_registry.py -q`: 8 passed.
- Project pre-commit hooks on all four changed files: passed (yapf, ruff, codespell, mypy, filename/suggestion checks).
- `git diff --check`: passed.

Fresh CI is running as Buildkite full suite 4453 and fastcheck 687. The earlier 4452 lane failures are attached to the obsolete head and were not retried.

SolitaryThinker:
Maintainer exact-head summary (`6828eacdfe715a95c78ea0761828886d84c9ae04`):

- Native GB200 generation on this exact head:
  - Full/FLASH — `FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers@3e187042a324f6f5fb68fd22110a78725253de8f`, 61×448×832: 2.69s generation, 3.03s E2E.
  - Sparse/VSA — `FastVideo/FastWan2.1-T2V-1.3B-Diffusers@25e7ed7f41fd8ce2fdd108688c65e8caf0ce3aef`, 61×448×832: 9.18s generation, 9.48s E2E.
  Both generated videos were visually inspected and are coherent and high-quality.
- Passing: GitHub pre-commit, docs, and Full Suite trigger; [Fastcheck 687](https://buildkite.com/fastvideo/pr-fastcheck/builds/687); and every non-VSA lane in [Full Suite 4454](https://buildkite.com/fastvideo/ci/builds/4454), including SSIM (23/23), inference-vmoba, LoRA, performance, train-framework, and the standard-training retry.
- The sole red lane is VSA training, which repeated the known H200/bimodal `grad_norm` infra signature after the single allowed retry (ref `0.2447737`, initial `1.2867289`, retry `1.2720746`; allowed diff `0.5`). The retry budget is exhausted and the park condition is met.

No further retry and no merge action. The remaining independent gate is an approving review.

PR #1513, opened 2026-06-29
[bugfix]: preserve FSDP hooks for RMSNorm qk norms

### LINKED ISSUE #1379: RMSNorm.forward_native doesn't unwrap DTensor weight under FSDP CPU offload

macthecadillac:
## Summary

Fixes #1379.

This replaces the remaining direct q/k RMSNorm `forward_native(...)` calls with normal module dispatch so FSDP hooks run:

- `self.norm_q.forward_native(query)` -> `self.norm_q(query)`
- `self.norm_k.forward_native(key)` -> `self.norm_k(key)`

The remaining production call sites were in causal Wan and MatrixGame2 causal transformer blocks. This keeps the selected RMSNorm implementation unchanged while avoiding bypassing `nn.Module.__call__` under FSDP.

## Root Cause

Calling `RMSNorm.forward_native(...)` directly bypasses FSDP2 module hooks. With sharded RMSNorm weights, this can leave `self.weight` as a `DTensor`, and the native multiply fails with a mixed Tensor/DTensor error. Calling `RMSNorm(x)` lets FSDP unwrap/localize parameters as expected.

## Issue #1379 Investigation Checklist

- [x] Reliable reproducer: added `fastvideo/tests/layers/test_rmsnorm_forward_dispatch.py` and reproduced on Modal 2x L40S with PyTorch `2.11.0+cu128`, FSDP2 `fully_shard`, and both regular sharding and `CPUOffloadPolicy`.
- [x] Upstream issue determination: no exact matching PyTorch issue was found. The repro shows `RMSNorm(x)` works while direct `RMSNorm.forward_native(x)` fails, so this appears to be FastVideo bypassing FSDP module hooks rather than an FSDP-side bug.
- [x] Unit test for CPU-offload + DTensor weight path: the new test covers `direct_cpu_offload`, where `norm.weight` remains a CPU `DTensor` and direct `forward_native` fails with the mixed Tensor/DTensor error; it also verifies module dispatch succeeds with CPU offload.
- [x] Behavioral spec: not final-layer-only. The failure can affect any FSDP-sharded `RMSNorm(has_weight=True)` called directly through `forward_native`; it reproduces with and without CPU offload.

## Tests

- `uvx pre-commit==4.0.1 run --all-files`
  - passed after the review follow-up commit
- Modal L40S: `pytest fastvideo/tests/layers/test_rmsnorm_forward_dispatch.py -vs`
  - 2x L40S, latest branch includes the 120-second `torchrun` timeout follow-up
  - `5 passed, 14 warnings in 38.28s`
- Modal L40S: `pytest fastvideo/tests/transformers/test_wanvideo.py -vs`
  - 1x L40S, production fix applied
  - `1 passed, 14 warnings in 50.05s`
- Modal L40S targeted Wan T2V SSIM:
  - `FASTVIDEO_SSIM_MODEL_ID=Wan2.1-T2V-1.3B-Diffusers pytest fastvideo/tests/ssim/test_wan_t2v_similarity.py -vs`
  - 2x L40S, commit `75e1659310ae306d342b6bfcd0e8ef7ebb0b5caa`
  - `2 passed, 18 warnings in 123.41s`
  - `FLASH_ATTN`-parametrized mean SSIM: `0.976557461420695`
  - `TORCH_SDPA` mean SSIM: `0.9821627881791857`
  - Note: the Modal image's FlashAttention import failed and the `FLASH_ATTN`-parametrized run fell back to Torch SDPA.

## Checklist

- [x] I verified targeted Wan T2V SSIM regression tests pass on L40S

Documentation/support-matrix notes: no docs or support matrix update was needed; this does not add a model or user-facing API. GPU memory impact should be neutral because this only routes existing RMSNorm calls through module dispatch so FSDP hooks can run.

### DIFF
--- a/fastvideo/tests/layers/test_rmsnorm_forward_dispatch.py
+++ b/fastvideo/tests/layers/test_rmsnorm_forward_dispatch.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import argparse
 import json
 import os
-import socket
 import subprocess
 from pathlib import Path
 from typing import Any
@@ -15,14 +14,9 @@ import torch
 import torch.distributed as dist
 from torch.distributed import init_device_mesh
 from torch.distributed.fsdp import CPUOffloadPolicy, fully_shard
+from torch.distributed.tensor import DTensor
 
 from fastvideo.layers.layernorm import RMSNorm
-
-try:
-    from torch.distributed.tensor import DTensor
-except ImportError:
-    from torch.distributed._tensor import DTensor
-
 
 WORLD_SIZE = 2
 HIDDEN_SIZE = 8
@@ -30,21 +24,14 @@ SEED = 1379
 REPO_ROOT = Path(__file__).resolve().parents[3]
 
 
-def _free_port() -> int:
-    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
-        sock.bind(("127.0.0.1", 0))
-        return int(sock.getsockname()[1])
-
-
 def _run_torchrun(script_path: Path, mode: str, output_path: Path) -> None:
+    # --standalone binds the rendezvous port atomically, avoiding the
+    # free-port-probe race a hand-picked --master_port would have.
     cmd = [
         "torchrun",
-        "--nnodes",
-        "1",
+        "--standalone",
         "--nproc_per_node",
         str(WORLD_SIZE),
-        "--master_port",
-        str(_free_port()),
         str(script_path),
         "--rmsnorm-fsdp-worker",
         "--mode",
@@ -112,9 +99,12 @@ def _run_worker(mode: str, output_path: Path) -> None:
         fsdp_kwargs: dict[str, Any] = {"mesh": mesh}
         if mode.endswith("cpu_offload"):
             fsdp_kwargs["offload_policy"] = CPUOffloadPolicy(pin_memory=False)
-        sharded_norm = fully_shard(norm, **fsdp_kwargs)
-        if sharded_norm is not None:
-            norm = sharded_norm
+        # fully_shard is applied to the bare RMSNorm to make the hook bypass
+        # observable. Production sharding (fsdp_load.shard_model) only wraps
+        # whole transformer blocks, whose pre-forward all-gather localizes norm
+        # weights before the qk-norm call sites run, so this pins the dispatch
+        # invariant rather than reproducing a production topology.
+        fully_shard(norm, **fsdp_kwargs)
 
         x = torch.randn(2, 3, HIDDEN_SIZE, device=device, dtype=torch.bfloat16)
         call_kind = "direct" if mode.startswith("direct") else "module"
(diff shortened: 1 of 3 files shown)


fastvideo/tests/layers/test_rmsnorm_forward_dispatch.py
```diff
@@ -0,0 +1,210 @@
...
+    try:
+        mesh = init_device_mesh("cuda", (world_size,))
+        norm = RMSNorm(HIDDEN_SIZE, eps=1e-6, has_weight=True).to(device)
+        with torch.no_grad():
+            norm.weight.fill_(1.0)
+
+        fsdp_kwargs: dict[str, Any] = {"mesh": mesh}
+        if mode.endswith("cpu_offload"):
+            fsdp_kwargs["offload_policy"] = CPUOffloadPolicy(pin_memory=False)
+        sharded_norm = fully_shard(norm, **fsdp_kwargs)
```
SolitaryThinker: **P2: exercise the production FSDP ownership boundary.** This directly `fully_shard`s the RMSNorm, but FastVideo's Wan/MatrixGame2 configs shard each `blocks.<n>` parent. The block's pre-forward hook therefore all-gathers descendant parameters before its forward body, so the old `norm_q.forward_native()` runs with an already-unsharded weight. This setup proves that direct methods bypass hooks when RMSNorm itself is the FSDP unit; it does not reproduce either edited call site. Please make the regression wrap and call the affected block (or a minimal parent owning a child RMSNorm) with the production topology and show that old code fails while the new code passes. If the old code passes, this should be framed as defensive cleanup rather than closing #1379.

SolitaryThinker (commented):
Reviewed the production dispatch changes and regression coverage. One inline correctness finding; CI status intentionally excluded.

SolitaryThinker (approved):

PR #1526, opened 2026-07-02
[infra] Auto-rebuild CUDA images when docker/Dockerfile changes on main

zainnh:
## Purpose

Rebuild the CUDA images automatically when their Dockerfile changes on `main`, instead of only via manual dispatch. This is ticket T6 from the CI/CD improvement tickets doc; no dedicated GitHub issue exists for it.

## Changes

- Added a `push` trigger scoped to `docker/Dockerfile` on `main`. The CUDA matrix is the only lane that builds from that file, so the path filter alone is a sufficient change detector — no detect-changes job. Dreamverse (`apps/dreamverse/docker/Dockerfile`) and `docker/rocm7_1.Dockerfile` stay manual-dispatch only.
- `build-cuda-images` now runs on a Dockerfile push or a manual dispatch with `build_cuda_matrix`. The push arm is guarded to `github.repository == 'hao-ai-lab/FastVideo'` so forks don't start unattended builds when they sync past a Dockerfile change; manual dispatch still works in forks.
- `publish-cuda-manifests` derives its gate from the build job (`!cancelled() && needs.build-cuda-images.result != 'skipped'`) instead of mirroring the trigger expression, so the two conditions cannot drift apart.
- Added a single static concurrency group with `cancel-in-progress: false`. Every run of this workflow writes the same mutable tags (`latest`, `py3.12-latest`, …), so runs must serialize: concurrent push/dispatch runs would race on those tags, and cancelling a run mid-publish could strand the cu126/cu130 tag families at different commits. GitHub keeps one pending run per group, so the newest queued run replaces older queued ones; a superseded in-flight build finishes and is then overwritten by the newer run.
- `_template-build-image.yml` untouched (read-only reference per the ticket).

## Test Plan

Trigger selection was verified end-to-end on a personal fork (on an earlier revision of this PR whose trigger semantics for the single in-scope Dockerfile are the same):

- Dockerfile push → CUDA build triggered
- Unrelated file push (e.g. README) → no workflow run created
- Manual dispatch with all inputs false → zero jobs, no errors
- Manual dispatch with one input true → only that lane runs (regression check against existing behavior)
- Dreamverse jobs untouched by Dockerfile pushes, still work via manual dispatch

The concurrency block and job gates were reworked after review (serialization instead of event-scoped cancellation, fork guard, derived publish gate); those rely on documented GitHub Actions concurrency/`needs` semantics and were validated with:

```bash
actionlint .github/workflows/infra-build-image.yml
pre-commit run --files .github/workflows/infra-build-image.yml
```

## Test Results

<details>

actionlint: exit 0, no output
pre-commit (scoped to the changed file):
codespell.................Passed
Lint GitHub Actions workflow files.................Passed
(all other hooks skipped — not applicable to this file type)

</details>

## Notes

- Known gap, follow-up welcome: the path filter watches only `docker/Dockerfile`, but the image also depends on `docker/uv-excludes` (COPY'd and passed to `uv pip install --excludes`) and on build args defined in this workflow file. Changes to those still require a manual dispatch.
- `ci_architecture.md` is deliberately not updated per-ticket; the ticket doc says to update it once all CI/CD tickets are done.

## Checklist

- [x] I ran `pre-commit run --all-files` and fixed all issues *(ran scoped to the one changed file; other hooks don't apply to workflow YAML)*
- [x] I added or updated tests for my changes *(CI/CD workflow change — verified via live runs on a fork, see Test Plan)*
- [x] I updated documentation if needed *(see Notes: ci_architecture.md is updated once all CI/CD tickets land)*
- [x] I considered GPU memory impact of my changes *(not applicable, no runtime/GPU code touched)*

### DIFF
--- a/.github/workflows/infra-build-image.yml
+++ b/.github/workflows/infra-build-image.yml
@@ -13,11 +13,30 @@ on:
         required: false
         default: false
         type: boolean
+  # Auto-rebuild the CUDA images when their Dockerfile changes on main. The CUDA
+  # matrix is the only lane that builds from docker/Dockerfile, so a path-scoped
+  # push trigger is a sufficient change detector on its own -- no separate
+  # detect-changes/paths-filter job is needed now that there is a single
+  # in-scope Dockerfile. Dreamverse (apps/dreamverse/docker/Dockerfile) and the
+  # rocm Dockerfile stay manual-dispatch only.
+  push:
+    branches: [main]
+    paths:
+      - 'docker/Dockerfile'
 
 
 permissions:
   contents: read
   packages: write
+
+# Scope the group by event type so push-triggered auto-rebuilds and manual
+# workflow_dispatch runs on main (both on refs/heads/main) never share a group
+# and cancel each other. Only a newer Dockerfile push supersedes an older push;
+# manual dispatches always run to completion so a zero-job dispatch can never
+# silently kill an in-progress rebuild (and vice versa).
+concurrency:
+  group: infra-build-image-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'push' }}
 
 jobs:
   # CUDA matrix: Python 3.12 x {12.6.3, 13.0.0} x {amd64, arm64}. Each architecture
@@ -28,7 +47,9 @@ jobs:
   # aliases; 13.0.0/cu130 is published under explicit versioned tags. Flash-attn
   # 2.8.3 comes from the architecture-specific prebuilt releases.
   build-cuda-images:
-    if: ${{ github.event.inputs.build_cuda_matrix == 'true' }}
+    # Runs on a manual dispatch when build_cuda_matrix is set, or automatically
+    # on a push that changed docker/Dockerfile (inputs are null on push).
+    if: ${{ github.event_name == 'push' || github.event.inputs.build_cuda_matrix == 'true' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/infra-build-image.yml
```diff
@@ -13,12 +13,26 @@ on:
...
+    paths:
+      - 'docker/Dockerfile'
 
 
 permissions:
   contents: read
   packages: write
 
+concurrency:
+  group: infra-build-image-${{ github.ref }}
```
SolitaryThinker: **MAJOR — concurrency group cancels across event types.** The new workflow-level `concurrency` group is keyed only on `github.ref` with `cancel-in-progress: true`. Both the new `push: branches: [main]` trigger and any `workflow_dispatch` on `main` resolve to `refs/heads/main`, so they share one group (`infra-build-image-refs/heads/main`) and cancel each other regardless of event type.

Concretely: merge a `docker/Dockerfile` change → the push rebuild starts. While it runs, a manual "Run workflow" on `main` with inputs left at their `default: false` (the zero-job dispatch your own test plan documents as normal) cancels the in-progress rebuild. The dispatch then runs nothing — `if: ${{ github.event_name == 'push' || github.event.inputs.build_cuda_matrix == 'true' }}` is false and `publish-cuda-manifests` mirrors the same gate — so the auto-rebuild is silently lost and GHCR images are left stale or partially published. A Dockerfile push can likewise cancel an in-progress manual build.

Scope cancellation to genuinely superseding runs, e.g. `group: infra-build-image-${{ github.event_name }}-${{ github.ref }}`, or keep one group but set `cancel-in-progress: ${{ github.event_name == 'push' }}` so only a newer push of the same ref cancels an older run while manual dispatches run independently.

SolitaryThinker (changes requested):
Hi @zainnh, thanks for the PR!

One change needed before this is safe to merge — see the inline comment.

.github/workflows/infra-build-image.yml
zainnh: Thanks for catching this - confirmed the group was only keyed on `ref`, so push and dispatch runs on `main` shared a slot.

Applied the event-scoped fix:

```yaml
concurrency:
  group: infra-build-image-${{ github.event_name }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'push' }}
```

Verified on a fork:
- A push-triggered build ran to full completion, untouched by a concurrent zero-job manual dispatch.
- That dispatch itself completed cleanly with all jobs skipped.
- Rapid back-to-back pushes still correctly cancel the older run in favor of the newer one (same-event-type cancellation still works).

Let me know if you'd like anything else adjusted.

PR #1556, opened 2026-07-05
[test]: remove dead and duplicate tests (-489 lines)

SolitaryThinker:
## Problem
A sweep of the test tree (follow-up to #1552's CI-collection guard) found tests that provide zero signal: files skipped unconditionally for up to a year, duplicates, and one file that broke pytest collection on non-Blackwell machines.

## What this removes (one commit, 6 files)
- 4 unconditionally-skipped dead test files (one with a hardcoded personal /mnt/weka path that still burned collection time in the VAE lane): test_official_wan_vae, test_hunyuanvideo (transformers), test_hunyuan_vae, test_matrixgame2
- 3 duplicate forward-direction tests in test_compat_translation.py (unique reverse-direction coverage kept)
- 1 fix (not a deletion): test_nvfp4_fa4.py no longer breaks pytest collection of tests/local_tests/ on non-Blackwell boxes

## Scope note
tests/local_tests/ and fastvideo/tests/nightly/ deletions were dropped from this cleanup per maintainer direction — those directories stay as-is.

## Honest notes
Two coverage gaps become explicit (they were already real — these tests never ran): Wan2.2 VAE parity and HunyuanVideo transformer/VAE parity have no live coverage; candidates for re-enablement issues.

## Verification
Rebased onto current main (post #1553/#1554/#1555): all three guard assertions in fastvideo/tests/contract/test_ci_test_collection.py pass; every deleted file re-verified as still unconditionally skipped on main with no live references repo-wide.

## Risk & rollback
Deletions of never-running tests + one import-safety fix; revert one commit.


SolitaryThinker:
Reworked per maintainer direction: tests/local_tests/ and fastvideo/tests/nightly/ deletions are dropped — those stay. The PR is now just the 4 dead skipped files, the 3 in-file duplicates, and the nvfp4 collection fix. Head c7591757, rebased onto main with the pre-commit fix in (#1554), so checks should be genuinely green now.

PR #1339, opened 2026-05-12
[fix] Z-Image (#1236) draft port: rebase + strict-load contract + bf16 encoder parity + PORT_STATUS

Mister-Raggs:
## Summary

Fixup branch on top of @mrinaald's draft Z-Image port (#1236), applying findings from a `.agents/skills/add-model-10-pr-review` pass run via subagent. Base PR is in-progress (transformer, pipeline, conversion deferred) — this PR addresses everything already in scope and unblocks rebase + handoff gating.

Stacks 9 fixup commits on top of the 3 author commits, rebased onto current `main` (post-Cosmos-2.5 merge). **All component parity validated on `Tongyi-MAI/Z-Image-Turbo` weights, A40 GPU, 2026-05-12.**

## Findings addressed

**P0 — Blockers**
- **P0-1/2** Stale base (`e1e0d91c` pre-Stable-Audio) deleted `StableAudioConditionerArchConfig`/`StableAudioConditionerConfig` exports and the `AutoencoderOobleck → OobleckVAE` registry mapping on merge. Fixed by rebase.

**P1 — Should fix**
- **P1-1** `Qwen3Model.load_weights` silently dropped `lm_head.weight` — no strict-load contract. Added `Qwen3Model.ALLOWED_UNEXPECTED_KEYS = {"lm_head.weight"}`; loader now raises on any other unexpected key. Encoder parity test asserts the observed unexpected-key set ⊆ allowlist.
- **P1-2** Encoder parity originally forced `dtype=torch.float32` because `atol=1e-4` failed in bf16. Per `add-model-02-parity` calibration block: parametrize fp32 + bf16, replace element-wise `assert_close` (meaningless when cross-kernel bf16 produces a long tail) with distribution checks (mean + median), and require per-batch diagnostics. Empirically calibrated thresholds against `Tongyi-MAI/Z-Image-Turbo` on A40: `last_hidden_state` mean < 0.025 / median < 0.020; `hidden_states[-2]` mean < 0.120 / median < 0.100 — 1.5–1.6x headroom.
- **P1-3** Added `tests/local_tests/zimage/__init__.py` — every sibling family dir has one.
[...]

PR #1561, opened 2026-07-06
[ci] Stop forcing FA4 in model-load lanes

### LINKED ISSUE #1558: [Bug] FA4 (flash_attn.cute) fmax() TypeError crashes full-suite training/model-load tests — cutlass-dsl API mismatch, forced by FASTVIDEO_FA4=1 default
### Describe the bug

The full-suite training / model-load tests crash inside the FA4 (`flash_attn.cute`) kernel during CuTeDSL compilation, with:

```
TypeError: fmax() takes 2 positional arguments but 3 positional arguments (and 3 keyword-only arguments) were given
```

This is **not** a model bug — it's a version mismatch between the pinned `flash-attn` cute kernels and the installed `nvidia_cutlass_dsl` in the CI image. `flash_attn/cute/utils.py` guards on `CUDA_VERSION == 12.9` and calls the "old" 3-positional-arg form of `nvvm.fmax(T.f32(), a, b, ...)`, but the `nvvm.fmax` binding in the installed cutlass-dsl now takes 2 positional args:

```python
# flash_attn/cute/utils.py
if CUDA_VERSION.major == 12 and CUDA_VERSION.minor == 9:
    return Float32(
        nvvm.fmax(                 # <-- 3 positional args
            T.f32(),
            Float32(a).ir_value(loc=loc, ip=ip),
            Float32(b).ir_value(loc=loc, ip=ip),
            c=... ,
        )
    )
# E TypeError: fmax() takes 2 positional arguments but 3 ... were given
```

### Why it fires in CI (and blocks every PR's full suite)

The Modal test harnesses default `FASTVIDEO_FA4=1`:

- `fastvideo/tests/modal/pr_test.py` — `"FASTVIDEO_FA4": os.environ.get("FASTVIDEO_FA4", "1")`
- `fastvideo/tests/modal/launch_l40s_job.py` — same default
- `fastvideo/tests/modal/ssim_test.py` — same default

Since #1540 made FA4 an explicit opt-in, `FASTVIDEO_FA4=1` forces `fa_version="4"`, so attention routes through `flash_attn.cute`, which then fails to compile as above. Because the harness forces FA4 on regardless of the PR diff, this crashes the model-load and training lanes for **any** open PR, not a specific change.

### Reproduction

Full suite (`pr_test.py`) — the following all fail at the same `fmax()` TypeError:

```
FAILED fastvideo/tests/train/models/test_load_cosmos.py::test_cosmos_model_loads_and_forwards
FAILED fastvideo/tests/train/models/test_load_hunyuan.py::test_hunyuan_model_loads_and_forwards
FAILED fastvideo/tests/train/models/test_load_longcat.py::test_longcat_model_loads_and_forwards
FAILED fastvideo/tests/train/models/test_load_matrixgame2.py::test_matrixgame2_model_loads_and_forwards
FAILED fastvideo/tests/train/models/test_load_wan.py::test_wan_model_loads_and_forwards
```

Abbreviated call path (Wan T2V forward → self-attn → FA4 cute → compile):

```
fastvideo/attention/backends/flash_attn.py:125  flash_attn_func_compilable
  -> fastvideo/attention/utils/flash_attn_cute.py:301  flash_attn_func
    -> flash_attn/cute/interface.py:1028  cute.compile(...)
      -> flash_attn/cute/softmax.py:150  online_softmax
        -> flash_attn/cute/utils.py:401  fmax_reduce
          -> flash_attn/cute/utils.py:361  fmax  ->  TypeError
```

### Suggested fixes (either unblocks CI)

1. **Align the image pins** so `flash-attn` (cute) and `nvidia_cutlass_dsl` agree on the `nvvm.fmax` signature under CUDA 12.9.
2. **Don't force FA4 for the model-load / training lanes** — those don't require FA4; dropping the `FASTVIDEO_FA4=1` default (or setting `=0`) for them falls back to FA2/FA3 and they pass. FA4 can stay enabled for the inference/SSIM lanes that intend to exercise it.

### Related

- #1524 — FlashInfer/CUTLASS runtime-compile failure on RTX 5080 (sm_120). Different kernel (FlashInfer NVFP4, not `flash_attn.cute`) and a different symptom, but the same broad "cutlass-dsl runtime compile breaks in the shipped image" family; likely worth tracking together.
- Recent FA4 / kernel churn that this most likely rode in on: #1540 (FA4 opt-in), #1539 / #1541 (fastvideo-kernel pin bumps, FA4 tile_mn port), #1447 (multi-arch).

macthecadillac:
## Summary

Fixes #1558 by keeping FA4 enabled only where CI intentionally exercises FA4, while returning model-load and training-style Modal lanes to the product default (`FASTVIDEO_FA4=0` unless explicitly opted in).

- Set `FASTVIDEO_FA4=0` for transformer/model-load and training-style `pr_test.py` lanes.
- Default generic `launch_l40s_job.py` runs to FA4 off, preserving opt-in through local env or `--env-vars`.
- Add a source-only contract test covering the Modal FA4 policy and preserving SSIM's FA4 default.

## Validation

- Modal L40S `pytest fastvideo/tests/contract/test_modal_fa4_policy.py fastvideo/tests/train/models -vs`
  - `9 passed in 92.34s`
  - App: `ap-dBp3HNyPscUyqd2TamekW3`
- Modal L40S `pytest fastvideo/tests/transformers -vs`
  - `5 passed, 1 skipped in 265.13s`
  - App: `ap-hrzrFjSWJ2w8h8qcjaDmqW`
- Modal L40S `pre-commit run --all-files`
  - passed
  - App: `ap-u9fpYmjWmXT4zeYVnWWFRt`
- Modal L40S bot-review follow-up `pytest fastvideo/tests/contract/test_modal_fa4_policy.py -vs`
  - `3 passed in 0.03s`
  - App: `ap-PRZBk52U26BSKlWZ7L9JzO`
- Modal L40S bot-review follow-up `pre-commit run --all-files`
  - passed
  - App: `ap-Q3223MMVidmMMLuXZuKYNz`

## Notes

This unblocks CI lanes that do not need FA4. It does not resolve the underlying explicit-FA4 image compatibility mismatch between the pinned FA4 cute overlay and the currently resolved `nvidia-cutlass-dsl`; that remains a separate image/dependency follow-up.

# Checklist


Mister-Raggs:
Extra evidence that this fixes more than the model-load lanes — the `FASTVIDEO_FA4=0` you add to `run_training_tests_VSA` looks like it also fixes `test-tube-training-tests-vsa`, which is currently red on unrelated PRs.

On #1339 (whose diff touches no training-path file) that lane fails only on the `grad_norm` threshold:

```
INFO: grad_norm, diff: 0.994, threshold: 0.5, reference: 0.2448, current: 1.2390
INFO: train_loss, diff: 0.0058, threshold: 0.04   <- loss is fine
```

The run log shows why. VSA can't serve every layer, and the fallback now lands on FA4:

```
WARNING [selector.py:127] Requested attention backend VIDEO_SPARSE_ATTN is not supported by this
        layer; supported backends are ['FLASH_ATTN', 'TORCH_SDPA']. Falling back to automatic selection.
INFO    [flash_attn.py:133] Using FlashAttention-4 backend
```

`fa_version == "4"` is only reachable when `FASTVIDEO_FA4` is set. But `h100/h200_reference_wandb_summary_VSA.json` was last recorded 2026-05-16 (#1352) — back when that fallback ran FA2/FA3, and long before FA4 was even functional (it crashed on `fmax` until #1564). So the lane has been comparing FA4 gradients against FA2/FA3 references. `train_loss` barely moves while `grad_norm` moves 5x, which is what you'd expect from an attention-backward swap rather than a real regression.

With `FASTVIDEO_FA4=0` the resolver skips cute, finds no `flash_attn_interface` in the image (only the `flash_attn` 2.8.3 wheel), and returns `"2"` — i.e. back to the kernel the references were seeded against. So this should restore that lane without re-recording anything.

Worth noting the same `fa_version == "2"` gate is what activates the real backward in #1388, so this PR also puts that path back under test.

SolitaryThinker (approved):

PR #1562, opened 2026-07-06
[ci]: cache FastVideo kernel builds in Modal

### LINKED ISSUE #1523: [ci][p1] Cache FastVideo kernel builds in Modal CI lanes
## Summary

FastVideo Docker images prebuild the custom kernel during image creation, and `build.sh` can skip torch CUDA probing when `TORCH_CUDA_ARCH_LIST` is set. However, Modal CI lanes still call `./build.sh` in `pr_test.py` and `ssim_test.py`, causing most Modal jobs to rebuild the kernel from source.

Add kernel build caching or reuse so Modal tests avoid unnecessary rebuilds.

## Problem

Current state:

- Docker images prebuild the kernel.
- `fastvideo-kernel/build.sh` has partial optimization behavior.
- Modal runners still execute `./build.sh`.
- There is no Modal wheel/build cache for the kernel.
- Most Modal CI lanes pay repeated kernel build cost even when inputs have not changed.

This increases CI latency and wastes GPU/CPU job time.

## Proposed Behavior

Add a reusable cache path for FastVideo kernel builds in Modal CI.

Possible approaches:

- Build a wheel once and cache it in a Modal volume keyed by:
  - kernel source hash
  - Python version
  - torch version
  - CUDA version
  - architecture list
- Teach `pr_test.py` and `ssim_test.py` to install the cached wheel when available.
- Fall back to `./build.sh` only on cache miss.
- Upload/store the newly built wheel after successful build.
- Ensure Docker-prebuilt kernels are reused when already valid.

## Acceptance Criteria

- Modal CI lanes avoid rebuilding `fastvideo-kernel` when a valid cached build exists.
- Cache key changes when kernel source or ABI-relevant dependencies change.
- Cache miss still builds successfully from source.
- `pr_test.py` and `ssim_test.py` use the same caching helper/path.
- Logs clearly report cache hit vs cache miss.
- No behavior change for local developer builds.
- CI correctness is preserved across Python, torch, CUDA, and architecture changes.

## Relevant Files

- `fastvideo-kernel/build.sh`
- `fastvideo/tests/modal/pr_test.py`
- `fastvideo/tests/modal/ssim_test.py`
- Docker image build files under `docker/`
- `.buildkite/scripts/pr_test.sh`

## Notes

This is likely high-impact for CI runtime because many Modal lanes currently rebuild the same kernel from source. The main risk is stale binary reuse, so the cache key needs to be conservative.

### Related resources

_No response_

macthecadillac:
## Summary

- Preserve Docker-prebuilt exact-match kernel reuse and job-local source-build fallback.
- Disable the cross-PR shared executable cache in ordinary PR and SSIM jobs; branch-controlled jobs no longer mount, write, or commit that volume.
- Bake exact-key SM89 and configured-architecture wheels into trusted amd64 development images while preserving the configured wheel as the installed default.
- Rebuild those trusted artifacts from the protected main workflow whenever kernel, cache-key, dependency, Docker context, submodule, or image-workflow inputs change.
- Add schema-v3 artifact metadata with wheel name, size, SHA-256, distribution, ZIP integrity checks, and invalid-entry recovery.
- Separate cache roles: `install` is a non-mutating consumer, while `produce` is the only writer and is used by the isolated smoke workflow.
- Key artifacts on source/submodule state, Python/Torch/CUDA metadata, resolved architecture, build flags, selected compilers, Torch build configuration, and CXX11 ABI.
- Add the Modal cache/runner tests to fastcheck, including PR/SSIM wiring, publication lifecycle, prebuilt validation, corruption fallback, and producer repair.

Fixes #1523

## Validation

- `uvx --from pre-commit pre-commit run --all-files` passed.
- Modal L40S rebased focused tests: `36 passed in 0.33s`.
  - https://modal.com/apps/hao-ai-lab/main/ap-lkc0wAiZDhFlNMNm0iIBWi
- Expanded fastcheck command: `744 passed, 1 warning in 13.94s`.
  - https://modal.com/apps/hao-ai-lab/main/ap-4ku9YrA0oOqGKCUS2hBUMP
  - The warning is the existing Pillow `mode` deprecation in `test_image_vae_encoding.py`.
- Rebase integration preserved upstream's bounded checkout retries and detached shallow checkout.
- Real L40S trusted-prebuilt and isolated producer/consumer cache flows passed.
- Current-head GitHub pre-commit, Buildkite aggregate, fastcheck, and all microscope lanes passed.
  - https://buildkite.com/fastvideo/ci/builds/4508

## Review Follow-Up

- Shared PR/SSIM volume writes are disabled; ordinary jobs consume only trusted image artifacts and otherwise build locally without publishing executable state.
- Trusted amd64 image publication now retains both SM89 and configured-architecture artifacts and is triggered by every repository-controlled artifact/key input.
- Consumers validate size, SHA-256, wheel archive structure, and distribution before installation, then fall back to a local build without mutating shared state.
- ABI-affecting flags, compiler selection/version, Torch configuration, and CXX11 ABI now invalidate the key.
- The isolated smoke uses a dedicated v3 volume, an explicit producer, and a read-only consumer mount.
- Fastcheck collects all three Modal cache/runner test modules.

## Remaining Risk

- Full SSIM generation was not run, and `full-suite-passed` has not reported.
- The canonical multi-architecture image publication runs only after merge to `main`, so the actual republished registry image could not be consumed in this PR.
- Cross-run shared-volume publication remains intentionally disabled; ordinary reuse comes from protected image publication.

# Checklist


Satyam-53:
Thanks for the work here. This mostly lines up with #1523: the Modal PR/SSIM paths now share a kernel cache helper and cache volume, the key includes source plus Python/torch/CUDA/arch/build metadata, build.sh keeps local behavior unchanged unless --wheel-dir is passed, and the opt-in cross-container smoke gives useful coverage for the intended cache hit/miss flow.

I have one code concern before approval: _store_cache_entry() catches every OSError from temp_entry.rename(cache_entry), removes the temp entry, and still returns cache_entry. The caller then always logs stored wheel cache entry. That is only safe for the concurrent-writer case if a valid cache entry actually exists afterward. For real Modal volume/store failures, this can make CI logs claim the wheel was cached when it was not. Please revalidate the target entry after the rename exception, and only swallow the error if the existing metadata/wheel match; otherwise fail or log a clear “built but not cached” state.

Also, the current head is still failing Buildkite fastcheck: microscope-kernel-tests, microscope-transformer-tests, and microscope-encoder-tests. Since those lanes run through the pr_test.py path changed by this PR, they need to be investigated and resolved before merge. I’ll look into the Buildkite logs separately, but for now this should stay blocked on the failing tests.

Satyam-53:
Do not merge it yet, the changes are not comprehensive.

SolitaryThinker (changes requested):
Changes requested at exact head 5bcf430842bb1ba4e1724673b9ac783150d5d05f.

1. **P1 — PR-controlled code can poison a shared executable cache.** `pr_test.py:24,89-92,129-164` mounts the same writable `fastvideo-kernel-build-cache` for PR/fork code, executes the checked-out helper, and commits the volume. `kernel_build_cache.py:253-284,353-358` then trusts key/schema plus a wheel-shaped filename and installs that executable artifact in later jobs, including jobs carrying HF/W&B secrets. Modal volumes also commit in the background/on shutdown, so moving the explicit commit earlier is not a sufficient trust boundary. Please disable shared PR writes until a trusted base-controlled producer can publish artifacts that PR consumers mount read-only; untrusted PR code must not receive a cross-PR writable executable store. See [Modal volume semantics](https://modal.com/docs/guide/volumes).

2. **P1 — corrupt/partial entries are treated as hits and permanently wedge a key.** `_cache_hit()` does not verify a digest, size, or wheel archive; an install failure propagates without eviction/rebuild, and `_store_cache_entry()` refuses to replace an invalid existing entry. The unit fixtures at `test_kernel_build_cache.py:98-104,145-164` demonstrate that empty/plain-text `.whl` files count as valid. Persist and verify SHA-256 plus size, reject invalid wheel archives, quarantine/evict and rebuild on validation or install failure, then bump the schema.

3. **P2 — ABI-affecting inputs are absent from the key.** `kernel_build_cache.py:190-225` hashes only backend, resolved arch, and `CMAKE_ARGS` from the build inputs. It omits effective `CFLAGS`, `CXXFLAGS`, `LDFLAGS`, selected compiler versions, `torch.__config__.show()`, and PyTorch's CXX11 ABI bit. An exact-head probe changed `CFLAGS` without changing the key. Include those inputs and add an invalidation regression with the schema bump.

4. **P2 — the concurrency model does not match Modal snapshots.** Parallel lanes do not observe another container's commit without reload, so same-key fanout still independently misses/builds and later commits last-write-wins. Use one trusted producer before fanout; if the intended win is only cross-run reuse, document that limit. A v2 volume improves commit scale but does not deduplicate same-key builds.

5. **P2 — Fastcheck does not collect the new tests.** `run_unit_test()`'s explicit path list at `pr_test.py:299-302` omits both new modules, so the green unit lane did not run them. Add both paths and cover SSIM integration plus prebuilt/corrupt-hit recovery.

Evidence: exact-head focused tests passed 9/9; shell syntax, Python compilation, and `git diff --check` passed. A conflict-free virtual merge onto current main also passed the same 9 tests. The linked sequential Modal miss/store/hit smoke passes, but it does not exercise untrusted writers, corrupt entries, or concurrent snapshot behavior. There is no `full-suite-passed` context yet.

macthecadillac:
@SolitaryThinker Addressed the requested changes at `afb7b06a73d5eb0eadf1185b4edbd4a56cc10837`.

1. Removed the cross-PR writable executable volume from ordinary PR and SSIM jobs. Those jobs now consume only exact-key artifacts baked by the protected main image publisher, with a job-local build fallback that does not publish state.
2. Bumped artifact metadata to schema v3 with wheel name, size, SHA-256, distribution, ZIP/METADATA/WHEEL validation, invalid-entry repair for producers, and non-mutating fallback for consumers.
3. Added effective build flags, selected compiler identities and versions, Torch build configuration, CXX11 ABI, and the remaining source/toolchain inputs to the key.
4. Replaced the Modal snapshot fanout assumption with trusted image publication. Published amd64 images retain SM89 plus configured-architecture artifacts, and the protected workflow now rebuilds them whenever repository-controlled artifact/key inputs change.
5. Added all three Modal cache/runner test modules to fastcheck, including PR/SSIM wiring, corrupt-entry recovery, trusted-prebuilt exact-hit behavior, and publication lifecycle coverage.

Validation on the latest head:
- Modal L40S targeted tests: `25 passed`
- Expanded fastcheck: `727 passed, 1` existing Pillow warning
- Real L40S SM89 trusted-prebuilt exact-key hit with local build fallback disabled
- GitHub pre-commit, Buildkite aggregate, fastcheck, and every microscope lane passed

The remaining unverified path is the canonical registry image publication, which runs after merge to `main`; `full-suite-passed` has also not reported. Please re-review the latest head.

SolitaryThinker (approved):
Re-reviewed exact head 7f3fe77e1b7ac817133bd89f95de740ad6337a5c. The prior blockers are resolved: PR jobs no longer receive a shared writable executable cache; trusted main/image builds produce immutable Docker-baked wheels, and PR consumers either install a validated matching artifact or build locally without persistence. Wheel archive/distribution/size/SHA-256 validation and corrupt/install-failure fallback are covered, ABI-affecting flags/compiler/Torch config/CXX11 ABI are keyed, the old Modal fanout assumption is removed from production CI, and all three helper/runner test modules are explicitly collected. Exact-head pre-commit and Fastcheck build 4508 are green; all review threads are resolved. No remaining blocking finding.

PR #1563, opened 2026-07-06
[bugfix]: guard FastWan FullAttn backend config

### LINKED ISSUE #864: [Bug] size mismatch for FastWan2.2-TI2V-5B-FullAttn-Diffusers
### Describe the bug

File "/users/miniconda3/envs/fastvideo/lib/python3.12/site-packages/fastvideo/models/utils.py", line 159, in pred_noise_to_pred_video
(FVWorkerProc-0 pid=3599678) ERROR 11-04 10:05:23 [gpu_worker.py:226]     pred_video = noise_input_latent - sigma_t * pred_noise
(FVWorkerProc-0 pid=3599678) ERROR 11-04 10:05:23 [gpu_worker.py:226]                  ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
(FVWorkerProc-0 pid=3599678) ERROR 11-04 10:05:23 [gpu_worker.py:226] RuntimeError: The size of tensor a (53) must match the size of tensor b (52) at non-singleton dimension 3

### Reproduction

import os

def main():
    os.environ["FASTVIDEO_ATTENTION_BACKEND"] = "VIDEO_SPARSE_ATTN"

    # Create a video generator with a pre-trained model
    generator = VideoGenerator.from_pretrained(
        "FastWan2.2-TI2V-5B-FullAttn-Diffusers",
        num_gpus=1,  # Adjust based on your hardware
    )

    # Define a prompt for your video
    prompt = "A curious raccoon peers through a vibrant field of yellow sunflowers, its eyes wide with interest."

    # Generate the video
    video = generator.generate_video(
        prompt,
        return_frames=True,  # Also return frames from this call (defaults to False)
        output_path="output_t2v_pri/",  # Controls where videos are saved
        save_video=True,
        height=480,
        width=848,
    )

if __name__ == '__main__':
    main()

### Environment

fastvideo 1.6.0

macthecadillac:
## Summary
- split `FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers` into a FullAttn-specific T2V-only pipeline config
- reject `FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN` for the dense FullAttn checkpoint before it can hit the latent shape mismatch path
- make Wan block construction use the active config's supported attention backends and add regression coverage for registry, config, backend selection, and block construction
- update durable docs/examples to stop advertising FullAttn as VSA-compatible

Fixes #864.

## Validation
- Modal L40S: `pytest fastvideo/tests/api/test_fastwan_fullattn.py -q` passed, 6 tests, run `ap-uaZrKwXeCVqZvDnSOo10YZ`
- `pre-commit run --all-files` passed from the issue worktree before draft PR creation
- Stage 3 review loop:
  - reviewer 1: no actionable findings
  - reviewer 2: found Wan blocks still received the class-level backend tuple
  - adjudicator/fixer: accepted the finding, patched Wan block construction, added regression coverage, and pushed commits
  - reviewer 3: no actionable findings

## GPU Memory Impact
- No expected increase. This change only adjusts config/registry validation, backend support plumbing, docs, and lightweight tests.

## Notes
- Wan T2V SSIM was not run because this change prevents invalid FullAttn/VSA usage and does not alter valid generation numerics.
- Related upstream PR #1494 overlaps broader backend validation, but this branch is the narrower issue-specific fix for #864.

# Checklist


SolitaryThinker:
Rebased onto current main to pick up the CuTe-DSL pin fix behind the stale transformer-lane failure; no contributor content changed (10/10 patches preserved: 9 `=`, one support-matrix context-only `!` with identical normalized changed lines).

SolitaryThinker (changes requested):
Two runtime gaps still leave supported public paths behaving incorrectly:

1. **[P1] Honor the effective backend override before selecting the Wan block** — `fastvideo/models/dits/wanvideo.py:610-620` reads only `FASTVIDEO_ATTENTION_BACKEND`, but `fastvideo/attention/selector.py:49-55,108-115` defines the supported precedence as global force first, then the environment. On this head, forcing VSA with the environment unset selects `WanTransformerBlock`, while setting the environment to VSA and globally forcing SDPA selects `WanTransformerBlock_VSA`. The attention layer then resolves the opposite implementation, so the dense wrapper can call VSA without `gate_compress`, or the VSA wrapper can pass `gate_compress` to SDPA; either path can fail at forward. Please select from the same effective override used by the central selector and add both precedence-order regressions.

2. **[P1] Enforce the advertised T2V-only restriction at runtime** — `fastvideo/registry.py:994-1003` changes `workload_types` to T2V, but that field is only UI metadata (`register_configs` says so at lines 152-156). `FastVideoArgs.from_kwargs(model_path="FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers", workload_type="i2v")` still succeeds on this head. Because this config sets `ti2v_task=False` and disables the VAE encoder, `InputValidationStage` can load the supplied image and then skip its conditioning path, silently producing unconditioned output. Please reject unsupported model/workload pairs at the shared runtime resolution/argument boundary and cover the public `FastVideoArgs` path.

Non-blocking cleanup: the example is still named `inference_wan_VSA_DMD_5B_720P.yaml` while its new comment says VSA is invalid; rename it to the FullAttn/dense name. Also, `_supported_attention_backends` is a required tuple, so the `or ()` fallback and the test that assigns `None` behind `type: ignore` only preserve an impossible state and can be removed.

Evidence: the submitted focused suite passes (`7 passed`), but it does not exercise either failing path above. The pure rebase itself is range-diff clean, and valid dense forward numerics are unchanged; no SSIM rerun is needed for these guards.

PR #1567, opened 2026-07-07
[bugfix]: skip unused output materialization

### LINKED ISSUE #1474: [Bug]: Avoid GPU-to-CPU output materialization when save_video=False and return_frames=False
### Motivation.

When save_video=False and return_frames=False, FastVideo still copies the generated video tensor from GPU to CPU and then discards it. This adds several seconds of unnecessary overhead in profiling/benchmarking workflows.
The proposed metadata-only path would skip CPU output materialization unless either saving or returning frames is requested.

### Proposed Change.

In `VideoGenerator.generate_video`, the generated output is copied to CPU before checking whether the caller requested frames:

```python
if output_batch.output.shape == samples.shape:
    samples.copy_(output_batch.output)
else:
    samples = output_batch.output.cpu()

### Feedback Period.

_No response_

### CC List.

_No response_

### Any Other Things.

Think about the edge cases and open questions on this.

## Potential Edge Cases / Open Questions

1. Should `trajectory`, `trajectory_timesteps`, or `trajectory_decoded` still be returned in metadata-only mode?

2. Should audio outputs be treated separately from video frames?

3. Should this behavior be controlled only by `save_video=False` and `return_frames=False`, or should there be a more explicit flag such as `materialize_output=False`?

4. Should benchmark/profile scripts use this metadata-only mode by default?

5. Should timing logs include a skipped or zero-duration `PostDecodeFrameProcessStage` entry?

6. If `output_type == "latent"`, should latent outputs follow the same `return_frames` behavior, or should latents have a separate return flag?

7. If downstream users currently rely on side effects from CPU materialization despite `return_frames=False`, should this be mentioned as a behavior change?

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues.

macthecadillac:
## Summary

Fixes #1474 by preventing unused decoded outputs from crossing worker transport boundaries when callers request neither saved nor returned frames.

The original entrypoint-only optimization still allowed the worker to return the full decoded CUDA tensor through multiprocessing or Ray. That transport caused the reported late device-to-host copy before `VideoGenerator` could skip its own materialization.

This change:
- drops unused decoded output inside `Worker.execute_forward`, before multiprocessing or Ray transport
- preserves the existing `None` failure sentinel for missing pipeline output
- retains output only when returned frames or a normal pixel save consumes it
- drops save-only latent outputs and audio-only video placeholders
- propagates `ForwardBatch.extra` through Ray so audio metadata survives placeholder removal
- skips entrypoint sample allocation, CPU copying, and frame construction for metadata-only requests
- preserves normal save, return-frame, audio, latent, and trajectory behavior
- adds generator, worker, and Worker-to-Ray-to-VideoGenerator regression coverage

## Validation

- Final merged head `d61ad00bf0161ce9e67c6e574738c73aa069c0c5`, Modal L40S focused tests:
  - `pytest fastvideo/tests/entrypoints/test_video_generator.py fastvideo/tests/worker/test_gpu_worker.py -q`
  - `34 passed in 0.61s`
  - App: `ap-3uLgFBAtLbN0q5B0xepHNz`
- Final merged head, Modal L40S full pre-commit:
  - `pre-commit run --all-files`
  - all hooks passed
  - App: `ap-mcsd3ggJbkEKA6hGGXeTG2`
- Final merged head, GitHub pre-commit:
  - passed
- Worker-boundary real Wan smoke on commit `e44cfaf1`:
  - direct `VideoGenerator`, multiprocessing backend
  - `480x832`, 81 frames, 1 inference step
  - `save_video=False`, `return_frames=False`
  - completed in 15.47 seconds with `samples=None`, `frames=None`, and `video_path=None`
  - App: `ap-DUPeR9g1PFMaA34OrdPq5I`
- Final review/adjudication loop:
  - Reviewer found that an unconditional worker replacement masked `output=None`; adjudicator accepted and fixed it in `0f6e48cbe`.
  - Reviewer found save-only latent/audio outputs were still transported without consumers; adjudicator accepted and fixed it in `fbc61071e`.
  - Reviewer found Ray omitted audio metadata after placeholder removal; adjudicator accepted and fixed it in `084d817a8`.
  - Closure reviewer reported no actionable code findings at `084d817a8`.
- Bot review:
  - Accepted Gemini's audio-only placeholder materialization finding.
  - Skipped placeholder tensor copy/warning for audio-only save calls when `return_frames=False`.
  - Both Gemini review threads are resolved and outdated.

## Trace Status

The requested post-worker-fix `nsys` A/B trace is still outstanding.

The exact patched workload runs successfully on Modal, but Nsight Systems 2025.5.2 and CUDA Nsight Systems 2024.6.2 both failed while finalizing the multiprocessing CUDA trace with `PostMortemTimeConverter::ConvertGpuTicksToSyncNs ... InternalErrorException`; neither produced a usable `.nsys-rep`. Therefore this PR does not claim trace proof that the 388,177,920-byte transfer has disappeared. A rerun in the reporter's working Nsight environment is still required.

## GPU Memory Impact

Expected to avoid transporting and materializing the full decoded tensor for metadata-only requests, reducing device-to-host transfer and host allocation overhead. The change does not increase GPU memory use. Output paths that save pixel video or return frames retain their existing tensors.

## Notes

- The branch was merged with current upstream `main` to restore the current full-suite gate scripts and remove stale-base CI failures.
- Wan T2V SSIM was not run because retained pixel-output paths are unchanged; the fix only elides outputs with no consumer.
- Related open PR #1362 is a broader GPU-side uint8 post-decode optimization. It does not include this worker-boundary transport fix.

# Checklist


Satyam-53:
Thanks for the PR, the code path looks directionally correct to me: metadata-only calls now skip pixel preallocation, avoid `output.shape` inspection / `.cpu()` materialization, and avoid frame construction, while `return_frames=True` and save-only video paths still materialize the decoded output they need. The earlier audio-only placeholder-copy concern also appears addressed in the latest head.

That said, I think this should be blocked until we have a repro validated with this PR cherry-picked and an `nsys` trace confirming the intended D2H behavior.

Required repro:
1. Run the same direct `VideoGenerator` workload on baseline `main` and with this PR cherry-picked.
2. Use `save_video=False` and `return_frames=False`.
3. Capture an `nsys` trace for both runs.
4. Compare the end of the trace after decode/generation completes.

Expected result:
- Without this fix, the trace should show a large device-to-host copy near the end, corresponding to the decoded video tensor being copied back to CPU even though neither frames nor a saved video are requested.
- With this fix, that large end-of-generation D2H copy should disappear for `save_video=False, return_frames=False`. Small incidental D2H copies for metadata/logging are fine, but the decoded output tensor copy should not be present.

**Until that trace-based repro is attached, I’d prefer not to merge this because the change affects the direct video generation output path and the main value of the fix is specifically eliminating that late D2H transfer.**

Satyam-53:
I inspected the pre-fix trace and two post-fix traces. The expected large D2H elimination is **not** observed.

| Trace | Final D2H size | Duration |
|---|---:|---:|
| pre-fix | 388,177,920 bytes | 3.301 s |
| post-fix | 388,177,920 bytes | 3.467 s |
| post-fix rerun | 388,177,920 bytes | 3.311 s |

In both post-fix traces, there is still a **370.195 MiB device-to-host copy** near the end of generation on stream 7. It copies from device memory to pageable host memory through `cudaMemcpyAsync`, and the API call remains occupied for approximately the full transfer duration.

The byte count exactly matches a float32 decoded video tensor with shape `[1, 3, 81, 480, 832]`, so this appears to be the decoded output materialization that the PR intends to skip—not an incidental metadata copy.

Therefore, the current trace evidence indicates that the PR does **not** eliminate the late D2H transfer for the tested workload. The post-fix timing difference is only run-to-run variation; the transfer itself remains present with the same size.

Could you please verify that the tested direct `VideoGenerator` request actually reaches this branch with `save_video=False` and `return_frames=False`, and investigate where the remaining CPU materialization is triggered?

I would keep this blocked until a post-fix trace shows this 388 MB transfer disappearing.

macthecadillac:
Investigated the remaining 388,177,920-byte D2H copy from the direct `VideoGenerator` workload.

The request does reach `save_video=False, return_frames=False`. The missing ownership boundary was earlier than `VideoGenerator`: `Worker.execute_forward` still returned the full decoded CUDA tensor through multiprocessing/Ray, so transport materialized it before the entrypoint could skip its own `.cpu()` path.

Current head `d61ad00bf` now drops unused decoded output inside the worker before transport. Follow-up fixes preserve the `None` failure sentinel, align save-only latent/audio behavior with actual consumers, and propagate Ray `extra` metadata. The exact 480x832x81 one-step multiprocessing workload completed with `samples=None`, `frames=None`, and `video_path=None`; final-head focused tests are 34 passed and full pre-commit is green on Modal and GitHub.

I also attempted the requested `nsys` capture on Modal with Nsight Systems 2025.5.2 and CUDA Nsight Systems 2024.6.2. Both failed during multiprocessing CUDA trace finalization with `PostMortemTimeConverter::ConvertGpuTicksToSyncNs ... InternalErrorException` and produced no usable report, so I cannot claim trace confirmation from that environment.

Could you rerun the same A/B trace against current head `d61ad00bf` in the Nsight environment that produced your earlier reports? The expected result is that the final 388,177,920-byte transfer is absent. If it remains, please share the updated trace details and I will continue from that boundary.

Satyam-53:
Thank you for this distributed worker fixes, I will check with the current head.

Satyam-53:
I am still seeing the large D2H transfer towards the end in the nsys sqlite file.

Satyam-53:
I tested the latest changes, they work.
@macthecadillac sorry for the last comment, I mistakenly ran it on a different config file.

Satyam-53 (commented):
A recently introduced unit test- test_every_test_directory_is_collected_or_allowlisted is failing on this PR since the test_gpu_worker.py introduced inside the worker folder as part of this pr is either not collected or allowlisted, can you please fix this @macthecadillac ?

Satyam-53:
> A recently introduced unit test- test_every_test_directory_is_collected_or_allowlisted is failing on this PR since the test_gpu_worker.py introduced inside the worker folder as part of this pr is either not collected or allowlisted, can you please fix this @macthecadillac ?

Resolved in the commit - [b302ba8](https://github.com/hao-ai-lab/FastVideo/pull/1567/commits/b302ba818dff9df8ddd1425b749938056211fd26)

Satyam-53 (approved):

Issue #1576, opened 2026-07-10
[Bug] Compatibility issue with `transformers==5.12.x` for Cosmos 2.5 / Qwen2.5-VL


klhhhhh:
## Description

While trying to run the Cosmos 2.5 model in FastVideo using `transformers==5.12.x`, I encountered a series of initialization failures related to the custom Qwen2.5-VL implementation.

The current implementation uses a customized `qwen2_5_vl_custom.py`, while importing the official HuggingFace configuration:

```python
from transformers.models.qwen2_5_vl.configuration_qwen2_5_vl import (
    Qwen2_5_VLConfig,
    Qwen2_5_VLVisionConfig,
)
```

However, the configuration API in `transformers==5.12.x` has changed compared to the version this implementation appears to be based on.

As a result, the custom model implementation and the latest configuration class are no longer fully compatible.

---

### 1. Missing `pad_token_id`

```
AttributeError:
'Qwen2_5_VLConfig' object has no attribute 'pad_token_id'
```

In transformers 5.12:

```python
config.text_config.pad_token_id
```

exists, while

```python
config.pad_token_id
```

does not.

---

### 2. Missing `vocab_size`

```
AttributeError:
'Qwen2_5_VLConfig' object has no attribute 'vocab_size'
```

Similarly,

```python
config.text_config.vocab_size
```

exists, but

```python
config.vocab_size
```

does not.

---

### 3. `rope_scaling` / `rope_parameters` incompatibility

After patching the previous issues, another initialization failure occurs:

```
AttributeError:
'Qwen2_5_VLConfig' object has no attribute 'rope_parameters'
```

The custom implementation accesses

```python
config.rope_scaling
```

while in transformers 5.12 this is implemented as a property backed by `rope_parameters`, making it incompatible with the assumptions in the current implementation.

---

## Root cause

The current `qwen2_5_vl_custom.py` appears to be forked from an earlier HuggingFace implementation, while the configuration class is imported from the latest transformers release.

This results in a mismatch between:

- the custom model implementation (older config API)
- the latest `Qwen2_5_VLConfig` (new config API)

leading to multiple AttributeErrors during model initialization.

---

## Suggested solution

Rather than patching individual missing attributes, it may be preferable to introduce a compatibility layer for `transformers>=5.x`, for example:

- Normalize fields from `text_config` onto the top-level config expected by the custom implementation.
- Handle both `rope_scaling` and `rope_parameters` to support different transformers versions.
[...]

PR #1577, opened 2026-07-10
[ci]: cover transformers 5.x config compat for Qwen2.5-VL (#1576)

Mister-Raggs:
## Problem

`transformers` 5.x changed the `Qwen2_5_VLConfig` API in four ways that all break the custom Qwen2.5-VL encoder used by Cosmos 2.5's Reason1 text encoder:

1. The composite config no longer forwards text-model attributes (`vocab_size`, `pad_token_id`, `hidden_size`, ...) down from `text_config`.
2. `rope_scaling` became a property backed by `rope_parameters`, so it *raises* instead of returning `None`.
3. `"default"` was dropped from `ROPE_INIT_FUNCTIONS`.
4. `rope_theta` moved inside `rope_parameters`.

`fastvideo/models/encoders/qwen2_5_vl_custom.py` already absorbs all four, via `_flatten_text_config()`, a module-local `_compute_default_rope_parameters()`, and a `getattr` guard on `pad_token_id`. But that compat layer arrived incidentally as part of an unrelated pipeline PR (`02e1143f`, #1471) and **has no test coverage**.

Meanwhile #1431 relaxed the pin from `transformers==4.57.3` to `transformers>=4.57.3`, and no lockfile or CI config re-pins it. So a fresh install today resolves to transformers 5.13.0: every new user is on the v5 path, and the only thing holding it up is an untested helper that arrived as a side effect of a Kandinsky PR.

This surfaced as #1576, filed against a checkout predating `02e1143f`.

## Change

Adds `fastvideo/tests/encoders/test_qwen2_5_vl_config_compat.py` — five tests pinning both the config normalization and the resulting RoPE numerics. One covers the `reason1.py` fallback config, which carries no `rope_scaling` and takes a different branch on each transformers line.

No production code changes. No `pyproject.toml` change: the `>=4.57.3` floor is correct and 5.x genuinely works, so capping it would pin users off a version the code supports.

Unlike the existing encoder tests, this one is **CPU-only and weight-free** — the config is built inline at the `Qwen2.5-VL-7B-Instruct` shape rather than downloaded, so it costs nothing in CI.

## Test evidence

Passing on transformers 4.57.3, 5.12.1 and 5.13.0 (torch 2.13 CPU):

```
$ pytest fastvideo/tests/encoders/test_qwen2_5_vl_config_compat.py -q
5 passed
```

The tests assert more than "does not raise". Against the real `Qwen/Qwen2.5-VL-7B-Instruct` config, the RoPE parameters are identical across all three versions — base `1000000.0`, `head_dim` 128, `inv_freq` dim 128, `mrope_section` `[16, 24, 24]`, `vocab_size` 152064 — so the compat layer is behavior-preserving, and the tests pin that rather than just the absence of an exception.

Mutation-checked, to confirm the tests can actually fail: neutering `_flatten_text_config` to a no-op (simulating the pre-`02e1143f` code) fails all five on 5.12.1, reproducing the exact `AttributeError`s from #1576:

```
FAILED test_flatten_text_config_exposes_text_attrs_on_composite_config
FAILED test_flatten_text_config_preserves_mrope_settings
FAILED test_compute_default_rope_parameters_reads_theta_on_any_transformers
FAILED test_rotary_embedding_handles_config_without_rope_scaling
FAILED test_rotary_embedding_builds_from_flattened_config
5 failed
```

The `rope_scaling`-less fallback fails too: on 5.x that config also loses `max_position_embeddings` from the top level, which `Qwen2_5_VLRotaryEmbedding` reads directly.

## Reviewer note on scope

The same mutation still **passes** under 4.57.3, because 4.x forwards those attributes natively. This suite therefore only guards the transformers 5.x path. That is the path CI and fresh installs land on given the unbounded pin, so it's where the coverage is needed — but it is worth knowing the guard is version-dependent rather than assuming it protects both.

Related: #1576

### DIFF
--- a/fastvideo/tests/encoders/test_qwen2_5_vl_config_compat.py
+++ b/fastvideo/tests/encoders/test_qwen2_5_vl_config_compat.py
@@ -1,0 +1,125 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Config-compat regression tests for the custom Qwen2.5-VL encoder.
+
+transformers 5.x stopped forwarding text-model attributes (``vocab_size``,
+``pad_token_id``, ``hidden_size``, ...) from the composite ``Qwen2_5_VLConfig``
+down from ``text_config``, turned ``rope_scaling`` into a property backed by
+``rope_parameters``, dropped the ``"default"`` entry from ``ROPE_INIT_FUNCTIONS``
+and moved ``rope_theta`` inside ``rope_parameters``. See issue #1576.
+
+``qwen2_5_vl_custom`` absorbs all of that. These tests pin both the config
+normalization and the resulting RoPE numerics, so the compat layer cannot
+silently regress as transformers evolves. They are CPU-only and weight-free:
+the config is built inline rather than downloaded.
+"""
+import torch
+from transformers import Qwen2_5_VLConfig
+
+from fastvideo.models.encoders.qwen2_5_vl_custom import (
+    Qwen2_5_VLRotaryEmbedding,
+    _compute_default_rope_parameters,
+    _flatten_text_config,
+)
+
+# Shape of Qwen/Qwen2.5-VL-7B-Instruct, the checkpoint Cosmos 2.5's Reason1 text encoder is built on.
+HIDDEN_SIZE = 3584
+NUM_ATTENTION_HEADS = 28
+VOCAB_SIZE = 152064
+ROPE_THETA = 1000000.0
+MROPE_SECTION = [16, 24, 24]
+HEAD_DIM = HIDDEN_SIZE // NUM_ATTENTION_HEADS  # 128
+
+
+def _build_config() -> Qwen2_5_VLConfig:
+    # Flat, 4.x-style kwargs: this is what a real checkpoint's config.json carries.
+    return Qwen2_5_VLConfig(
+        hidden_size=HIDDEN_SIZE,
+        intermediate_size=18944,
+        num_attention_heads=NUM_ATTENTION_HEADS,
+        num_hidden_layers=28,
+        num_key_value_heads=4,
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
Rebased onto current main after build 4265 spent most of its encoder budget in clone/bootstrap before reaching the existing Qwen tests; no content changes (range-diff clean).

SolitaryThinker:
Infra triage update for exact head `c0a4f80752068c7c9b78f5e9fcb405fb05f2b268`:

- Encoder build 4385 failed only the existing Cosmos T5 setup while all five PR-added Qwen compatibility tests passed.
- The isolated encoder retry (build 4391) reproduced the same Hugging Face Xet CAS reconstruction `401 Unauthorized`; 17 other encoder tests passed.

This is now confirmed as an infra download/authentication failure, not a PR regression. The retry budget is exhausted, so no branch change or further rerun is requested.

SolitaryThinker:
Maintainer reconciliation at `c0a4f807`: I resolved the three outdated bot threads because the current head already uses the public top-level `Qwen2_5_VLConfig` import, documents/asserts the supported normalization behavior, and adds the requested no-`rope_scaling` fallback coverage.

The remaining Fastcheck red is not a branch defect. [Build 4385](https://buildkite.com/fastvideo/ci/builds/4385) and the targeted [encoder retry 4391](https://buildkite.com/fastvideo/ci/builds/4391) both fail only while reconstructing the existing Cosmos T5 snapshot through HF Xet CAS (`401 Unauthorized`); all five PR-added Qwen compatibility tests pass. The retry budget is exhausted, so I am not launching another paid job or asking for a code change. We can resume when the external snapshot path is healthy or a main-side fix makes a fresh run meaningful.

PR #1579, opened 2026-07-10
[new-model] Port LingBot-World-v2

Davids048:
## Summary

Adds native FastVideo support for the 14B causal-fast image-to-video model from
[LingBot-World-v2](https://github.com/robbyant/lingbot-world-v2).

- Ports the causal-fast DiT, camera conditioning, UMT5 encoder, Wan2.1 VAE, and
  chunked inference pipeline into FastVideo.
- Registers the `lingbotworld2` model family, pipeline, and inference preset.
- Adds a checkpoint conversion script for the released
  `robbyant/lingbot-world-v2-14b-causal-fast` checkpoint.
- Adds an inference example with a self-contained image and camera dataset.
- Keeps inference native to FastVideo without importing the LingBot-World-v2
  source repository at runtime.

## Numerical parity

The official repository and FastVideo generated videos using the same prompts,
input image, camera poses, intrinsics, random seed, and generation settings:

| Setting                     | Value                            |
| --------------------------- | -------------------------------- |
| Model                       | LingBot-World-v2 14B causal-fast |
| Hardware                    | 8 NVIDIA H200 GPUs               |
| Parallelism                 | Sequence parallelism 8 with FSDP |
| Requested resolution        | `480x832`                        |
| Decoded resolution          | `464x832`                        |
| Local attention size        | `18`                             |
| Sink size                   | `6`                              |
| Temporal chunk size         | `4`                              |
| Causal-fast denoising steps | `4`                              |
| Output frame rate           | `16 FPS`                         |

Decoded `uint8` video frames were compared element by element:

| Requested frames | Maximum absolute error | Mean absolute error |
| ---------------- | ---------------------- | ------------------- |
| `17`             | `0`                    | `0.0`               |
| `33`             | `0`                    | `0.0`               |
| `65`             | `0`                    | `0.0`               |

Each frame count was tested with three different prompts, producing nine
exactly aligned reference and FastVideo video pairs.

The 5-frame and 9-frame cases are not supported by the official repository.
With its temporal compression and chunk size of 4, these requests do not
produce a valid latent chunk.

### Pair 1: lakeside prompt, 65 frames requested

Reference 

https://github.com/user-attachments/assets/f7775a20-d9b2-497b-abc9-1705cd07fa29

FastVideo

https://github.com/user-attachments/assets/3f409dde-edb7-4b70-8d52-5c84c678e82c

### Pair 2: coastal-village prompt, 33 frames requested

| Reference | FastVideo |
| --------- | --------- |
|  |  |


tests/local_tests/lingbotworld2/PORT_STATUS.md
Davids048, line 1: Cleanup

SolitaryThinker (changes requested):
Full-port review at b6aae87 (six rubric lanes, every finding adversarially verified). The parity work is impressive — decoded-frame 0-max-abs-error against the official repo is the strongest evidence a port can bring. But the PR isn't landable yet.

**Blockers**
1. **No user-loadable weights.** `fastvideo/registry.py` points at the raw official `robbyant/lingbot-world-v2-14b-causal-fast` layout, which FastVideo cannot load, and the conversion script emits a symlink-only local bundle (non-portable, non-idempotent, zero verification — no strict-load, no config validation, no skipped-key report). Make the converter emit a portable verified bundle and point the registry at a loadable repo.
2. **License.** `examples/dataset/lingbotworld2/` (image.jpg, poses.npy, intrinsics.npy) is copied verbatim from the official repo, which is CC BY-NC-SA 4.0 — incompatible with Apache-2.0 FastVideo. Replace with self-made or compatibly-licensed assets.
3. **No committed parity evidence.** Zero component parity tests (DiT, VAE, T5, cam_utils, scheduler); the recorded encoder "parity" is strict-load only; the pipeline parity test wraps uncommitted scripts on your cluster and skips everywhere else, so the 0-error claim is unauditable as committed. Commit the compare methodology plus per-component parity tests per the add-model gates.
4. **Red unit lane is PR-caused.** The new `action_path` field needs classification in `docs/design/inference_schema_parity_inventory.yaml` and the roundtrip dict in `fastvideo/tests/api/test_parser.py` (that's both failing tests).

**Majors**
5. Attention bypasses the FastVideo registry — hand-rolled `flash_attn` dispatch and raw `nn.Linear`. `causal_wanvideo.py` shows `LocalAttention`+`ReplicatedLinear` handling this exact rolling-KV-cache pattern. Related: TORCH_SDPA is advertised, but cross-attention calls `flash_attention` directly, so the SDPA backend crashes.
6. Component duplication: `lingbotworld2_wanvae.py` is a third in-tree Wan2.1 VAE — `Cosmos25WanVAE` has identical arch constants, latent stats, and the same `handles_latent_denorm` contract (PORT_STATUS rejects `AutoencoderKLWan` but never evaluated it). `lingbotworld2_t5.py` forks the official T5, duplicating native `UMT5EncoderModel` with no reuse-rejection evidence.
7. The committed example can't run: hardcoded author-cluster weights path, and `generate_video` never passes `image_path`, so the pipeline raises ValueError.
8. Sampling defaults (sample_shift/timesteps) live on DiTArchConfig, leaving pipeline `flow_shift` a dead knob.
9. No SSIM/media regression added and no explicit deferral recorded.
10. pre-commit is red at head.

**CI attribution:** the encoder-lane exit-124 (`test_qwen2_5_encoder`) is NOT this PR — your `component_loader.py` hunk is gated on the LingBotWorld2 VAE branch, which that lane never executes; that's Modal host slowness. The unit-lane failures are yours (item 4).

**Two-stack note:** everything here is additive inference surface; no existing model's forward path changes. Item 5 is what will let the training stack share this forward later — worth doing now, not after.

Happy to help stage weight publishing to an org repo once the converter emits a portable bundle.

SolitaryThinker:
I carried the mechanical portion of the review at the author's request.

What changed:

- Rebasing onto current `main` resolved the only conflict by retaining main's Flux/GLM exports and adding LingBot World 2. The first three commits are range-diff identical; the final commit differs only by that union resolution.
- Classified `SamplingParam.action_path` in the typed-schema inventory and updated the parser roundtrip expectation, clearing the PR-caused unit failure.
- Made the example take its converted bundle from `LINGBOTWORLD2_MODEL_PATH` instead of a private `/mnt/weka` path, and passed the missing I2V `image_path`.
- Applied the repository formatter to the two files it flagged.

Why: these are mechanical CI/reproducibility fixes and do not alter the transformer, VAE, encoder, conversion math, or claimed parity.

Verification at `7568495e4`:

- `pre-commit run --files ...` — passed (yapf, ruff, codespell, mypy, filename checks).
- `fastvideo/tests/api/test_parser.py` — 8 passed.
- LingBot World 2 `request_preserves_action_path` preflight — passed.
- The schema-inventory test is left to CI because the local `fv` environment lacks `aiofiles`; an independent static equality check confirms all 78 `SamplingParam` fields are classified and `action_path` targets `request.inputs.action_path`.

Still intentionally contributor-owned: license-clean example assets, committed official/native parity evidence, loadable published weights, attention-backend integration, and SSIM/explicit deferral. I did not arm merge or approve the PR.

SolitaryThinker (approved):

PR #1580, opened 2026-07-10
[bugfix]: propagate Qwen2.5-VL visual dtype

### LINKED ISSUE #1578: [Bug] Qwen2.5-VL vision tower always resolves to float32: torch_dtype read from vision_config, where it is never set
### Describe the bug

In `fastvideo/models/encoders/qwen2_5_vl_custom.py`, the vision tower resolves its working dtype from the **vision** sub-config:

```python
class Qwen2_5_VisionTransformerPretrainedModel(nn.Module):
    def __init__(self, config) -> None:
        ...
        self.dtype = torch.bfloat16 if config.torch_dtype == "bfloat16" else torch.float32   # line 310
```

But `torch_dtype` is declared at the **top level** of the Qwen2.5-VL `config.json`, not inside `vision_config`. So `config.torch_dtype` is `None` here and the comparison is always `False` — `self.dtype` silently resolves to `torch.float32` even for a bf16 checkpoint.

This is not dead code. `self.visual.dtype` is what casts the pixel inputs:

```python
pixel_values = pixel_values.type(self.visual.dtype)          # line 1745
pixel_values_videos = pixel_values_videos.type(self.visual.dtype)  # line 1763
```

so image/video inputs get cast to fp32 while the vision weights are bf16.

### Evidence

Loading the real `Qwen/Qwen2.5-VL-7B-Instruct` `config.json` and reading both configs:

| | value | type | `== "bfloat16"` |
|---|---|---|---|
| `config.torch_dtype` (composite) | `'bfloat16'` | `str` | `True` |
| `config.vision_config.torch_dtype` | `None` | `NoneType` | `False` |

`config.json` has `"torch_dtype": "bfloat16"` at the top level, and its `vision_config` block has no `torch_dtype` key at all.

Reproduces identically on transformers **4.57.3** and **5.12.1**, so this is pre-existing and *not* related to the transformers 5.x config changes in #1576.

There is a second, latent version of the same mistake: the comparison is against the **string** `"bfloat16"`. Some transformers versions materialize `config.torch_dtype` as a `torch.dtype` object rather than a string, in which case the check would be `False` even when the attribute is populated.

### Impact

The cast only runs when `pixel_values` / `pixel_values_videos` are passed. Cosmos 2.5 uses Reason1 as a text encoder, so I believe the image path is currently dormant in FastVideo's pipelines — which is likely why this hasn't bitten anyone. I have **not** executed the image path against real weights, so I can't say whether the practical result is a dtype-mismatch `RuntimeError` at the patch-embed conv or a silent fp32 upcast of the vision tower.

Flagging it rather than fixing it blind, since the right fix depends on intent: read `torch_dtype` off the parent config, compare against `torch.bfloat16` as well as the string, or drop `self.dtype` in favor of the module's actual parameter dtype.

### Reproduction

```python
import json, urllib.request
from transformers.models.qwen2_5_vl.configuration_qwen2_5_vl import Qwen2_5_VLConfig

raw = json.loads(urllib.request.urlopen(
    "https://huggingface.co/Qwen/Qwen2.5-VL-7B-Instruct/raw/main/config.json").read())
cfg = Qwen2_5_VLConfig(**raw)

print(cfg.torch_dtype, type(cfg.torch_dtype))                # 'bfloat16' <class 'str'>
print(cfg.vision_config.torch_dtype)                         # None
print(cfg.vision_config.torch_dtype == "bfloat16")           # False  -> self.dtype = torch.float32
```

### Environment

Config-level reproduction; no GPU required. Verified against transformers 4.57.3 and 5.12.1.

macthecadillac:
## Summary

Fixes #1578 by resolving the Qwen2.5-VL visual tower dtype from the parent config when `vision_config.torch_dtype` is absent.

This change:
- normalizes `torch_dtype` values supplied as strings or `torch.dtype` objects
- keeps explicit `vision_config.torch_dtype` as the highest-priority source
- falls back to parent `config.torch_dtype` for real Qwen2.5-VL configs where dtype is top-level only
- preserves the existing fp32 fallback for missing or unknown dtype values
- adds CPU-only regression coverage for parent dtype propagation, dtype-object handling, vision override behavior, whitespace normalization, and fp32 fallback
- keeps the tiny full-model fixture compatible with the current Transformers RoPE initializer by defining Qwen2.5-VL's `rope_theta`

## Validation

- Modal L40S targeted pytest against exact pushed head `310ca429f4578f2aa6b6d4c297f162ae3f754516`:
  - `pytest fastvideo/tests/encoders/test_qwen2_5_vl_vision_dtype.py -q`
  - `6 passed in 0.16s`
  - App: https://modal.com/apps/hao-ai-lab/main/ap-cUvC2R8uMCr7euc1Hrxhio
- Modal L40S pre-commit gate with the fixture update applied:
  - `pre-commit run --all-files`
  - passed
  - App: https://modal.com/apps/hao-ai-lab/main/ap-m33DGcyhT3G7YEIqTdh6Hf
- GitHub Actions pre-commit against exact pushed head `310ca429f4578f2aa6b6d4c297f162ae3f754516`:
  - passed
- Earlier Modal L40S targeted pytest for the accepted Gemini whitespace-normalization fix:
  - `6 passed, 16 warnings in 0.24s`
  - App: https://modal.com/apps/hao-ai-lab/main/ap-DuKfh6m4CWOvTOuTCO5A5p
- Full Buildkite suite against exact pushed head `310ca429f4578f2aa6b6d4c297f162ae3f754516`:
  - all lanes passed, including SSIM, LoRA training, standard training, and VSA training
  - Build: https://buildkite.com/fastvideo/ci/builds/4330
- Stage 3 review loop:
  - the original fresh `review-code` sub-agent reported no actionable findings
  - a second fresh `review-code` sub-agent reviewed exact pushed head `310ca429f` after the rebase and fixture update and reported no actionable findings
  - no adjudicator/fixer sub-agent was needed because neither review reported actionable findings
- Bot review:
  - accepted Gemini's suggestion to strip whitespace from dtype strings before normalization
  - added regression coverage for whitespace-padded parent dtype strings
  - the addressed review conversation is resolved

## GPU Memory Impact

No expected increase. For Qwen2.5-VL visual inputs, this avoids resolving `self.visual.dtype` to fp32 when the parent config declares bf16. The current text-only Reason1 path remains unchanged.

## Notes

- The branch was rebased onto current `upstream/main` to include the full-suite gate script required by GitHub pre-commit; `git range-diff` confirmed the original six-commit patch series was unchanged.
- Wan T2V SSIM was not run because this change is limited to Qwen2.5-VL encoder dtype plumbing and does not affect Wan generated pixels.
- The support matrix was not updated because no new model or user-facing support claim was added.

# Checklist


SolitaryThinker:
Reviewed. Verified against #1578: real Qwen2.5-VL checkpoints never set torch_dtype on vision_config, so the tower silently resolved fp32; the parent-config fallback plus torch.dtype-object handling fixes both failure modes, and the new CPU tests fail without the fix.

Blast radius: nil for existing paths — `visual.dtype` is only consumed on the pixel-input path, which no FastVideo caller exercises today (Reason1 is text-only for Cosmos 2.5 in both training stacks, and its load_weights skips visual weights entirely).

Two non-blocking notes for later: deriving dtype from actual params (`patch_embed.proj.weight.dtype`) would be more robust, since the loader — not the HF config — sets real precision; and transformers 5.x renames config `torch_dtype` to `dtype`, so a `getattr(config, "dtype", ...)` fallback is worth adding when that alias goes away.

Your pre-commit run fired while the PR was draft and got skipped; re-running it now, then the full suite.

SolitaryThinker:
@macthecadillac your pre-commit failure isn't your code — main's #1572 added a "Full-suite gate self-test" step that runs `.github/scripts/test_gate_full_suite.sh` from the PR-head checkout, and your branch predates that script (exit 127). A rebase onto current main fixes it.

I've staged the exact pure rebase at `maint/pr1580-rebase-only` (head 20c5c900, range-diff clean — all 6 commits unchanged; the classifier that guards this repo wouldn't let me force-push your fork directly). Either:

```bash
git fetch upstream main
git rebase upstream/main
git push --force-with-lease
```

or just reset your branch to `upstream/maint/pr1580-rebase-only` and push. The suite will re-run after the push.

PR #1560, opened 2026-07-06
[ci] Add exact identity performance statuses

macthecadillac:
## Summary

Fixes #1532 by making rolling performance comparisons use explicit statuses and exact v2 comparison identities.

This change:
- emits `PASS`, `REGRESSION`, `CALIBRATION_NEEDED`, `RECIPE_MISMATCH`, and `INFRA_ERROR`, with `QUALITY_BLOCKED` reserved for promotion policy
- compares v2 records only when all six identity fields match: `workload_id`, `variant_id`, `benchmark_version`, `hardware_profile_id`, `software_profile_id`, and `recipe_fingerprint`
- scans exact v2 cohorts independently of legacy `model_id` directories and GPU display strings, while current v1 artifacts remain static-threshold-only
- rejects partial v2 identities as infrastructure errors and preserves structural identity/recipe status precedence
- attributes static-threshold regressions to the measured record without contaminating passing sibling records
- excludes display-only `benchmark_id` changes from recipe fingerprint canonicalization while retaining the value as audit metadata
- keeps calibration artifacts visible but ineligible for automatic rolling-baseline promotion
- provides a reviewed baseline-seeding path with all-source provenance, identity, measurement, uniqueness, and 5% batch-consistency preflight before persistence
- records runtime software identity and the resolved container image reference consistently in performance CI audit metadata
- updates contributor guidance and focused regression coverage across comparator, producer, storage, seeding, identity, and result-schema behavior

## Why

The prior comparator could select unrelated history through legacy storage labels, did not distinguish missing calibration from true regressions, and lacked a safe path to seed the first exact-identity baseline. The v2 path now makes cohort selection and status remediation explicit while preventing malformed or inconsistent calibration data from entering the rolling baseline.

## Validation

- Exact pushed SHA `c1c660aee4a2049f57682ee7a9c8eb0995cb7913` on Modal L40S, with no local patch:
  - `pytest fastvideo/tests/performance/test_compare_baseline_policy.py fastvideo/tests/performance/test_inference_performance_identity.py fastvideo/tests/performance/test_identity.py fastvideo/tests/performance/test_inference_performance_result_schema.py -q`
  - `106 passed in 1.76s`
  - App: https://modal.com/apps/hao-ai-lab/main/ap-PaeGVJ0mEWJOjyKDBxSqH0
- Modal L40S full pre-commit on the final patch:
  - `pre-commit run --all-files`
  - all hooks passed
[...]

PR #1591, opened 2026-07-12
[ci]: pin VSA training regression to H100

### LINKED ISSUE #1586: [CI] VSA training lane flaky on H200: anomalous H200 grad_norm reference + unpinned GPU (exit 1)
## Summary

`VSA/test_training_loss_VSA.py::test_distributed_training` fails its `grad_norm` check whenever the run lands on an **H200**, because the checked-in H200 reference (`grad_norm = 0.245`) is anomalous — it disagrees with both the H100 reference and current runs by ~5×. The lane is also nondeterministic: the Modal function requests `H100:2` but Modal substitutes H200 under capacity pressure, so the same commit passes or fails depending on which GPU it lands on. This reddens PRs whose diffs have nothing to do with VSA training.

## Evidence

The two references were committed together (#1352, 2026-05-16) and represent the *same* training run on two Hopper GPUs:

| field | H100 ref | H200 ref |
|---|---|---|
| **grad_norm** | **1.118** | **0.245** |
| train_loss | 0.277 | 0.308 |
| step_time | 0.70 | 0.83 |
| avg_step_time | 1.00 | 1.41 |

`train_loss` and `step_time` are neighborly, but `grad_norm` is **4.6× apart** — implausible for the same training on H100 vs H200. A recent failing run on H200 reported `grad_norm = 1.237`, which matches the **H100** reference (diff 0.12, within the 0.5 threshold), not the H200 one (diff 0.99). So the true H200 grad_norm is ~1.2 like H100, and the recorded `0.245` is an anomalous/bad value rather than a real baseline. The kernel has been stable at ~1.1–1.2 the whole time — this is a bad reference number, not a regression the check is correctly catching.

## Nondeterminism (why it's flaky, not consistently red)

`run_training_tests_VSA` requests `gpu="H100:2"`, but the test selects its reference at runtime from `torch.cuda.get_device_name()`:
```python
if "H100" in device_name:  reference = h100_reference_wandb_summary_file
elif "H200" in device_name: reference = h200_reference_wandb_summary_file
```
Modal serves H200 for `H100:2` requests when H100 capacity is tight, so the lane compares against the (good) H100 ref on some runs and the (bad) H200 ref on others. Same commit, different outcome — which is why the lane passes on some PRs and fails on others (e.g. it passed on one unrelated PR and failed on another in the same period).

## Reproduction

Any PR whose CI happens to land the VSA training lane on an H200. The failure is:
```
AssertionError: FAILED: grad_norm difference 0.99… exceeds threshold of 0.5
  (reference: 0.2447…, current: 1.237…)
```

## Suggested solutions (in rough order of preference)

1. **Pin the VSA lane's GPU.** Set the modal function to a single GPU type that CI can reliably get, so the test only ever compares against one reference and stops depending on Modal's allocation. Removes the flakiness at the source and sidesteps maintaining two hardware-specific baselines against an allocator you don't control.
2. **Regenerate the H200 reference** on an actual H200 (`gpu="H200:2"`) against the current kernel and replace the anomalous `0.245`. Low-risk given the evidence above (it's a correction, not a re-baseline that hides drift), but leaves the two-reference/allocation coupling in place.
3. **Collapse to a single tolerant reference** if H100/H200 grad_norm genuinely track each other (the data suggests they do, ~1.1–1.2), removing the per-device split entirely.

Happy to prep whichever direction — e.g. a one-off `H200:2` regeneration function, or the GPU-pin change.

## Environment
- `hao-ai-lab/FastVideo`, `fastvideo/tests/training/VSA/`, observed 2026-07-10.
- References last changed 2026-05-16 (#1352); unchanged since.

macthecadillac:
## Summary

- request a strict two-GPU H100 allocation for the VSA training lane so Modal cannot upgrade it to H200
- remove the anomalous H200 W&B reference and make the regression explicitly H100-only
- add contract coverage that locks the Modal function to `H100!:2`

Fixes #1586.

## Problem

Modal may automatically upgrade `gpu="H100:2"` requests to H200. The VSA training test then selected a checked-in H200 reference whose `grad_norm` was about 0.245, while current H100/H200 runs produce values around 1.2. The same commit therefore passed on H100 and failed on H200 depending on allocation.

## Validation

- Modal L40S contract test: `4 passed in 0.03s` (`ap-dGXu4f3p0Bd5k1ICYiEYDX`)
- Modal 2x H100 VSA regression: `1 passed, 1 warning in 155.66s` (`ap-JYFvkbsICoWEeXD9ulz1bL`)
- Both GPUs reported `NVIDIA H100 80GB HBM3`
- Exact differences against the unchanged H100 reference:
  - `avg_step_time`: 0.5718568257969423 (threshold 15)
  - `grad_norm`: 0.11541739463806144 (threshold 0.5)
  - `step_time`: 0.19428527497333903 (threshold 15)
  - `train_loss`: 0.033021042466163664 (threshold 0.04)
- `pre-commit run --all-files`: passed
- Independent Stage 3 review: no actionable findings

The credential-free Modal validation used W&B offline storage plus a temporary uncommitted summary reader. That adapter was removed after validation; production online-W&B behavior is unchanged.

## GPU Memory Impact

No expected change. The lane still uses two Hopper GPUs and the patch changes only allocation policy and test/reference selection. Strict H100 allocation can reduce scheduling flexibility but does not increase per-GPU memory use.

## Remaining Risk

- The Buildkite wrapper with online W&B secrets was not executed end to end because those CI credentials are unavailable locally; that wiring is unchanged.
- The H100 `train_loss` result remains relatively close to its existing threshold, but neither the reference nor threshold changed.

# Checklist


SolitaryThinker:
Maintainer review: this is the correct root-cause fix, and it also explains a flake we'd been tracking as "bimodal VSA grad_norm" (retry-clears): the retry was just re-rolling Modal's GPU substitution lottery — current runs report grad_norm ~1.24, consistent with the H100 reference (1.118) and ~5× off the anomalous H200 reference (0.245) this PR deletes.

Verified details:
- `H100!:2` correctly targets `run_training_tests_VSA` (checked the decorator on the PR head), and `!` is Modal's strict no-substitution suffix.
- The contract test works as written: `_function_strings` walks the FunctionDef including its decorator constants, so `H100!:2` appears as its own line.
- Hard-failing on non-H100 with a clear message is right for a regression lane — better a loud capacity error than a silent wrong-reference comparison.

Two notes, neither blocking:
1. The VSA lane only runs at full-suite scope, so this PR's own fastcheck won't exercise the strict pin — first validation happens on the first full-suite run after merge (or a pre-merge `/test full`).
2. `fastvideo/tests/training/Vanilla/h200_reference_wandb_summary.json` still exists with the same dual-reference pattern on an unpinned lane — if its H200 ref is also anomalous the same flake lurks there; worth a follow-up check (out of scope here).

macthecadillac:
Thanks for the detailed review and for confirming the root cause.

I ran `/test full`; Buildkite build 4334 passed, including the VSA training lane in 10m51s. That validates the strict H100 allocation through the production CI and W&B path.

I also opened #1592 to track the Vanilla lane separately. Its checked-in H200 and L40S losses are close and the grad-norm difference is currently within the configured threshold, so I framed it as an investigation: reproduce on strict L40S and H200 allocations, inspect recent CI history, then decide whether to retain the references, refresh one, or pin the lane.

I'll keep that work out of this PR.

SolitaryThinker:
Rebased onto current `main` (`a2538561`, including #1560) to restore mergeability; no PR-content changes (range-diff clean—three commits are `=`, and the sole conflicted commit differs only because main added two adjacent contract tests).

Satyam-53 (commented):
I verified the changes here and they look fine to me. However this pull completely removes the support for H200, I think we should pin to H100 gpu but still kepp the h200_reference_wandb file in the codebase for future use. 

Apart from that, @SolitaryThinker are we waiting on anything else to merge this PR?

Satyam-53 (approved):

PR #1595, opened 2026-07-13
[new-model] Add LingBot-Video Dense and MoE/refiner T2V inference

Davids048:
## Summary

Port LingBot-Video text-to-video inference to FastVideo.

This PR supports both released model layouts:

- **Dense 1.3B**: one Dense transformer generates a 480p video.
- **MoE 30B-A3B**: an MoE transformer generates a 480p base video, with an optional second MoE transformer that refines it to 1080p.

The port adds:

- A shared native transformer implementation for the Dense, MoE base, and MoE refiner checkpoints.
- Text-only Qwen3-VL conditioning with the released LingBot prompt template.
- Flow-UniPC scheduling and the released sampling defaults.
- Dense and MoE/refiner inference presets.
- Native grouped matrix multiplication for the released MoE expert layout.
- A checkpoint converter for the official Dense and MoE layouts.
- Separate `LingBotVideoDensePipeline` and `LingBotVideoMoePipeline` checkpoint identities for unambiguous model discovery. Both identities use the same native FastVideo pipeline implementation.
- A Dense inference example and local parity/test runbook.

The checkpoint converter preserves the official transformer, VAE, and scheduler weights, converts the Qwen3-VL language-model weights to FastVideo’s fused text-encoder layout, and maps the official `refiner/` component to `transformer_2/`.

This PR covers **T2V inference only**. T2I, TI2V, vision conditioning, prompt rewriting, and automatic negative-prompt generation are outside its scope.

### Add `batch_cfg`

Add `batch_cfg` to the typed request schema, sampling parameters, CLI arguments, and internal pipeline batch.

Classifier-free guidance requires a conditional prediction and an unconditional prediction. When `batch_cfg` is enabled, FastVideo:

1. Duplicates the current latent.
2. Packs the conditional and unconditional text inputs into one batch.
3. Runs one transformer forward pass.
4. Splits the two predictions and applies classifier-free guidance.

When it is disabled, FastVideo runs the conditional and unconditional branches as two separate transformer calls.

The global default remains `false`. The LingBot presets enable it by default to match the released production configuration. See [fastvideo/pipelines/basic/lingbot_video/presets.py](fastvideo/pipelines/basic/lingbot_video/presets.py).

The original implementation exposes the same behavior:
https://github.com/robbyant/lingbot-video/blob/a638721cf2271804d02738b69f2ad788c4a559fc/lingbot_video/pipeline_lingbot_video.py#L514-L557

## Patch `VideoGenerator`

Validate `GenerationRequest.stage_overrides` against the preset registered for the selected model before generation begins.

For LingBot-Video, this means:

- `denoise` accepts only `num_inference_steps` and `guidance_scale`.
- `refine` accepts only `height_sr`, `width_sr`, `num_inference_steps_sr`, `guidance_scale_2`, and `t_thresh`.

Unknown stages and fields that belong to a different stage are rejected instead of being silently flattened into the sampling configuration.

`GenerationResult.size` now reports the dimensions of the produced pixel tensor. This matters for the MoE/refiner pipeline because the user requests a 480p base video while the final decoded result is 1080p.

## Mixed-precision FSDP inference

LingBot-Video uses mixed parameter precision inside one transformer:

- Numerically sensitive parameters—including timestep modulation, normalization, and MoE router parameters—remain in `float32`.
- Bulk attention, linear, and expert parameters use `bfloat16`.

Previously, FastVideo’s full-state-dict loader cast every parameter to one configured dtype. FSDP also expected all managed parameters in a mixed-precision group to follow that dtype policy. That could either lose LingBot’s required fp32 parameters or make mixed-dtype FSDP wrapping fail.

This PR adds a model-level parameter-dtype selector and updates the loader to honor it:

- `bfloat16` parameters are loaded, sharded, and managed by FSDP.
- Selected `float32` parameters are excluded from FSDP through `ignored_params` and remain replicated on every inference rank.
- The policy is applied to nested FSDP-wrapped transformer blocks and to the root model.

Mixed-dtype FSDP training is explicitly rejected. Replicated fp32 parameters would require separate gradient synchronization, which this inference port does not implement.

The per-parameter dtype policy matches the original LingBot implementation:

- fp32 parameter selection:
  https://github.com/robbyant/lingbot-video/blob/a638721cf2271804d02738b69f2ad788c4a559fc/lingbot_video/transformer_lingbot_video.py#L46-L64
- preservation of selected fp32 parameters during dtype conversion:
  https://github.com/robbyant/lingbot-video/blob/a638721cf2271804d02738b69f2ad788c4a559fc/lingbot_video/transformer_lingbot_video.py#L968-L997

The `ignored_params` integration and training rejection are FastVideo-specific adaptations for FSDP; they are not copied from the original repository.

## Decode parity

Patch `DecodingStage` to reproduce the released LingBot latent-to-VAE boundary. See [fastvideo/pipelines/stages/decoding.py](fastvideo/pipelines/stages/decoding.py).

The original implementation denormalizes latents using:

```python
latents.float() / (1.0 / std) + mean
```

FastVideo’s generic path used the mathematically equivalent expression:

```python
latents * std + mean
```

The two expressions round differently in float32. The difference appears before VAE decoding and can produce small pixel differences even when the final denoised latent is exact.

The LingBot path now preserves the original reciprocal arithmetic and converts the VAE input to `channels_last_3d`, matching the original decoder boundary:

https://github.com/robbyant/lingbot-video/blob/a638721cf2271804d02738b69f2ad788c4a559fc/lingbot_video/pipeline_lingbot_video.py#L275-L282

https://github.com/robbyant/lingbot-video/blob/a638721cf2271804d02738b69f2ad788c4a559fc/lingbot_video/pipeline_lingbot_video.py#L329-L348

A controlled H200 factor matrix identified the reciprocal denormalization arithmetic as the sufficient cause of the previous pixel difference. The patched production `DecodingStage` reproduced the original decoded pixels with zero differing values.

## Parity status

“Exact” below means `torch.equal(...)` passed with a maximum absolute difference of zero.

| Comparison                                        | Result                                                                 |
| ------------------------------------------------- | ---------------------------------------------------------------------- |
| Qwen3-VL text-only encoder                        | Exact                                                                  |
| Flow-UniPC scheduler                              | Exact                                                                  |
| Dense transformer                                 | Exact                                                                  |
| Dense controlled one- and two-step pipeline       | Exact final latents                                                    |
| Dense two-GPU sequence parallelism                | Exact when both sides use mathematical SDPA                            |
| MoE transformer block                             | Exact                                                                  |
| Complete 48-block MoE base transformer            | Exact                                                                  |
| Complete 48-block MoE refiner transformer         | Exact                                                                  |
| MoE base-only strict test, all 3 official prompts | Exact conditioning, all 40 latent steps, pixels, uint8 frames, and MP4 |
| Dense production pipeline                         | Smoke pass                                                             |
| MoE base production pipeline                      | Smoke pass                                                             |
| MoE plus refiner production pipeline              | Smoke pass; exact end-to-end parity is not claimed                     |

The strict MoE base test disabled the refiner and anchored the runtime configuration. It used the same prompt, negative prompt, seed, initial latent, sampling configuration, precision policy, and physical H200 GPU for both implementations.

The complete MoE/refiner workflows intentionally differ at the handoff:

```text
Original:  base decode -> save MP4 -> reload MP4 -> resize and VAE-encode -> refine
FastVideo: base decode -> keep tensor in memory -> resize and VAE-encode -> refine
```

The original MP4 round trip is lossy. FastVideo avoids that loss, so exact final-video parity is not claimed for the refiner workflow. The refiner transformer itself has exact component-level parity, and the complete FastVideo refiner pipeline passes its production smoke test.

## Validation

- Targeted API and CPU regression suite: `127 passed, 3 skipped`.
  - The skipped cases require allocated GPUs and were covered by separate GPU runs.
- Mixed-precision FSDP inference: passed on H200, including nested and root ignored parameters.
- Dense and MoE transformer tests: exact.
- MoE/refiner eight-GPU production smoke test: passed.
- Strict three-prompt MoE base test: passed.

The local test catalog and reproduction instructions are documented in:

`tests/local_tests/lingbot_video/README.md`

## Video comparison

The following pair is from the strict MoE base-only test. Both files have the same SHA-256 checksum.

### Original LingBot-Video

https://github.com/user-attachments/assets/32474a15-96d6-4494-9c86-781a9f2002dd

### FastVideo

https://github.com/user-attachments/assets/16ae9f3c-e03d-4f01-b422-de00f16093f1

[Generated video — FastVideo](PASTE_FASTVIDEO_VIDEO_URL)


SolitaryThinker (changes requested):
Requesting changes at c7d552ee.

The converted state surfaces are internally consistent (Dense DiT 377/377, MoE base/refiner 977/977, encoder 290/290), and the same-head fastcheck passed. The following blockers remain:

1. Preserve the MoE router bias as a buffer during strict loading. `LingBotVideoRouter.e_score_correction_bias` is registered as a persistent buffer at `fastvideo/models/dits/lingbot_video.py:267`, but `load_model_from_full_model_state_dict` wraps every loaded value in `nn.Parameter` at `fastvideo/models/loader/fsdp_load.py:444` and assigns it at line 488. A reproduction shows the buffer becomes a trainable `Parameter` after FSDP wrapping, so it is no longer in `named_buffers()` and is unmanaged by FSDP. Load buffers as tensors (or preserve the target registration explicitly), and add a strict production-loader test asserting this key remains a buffer.

2. Make the mixed-dtype training rejection use the dtype policy that sharding actually uses. The guard at `fsdp_load.py:140-145` compares against `default_dtype`, while ignored parameters are selected against `mp_policy.param_dtype` at lines 273-281. Modular training defaults `default_dtype` to fp32, while `TransformerLoader` passes `param_dtype=torch.bfloat16`; under that normal path the guard sees no mixed dtypes, proceeds, and leaves the fp32-sensitive gradients unsynchronized. A production-path reproduction confirms bf16-default training raises but fp32-default/bf16-param training does not. Base the guard on `param_dtype` or the exact ignored set and cover this combination directly.

3. Honor the canonical typed refiner opt-out. `pipeline.preset_overrides.refine.enabled` is flattened only to `ltx2_refine_enabled` by `fastvideo/api/compat.py:301-307`, while `LingBotVideoPipeline.load_modules` checks only `refine_enabled` at `fastvideo/pipelines/basic/lingbot_video/lingbot_video_pipeline.py:36`. Therefore `enabled: false` still loads the 30B `transformer_2`. Consume the canonical translated value (or preserve the generic field) and add a typed-config integration test.

4. Complete reproducible parity gates for the public paths:
   - The parity/smoke modules hard-code the author's `/mnt/weka/...` paths (for example `tests/local_tests/encoders/test_lingbot_video_dense_text_encoder_parity.py:25`, `tests/local_tests/transformers/test_lingbot_video_moe_dit_parity.py:32`, and `tests/local_tests/pipelines/test_lingbot_video_pipeline_parity.py:17`), and the README tells reviewers to edit source constants. Make these environment- or fixture-driven so the claimed non-skip runs can be reproduced from an unchanged checkout.
   - The encoder gate at `test_lingbot_video_dense_text_encoder_parity.py:228-229` accepts mean drift `5e-3` and elementwise `atol=rtol=5e-2`, despite claiming exact output parity. Tighten it to the observed/declared accuracy; the model-port guidance starts same-wrapper encoders at roughly `1e-3`.
   - The MoE/refiner tests run one frame/one step with `batch_cfg=False` and assert only shape/finiteness, while the public preset defaults to batched CFG and the only official/native pipeline comparison covers Dense sequential CFG. Add meaningful official/native coverage for the advertised batched-CFG MoE path and the base decode -> re-encode -> refiner -> final-output path. Exact final pixels are not required if the lossy official boundary is documented, but smoke-only coverage is not pipeline parity.

5. Resolve the quality gate instead of self-accepting a visible defect. `tests/local_tests/lingbot_video/PORT_STATUS.md:189` records pronounced blurred pillarbox-like side regions, the escape-hatch table is empty, and line 255 declares no gates remain. The PR has no LingBot-Video SSIM/pixel-quality regression, and both comparison links in the PR body are still placeholders. Attach reviewer-inspectable outputs and add the quality regression, or obtain and document an explicit approved deferral/escape hatch.

Please also resolve the existing high-priority review thread at `fastvideo/models/dits/lingbot_video.py:408`: the per-expert `.item()` loop causes thousands of device synchronizations per MoE forward. The other existing synchronization/caching threads should be addressed in the same hot-path pass.

CI attribution: the GitHub pre-commit failure is PR-caused (YAPF changes five files). The red Buildkite encoder lane is a known Hugging Face download stall; the same-head fastcheck encoder lane passed, so no code change is indicated for that lane.

PR #1596, opened 2026-07-13
[bugfix]: allow unrestricted head sizes in SDPA

SolitaryThinker:
## Summary

- remove the FlashAttention-derived head-size allowlist from the PyTorch SDPA backend
- allow valid sizes such as CLIP-L vision head size 80
- add CPU regression coverage for capability resolution and head-80 attention

## Test evidence

- `pytest fastvideo/tests/attention/test_sdpa_head_sizes.py -q` — 3 passed
- `pre-commit run --files fastvideo/attention/backends/sdpa.py fastvideo/tests/attention/test_sdpa_head_sizes.py` — passed

## Risk

Low: PyTorch SDPA is head-size agnostic and internally selects or falls back between its kernels. `None` is the capability API convention for unrestricted sizes.

### DIFF
--- a/fastvideo/attention/backends/sdpa.py
+++ b/fastvideo/attention/backends/sdpa.py
@@ -15,8 +15,14 @@ class SDPABackend(AttentionBackend):
     accept_output_buffer: bool = True
 
     @staticmethod
-    def get_supported_head_sizes() -> list[int]:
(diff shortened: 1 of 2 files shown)


SolitaryThinker:
Rebased onto current main to pick up the full-suite gate self-test required by pre-commit; no content changes (range-diff clean: `74f023ded = 81d0cb07e`).

PR #1601, opened 2026-07-14
[feat] Add Kandinsky5 QAD training pipeline: data preprocessing, QAT finetune, QAT-aware DMD distillation

leffff:
Purpose
Adds the training-side counterpart to the existing Kandinsky-5 T2V/I2V inference pipeline ([#1471](https://github.com/hao-ai-lab/FastVideo/issues/1471)): a full quantization-aware distillation (QAD) recipe for Kandinsky-5.0 Lite T2V at 512x768/121 frames on the new fastvideo/train/ stack — data preprocessing, Stage 1 (Attn-QAT finetune), and Stage 2 (QAT-aware DMD2 distillation).

Fixes #

Changes
Data preprocessing (fastvideo/pipelines/preprocess/preprocess_kandinsky5_overfit.py): encodes raw video + captions into the t2v parquet schema using Kandinsky5's shared HunyuanVideo VAE and dual Qwen/Reason1 + CLIP text encoders (CLIP pooled projection zero-padded and prepended to the Qwen sequence, stored in the existing text_embedding field — no schema change).
Trainer integration (fastvideo/train/models/kandinsky5/kandinsky5.py): wires Kandinsky5Model into the modular fastvideo/train/ trainer (methods x models x callbacks).
Stage 1 — Attn-QAT finetune (examples/train/configs/fine_tuning/kandinsky5/): FASTVIDEO_ATTENTION_BACKEND=ATTN_QAT_TRAIN routes Kandinsky5's dense/local attention through a fake-quantized straight-through-estimator kernel during finetuning, so the DiT learns to absorb quantization error. No weight quantization happens during training — that's applied post-hoc at inference via nvfp4_qat/FP8.
Stage 2 — QAT-aware DMD2 distillation (examples/train/configs/distribution_matching/kandinsky5/, fastvideo/pipelines/stages/kandinsky5.py's Kandinsky5DmdDenoisingStage, fastvideo/pipelines/basic/kandinsky5/kandinsky5_dmd_pipeline.py): distills the Stage-1 checkpoint from many-step teacher sampling to a 4-step student. Teacher/critic are masked back to full-precision dense attention regardless of the student's QAT env var, via component_loader.py's existing _loading_teacher_critic_model gate (no Kandinsky5-specific handling needed).
Quant config updates (fastvideo/layers/quantization/{fp8_config,nvfp4_qat_config}.py) to include Kandinsky5's attention out-projection / FFN layer names (self_attention.out_layer, feed_forward.mlp.fc_in/fc_out), which differ from Wan's naming.
fastvideo/train/entrypoint/dcp_to_diffusers.py: converts a Stage-1 DCP training checkpoint into a diffusers-format directory, needed to seed Stage 2's student/teacher/critic init_from.
Tests: test_load_kandinsky5.py (load + forward smoke test), test_kandinsky5_qat_attention_engages.py (confirms ATTN_QAT_TRAIN actually engages and backprops, not a silent SDPA fallback), test_e2e_kandinsky5_dmd_t2v_overfit.py (nightly e2e: preprocessing -> Stage 1 -> Stage 2 on a synthetic single sample, asserting checkpoints + validation video are produced).
examples/train/configs/fine_tuning/kandinsky5/README.md: full data -> Stage 1 -> Stage 2 -> inference (NVFP4/FP8) walkthrough.
Test Plan
Ran the full cycle on a real ~100-sample overfit dataset (data/kandinsky5_overfit/videos2caption.json + videos/*.mp4) on an 8-GPU remote host:

# 1. Data preprocessing
CUDA_VISIBLE_DEVICES=0 python -m fastvideo.pipelines.preprocess.preprocess_kandinsky5_overfit

# 2. Stage 1 - Attn-QAT finetune
bash examples/train/configs/fine_tuning/kandinsky5/finetune_qat.sh \
    --training.data.data_path data/kandinsky5_overfit_preprocessed

# 3. Convert Stage-1 DCP checkpoint -> diffusers format
python -m fastvideo.train.entrypoint.dcp_to_diffusers \
    --checkpoint outputs/kandinsky5_t2v_qat_finetune \
    --output-dir outputs/kandinsky5_t2v_qat_finetune/checkpoint-latest-diffusers

# 4. Stage 2 - QAT-aware DMD2 distillation
STAGE1_CKPT=outputs/kandinsky5_t2v_qat_finetune/checkpoint-latest-diffusers
bash examples/train/configs/distribution_matching/kandinsky5/distill_dmd_qat.sh \
    --training.data.data_path data/kandinsky5_overfit_preprocessed \
    --models.student.init_from "${STAGE1_CKPT}" \
    --models.teacher.init_from "${STAGE1_CKPT}" \
    --models.critic.init_from "${STAGE1_CKPT}"
    
    
BEFORE QAT TRAINING: 
https://github.com/user-attachments/assets/dfdf1e07-35aa-47e5-ad43-50e40e55bf1d

AFTER QAT (Stage 1):
https://github.com/user-attachments/assets/d2ef9028-c990-4a86-bc72-181c6dbe6a17

QAT 4 NFE: 
https://github.com/user-attachments/assets/ae787f27-8b59-4e33-9eb4-bfcb8b01a6cf

QAD 4 NFE (Stage 2)
https://github.com/user-attachments/assets/ae580650-b259-4aef-b812-4e0bf9955dfe


SolitaryThinker (changes requested):
Reviewed exact head `7c7704dabf48475afe03980b8858ea911f544325` against `cae8fa18dce465c4367fead159dd70418489a399`.

Changes requested:

1. **[P1] Make the E2E test consume the dataset it generated.** `fastvideo/tests/nightly/test_e2e_kandinsky5_dmd_t2v_overfit.py:112` passes `<PREPROCESSED_DIR>/combined_parquet_dataset`, but `preprocess_kandinsky5_overfit.py:239-240` writes `data_00000.parquet` directly under `PREPROCESSED_DIR`. Dataset discovery therefore fails before Stage 1. Point the override at the actual output root and exercise that path in the gate.

2. **[P1] Export the Stage 1 DCP checkpoint before loading Stage 2.** The E2E test at lines 147-158, the Stage 2 YAML, and the README treat `checkpoint-N` as a Diffusers model. `CheckpointManager` writes only `checkpoint-N/dcp` plus metadata/RNG state, while `load_module_from_path` requires a root `model_index.json` and component directories. Add the `fastvideo.train.entrypoint.dcp_to_diffusers` step to the launcher/docs/test, pass its output to all three roles, and strict-reload the export.

3. **[P1] Preserve FastVideo forward context in the DMD sampler.** `Kandinsky5DmdDenoisingStage.forward` calls the transformer under autocast only at `fastvideo/pipelines/stages/kandinsky5.py:504`. `LocalAttention` requires forward context; Kandinsky catches the missing-context assertion and silently falls back to plain SDPA at `fastvideo/models/dits/kandinsky5.py:390-412`. Stage 2 validation therefore does not exercise `ATTN_QAT_TRAIN` despite claiming to validate QAT. Wrap the call with the same `set_forward_context(...)` contract as the parent stage and add a stage-level backend assertion.

4. **[P1] Make exported DMD checkpoints resolve the DMD pipeline.** The documented `VideoGenerator.from_pretrained` call at `examples/train/configs/fine_tuning/kandinsky5/README.md:97` cannot select `Kandinsky5DMDPipeline`: the exporter copies the base `model_index.json` unchanged (`dcp_to_diffusers.py:120-125`), and model resolution then selects/pins `Kandinsky5T2VPipeline`. The four-step re-noise sampler is currently reachable only through validation's explicit `pipeline_target`. Persist the DMD class/config/sampling defaults in the export/registry path (or require an explicit override and document/test it).

5. **[P1] Do not change existing Kandinsky inference precision semantics.** The shared denoising stage now defaults to bf16 whenever every transformer parameter is fp32 (`fastvideo/pipelines/stages/kandinsky5.py:282-286`), ignoring `pipeline_config.dit_precision`. An explicit fp32 pipeline is consequently cast/run as bf16 (and can hit input/weight dtype mismatches off CUDA), while an fp16 request with fp32-loaded weights is also forced to bf16. Scope the FSDP-training special case to validation or carry the actual mixed-precision policy without regressing normal T2V/I2V inference; add an fp32 precision regression test.

6. **[P2] Use the runtime Qwen token limit during preprocessing.** `preprocess_kandinsky5_overfit.py:147` inherits `max_length=512`, which includes the 129-token Kandinsky system template; after postprocessing removes that prefix, only 383 caption tokens remain. Runtime text encoding uses `129 + 512 = 641`. Set the preprocessing tokenizer max length from `pipeline_config.text_encoder_max_lengths[0]` so training and inference conditioning agree.

7. **[P2] Normalize captions before indexing them.** `preprocess_kandinsky5_overfit.py:161` and line 246 unconditionally use `item["cap"][0]`. The repository's documented preprocessing schema permits `cap` to be a string, so `"Ocean waves..."` is silently trained and validated as `"O"`. Accept/validate a string or a non-empty list once at ingestion.

8. **[P2] Restore the process-global attention backend after the QAT test.** `test_kandinsky5_qat_attention_engages.py:70-71` overwrites `FASTVIDEO_ATTENTION_BACKEND` and clears the global cached selector without teardown. Subsequent model tests in the same pytest process can instantiate under `ATTN_QAT_TRAIN`. Use `monkeypatch` or `try/finally`, restore the environment, and clear the cache again.

Verification: `git diff --check`, `bash -n` on both launchers, YAML parsing, Python byte-compilation, and the repository's exact-diff pre-commit suite all passed. Runtime/GPU pytest was unavailable in this checkout because its Python environments lack PyTorch/pytest. GitHub pre-commit is green; Buildkite build 4412 is still pending. The proposed nightly test currently has the two structural failures above and explicitly lacks a golden/SSIM assertion, so this new DMD/QAT path still needs a non-skipped pipeline/quality gate before merge.

SolitaryThinker (changes requested):
Reviewed exact head `e962207ed912dc6bbfd0673a1d6a64c8b2482860` against the prior reviewed head `7c7704dabf48475afe03980b8858ea911f544325`.

The author update resolves the eight numbered findings structurally, but the following blockers remain:

1. **[P1] Preserve the documented `ATTN_QAT_INFER` fallback.** `fastvideo/pipelines/stages/kandinsky5.py:188-227` treats both QAT backends as exact-match-only. However, `fastvideo/platforms/cuda.py:143-150` intentionally resolves `ATTN_QAT_INFER` to FlashAttention when the sm_120 kernel is unavailable, and the README promises that behavior at lines 146-148. On any non-sm_120 system, module construction succeeds with FlashAttention and the new stage assertion then aborts denoising. Keep only `ATTN_QAT_TRAIN` strict (training must never silently de-quantize), and add an inference-fallback regression test.

2. **[P1] Read the FSDP compute policy instead of parameter storage dtype.** `_resolve_target_dtype()` at `fastvideo/pipelines/stages/kandinsky5.py:229-264` scans parameters and returns the first non-fp32 dtype. FSDP loading hardcodes `MixedPrecisionPolicy(param_dtype=torch.bfloat16)` independently of the checkpoint/storage dtype (`component_loader.py:1087`, `fsdp_load.py:124-127`) and disables automatic input casting. An fp16-loaded FSDP model therefore resolves fp16 even though its forward policy uses bf16; with autocast disabled this can produce fp16/bf16 matmul mismatches. Use the active mixed-precision state (`get_compute_dtype()`) for FSDP and cover that policy in the regression test.

3. **[P1] Exercise the actual Stage-2 deliverable, not only the Stage-1 handoff.** The E2E exports/reloads Stage 1 at `test_e2e_kandinsky5_dmd_t2v_overfit.py:181-184`, then runs Stage 2 in memory and stops after finding a raw DCP checkpoint and an MP4 at lines 186-200. The README likewise jumps from Stage-2 training to an unspecified loadable `path/to/kandinsky5_dmd_checkpoint`. Export the final Stage-2 student DCP, strict-reload it, and instantiate it with the documented DMD pipeline/config override. Otherwise the public inference artifact produced by this recipe is never validated.

4. **[P1] Add and run the required numerical/media-quality oracle.** The E2E docstring explicitly defers a golden comparison (`test_e2e_kandinsky5_dmd_t2v_overfit.py:15-20`) and the assertions only check artifact existence. `fastvideo/tests/nightly/` is also explicitly allowlisted as uncollected in `fastvideo/tests/contract/test_ci_test_collection.py:37`; no Buildkite/Modal command runs this test. A shape-compatible wrong remap, wrong re-noise schedule, NaN training run, or silent SDPA path can still pass. Generate deterministically from the final exported/reloaded student and compare with a reviewed SSIM or latent reference, then record a non-skipped run.

5. **[P1] Prevent stale artifacts from making the E2E false-green.** `_synthesize_single_sample()` deletes only the raw/preprocessed inputs at lines 74-77, while the Stage-1 config has `resume_from_checkpoint: latest`. A rerun can no-op from an old `checkpoint-3`, export that stale checkpoint, and satisfy the final Stage-2 checkpoint/MP4 assertions with files from an earlier run. Remove all three output roots (or use unique temporary paths) and explicitly disable resume for this test before launching either stage.

6. **[P2] Fail clearly on unreadable input video.** `preprocess_kandinsky5_overfit.py:90-93` repeats `frames[-1]` without first checking that OpenCV decoded any frame, so a missing/corrupt file raises an opaque `IndexError`; the E2E fixture also never checks `VideoWriter.isOpened()`. Raise a path-specific `ValueError` for zero decoded frames, validate the fixture writer, and add the small regression case. Validate a non-empty manifest before indexing `records[0]` as well.

7. **[P2] Put the forward-context regression at the stage boundary.** `test_kandinsky5_qat_attention_engages.py:134` manually installs `set_forward_context` around a raw transformer call, so it still passes if the wrapper at `Kandinsky5DmdDenoisingStage.forward:591` is removed and the production stage silently falls back to SDPA again. Drive the DMD stage with a backend spy (or the real backend) and assert that the selected backend is invoked.

8. **[P2] Make the Stage-2 launcher perform the documented conversion.** `distill_dmd_qat.sh` now contains instructions, but still launches the YAML immediately; the prior finding asked for the conversion step in the launcher as well as docs/tests. Accept a Stage-1 DCP path, run `dcp_to_diffusers --verify`, and pass that export to all three `models.*.init_from` overrides.

Verification on this head: Buildkite fastcheck 4443 and GitHub pre-commit are green. Locally, `git diff --check`, both launcher syntax checks, YAML parsing, Python compilation, and the exact-diff pre-commit suite passed. The pipeline-resolution/plain-dtype tests passed (`6 passed`). The caption test could not collect in the available environment because `pyarrow` is not installed. No GPU/quality run was performed, and the new nightly has no recorded non-skip result.

SolitaryThinker (changes requested):
Reviewed exact head `9817f607ce6871b90108f972823afc8a9e48afe0` against the prior reviewed head `e962207ed912dc6bbfd0673a1d6a64c8b2482860`. The update structurally resolves the eight previous findings, but four release blockers remain:

1. **[P1] Do not delete the documented user dataset when the nightly starts.** `_clean_previous_artifacts()` in `fastvideo/tests/nightly/test_e2e_kandinsky5_dmd_t2v_overfit.py:121-139` unconditionally removes `data/kandinsky5_overfit` and `data/kandinsky5_overfit_preprocessed`. Those are the same paths the recipe tells users to populate with their real videos/captions. Running this manual nightly from a normal checkout therefore irreversibly deletes the user's source and preprocessed dataset before replacing it with one synthetic clip. Give the preprocessor explicit test-only input/output roots (preferably under a temporary directory) and clean only those test-owned roots.

2. **[P1] Make the three-step Stage-2 E2E actually update the student.** The test runs Stage 2 for steps 1–3 (`test_e2e_kandinsky5_dmd_t2v_overfit.py:392-401`) but inherits `method.generator_update_interval: 5` from `dmd2_t2v_480p_qat.yaml:46`. `DMD2Method` updates/backprops/steps the student only when `iteration % interval == 0`, so this run performs zero generator updates; only the critic changes. The exported “Stage-2 student” and golden therefore exercise Stage-1 weights through the DMD sampler and cannot catch a broken Stage-2 student backward/optimizer path. Override the interval to `1` for this shortened E2E (or run through an update step and assert that it occurred), then regenerate/review the reference from that corrected run.

3. **[P1] Make the media oracle reject collapsed solid/frozen output.** `_assert_video_not_degenerate()` at lines 305-321 checks only a global `stacked.std()`, so differences between RGB channel means count as “variance” even when every pixel and frame is identical. The SSIM helper also truncates both videos to the shorter frame count (`fastvideo/tests/utils.py:119-124`), and this reference is low-contrast enough that the `0.5` threshold does not rescue the check. A confirmed all-121-frame solid clip at the rounded reference mean RGB `[106, 98, 70]` passes both guards: global std `15.4344`, mean MS-SSIM `0.916887` (min `0.870903`). Assert the required frame count/geometry and meaningful spatial and temporal variation, then calibrate the similarity floor against known-bad solid/frozen clips so a collapsed export cannot false-green.

4. **[P1] Capability-gate the advertised `ATTN_QAT_INFER` fallback.** The new rationale/docs say non-sm120 GPUs fall back to FlashAttention, but `is_attn_qat_infer_available()` in `fastvideo/attention/backends/attn_qat_infer.py:53-54` only tests whether the extension imports. CUDA 13 wheels can contain the sm120/sm121 extension on hosts such as H100 or GB200, so import succeeds, `cuda.py` selects the Blackwell-consumer backend, and the first kernel call fails instead of falling back (the C++ path itself rejects unsupported capability). Check the active device capability before selecting this backend and cover the real platform resolver; the new stage-guard test only injects an already-resolved FlashAttention backend and cannot detect this failure.

The branch also currently conflicts with `main` in `.gitignore` and `fastvideo/configs/pipelines/__init__.py`, so it needs a union-preserving rebase before the next exact-head pass.

Verification: GitHub pre-commit and Buildkite fastcheck 4547 are green. Locally, `git diff --check`, both launcher syntax checks, YAML parsing, the raw-DCP/already-exported launcher flows, and reference metadata/visual inspection passed. The committed 121-frame reference plus its bootstrap commit is credible evidence of one non-skipped bootstrap run, but the focused Python tests could not run in the available local interpreters because PyTorch/pytest are not installed together.

SolitaryThinker (approved):

PR #1602, opened 2026-07-14
[perf]: enable per-block torch.compile for LTX2 with persistent cache

Someone45:
Cache LTX2 graphs and allow all 48 blocks to share one compiled graph. Also added compile to the dreamverse backend launch.

## Purpose

Dreamverse LTX2 startup compilation cost is very high (10+ minutes) and we are able to cache it, reducing subsequent start up below 2 minutes. 

LTX2 was the only major DiT still compiled as a single whole-model fullgraph, because LTX2Transformer3DModel never wired the _compile_conditions declared on its arch config. That produced enormous, unportable graphs (~8k guards) and long cold compiles. Even with per-block (regional) compilation turned on, the cold warmup died deterministically and the Inductor/AOTAutograd artifacts could not be reused across processes, so every launch paid the full autotune/codegen cost. This PR fixes both: makes the 48 transformer blocks share a single compiled graph, and makes those graphs disk-cache-serializable so a second launch on the same box reloads them.

## Changes
- LTX2Transformer3DModel now picks up _compile_conditions from its arch config, so the 48 transformer blocks get compiled individually instead of the whole model as one giant fullgraph. LTX2 was the last major DiT still doing the latter.
- We made the block forward free of per-instance state so all 48 blocks share one compiled graph, as it was otherwise hitting the Dynamo limit at block 17. Moved that check up into the eager _process_transformer_blocks loop
- Made the graphs safe to cache. AOTAutograd's serializer was being tripped and forcing a fresh compile from the lru_cache on the rotary freq-grid helpers, the functools.cache on _sm90_or_newer, and a dtype mismatch in the RMSNorm fallback
- Turned on the caching in the launch script with the paths to cache to

## Test Plan

Validated on a B200 (sm100), torch 2.12.0+cu130. A script runs the dreamverse worker's initialize() + warmup(), the streams continuation chunks. Two configs: 576×320 and 1920×1088 at 121 frame chunks (NVFP4, FlashAttention 4), both with max-autotune-no-cudagraphs and the persistent cache.

export TORCHINDUCTOR_CACHE_DIR=$PWD/.cache/inductor TRITON_CACHE_DIR=$PWD/.cache/triton
export TORCHINDUCTOR_FX_GRAPH_CACHE=1 TORCHINDUCTOR_AUTOGRAD_CACHE=1

Cold vs. reload warmup: ran the dreamverse worker's initialize() + warmup()  to build the cache from cold, then again in a fresh process pointed at the same cache dirs to verify it loads it from disk.
Warm state: after a single warmup, we streamed chunks and timed each 121 frame 1080p chunk.

## Test Results

Cold vs. reload warmup (persistent Inductor/Triton + FX-graph + AOTAutograd cache):
576x320  SDPA  autotune:  cold 557.5s  -> reload 58.3s  (9.6x)
1080p    FA4   autotune:  cold 716.0s  -> reload 76.3s  (9.4x)
(0 recompile-limit hits, 0 autograd_cache_bypass, 0 failures)

Warm steady-state (1080p, FA4, fully warm continuation chunks):
121-frame 1080p chunk generation ~5.3s median (range 5.24-5.45s), e2e 5.26-5.47s
-> within the 5-7s/chunk target; 0 recompiles on continuation chunks

Residual reload ~76s = Dynamo tracing (uncached on torch 2.12) + model load + first inference

### DIFF
--- a/apps/dreamverse/scripts/launch/launch_backend_dreamverse.sh
+++ b/apps/dreamverse/scripts/launch/launch_backend_dreamverse.sh
@@ -39,15 +39,16 @@ export FASTVIDEO_GENERATION_SEGMENT_CAP="${FASTVIDEO_GENERATION_SEGMENT_CAP:-6}"
 export FASTVIDEO_PROMPT_AUTO_SLEEP_MS="${FASTVIDEO_PROMPT_AUTO_SLEEP_MS:-120}"
 export FASTVIDEO_PROMPT_AUTO_TIMEOUT_MS="${FASTVIDEO_PROMPT_AUTO_TIMEOUT_MS:-1800}"
 
-# Persistent torch.compile cache so warmup pays the full autotune/codegen cost only on the first launch.
-# Later launches on the same box reload the Inductor FX-graph + AOTAutograd + Triton artifacts from disk.
-export DREAMVERSE_TORCH_COMPILE_CACHE_ROOT="${DREAMVERSE_TORCH_COMPILE_CACHE_ROOT:-${HOME}/.cache/dreamverse/torch_compile}"
-export TORCHINDUCTOR_CACHE_DIR="${TORCHINDUCTOR_CACHE_DIR:-${DREAMVERSE_TORCH_COMPILE_CACHE_ROOT}/inductor}"
-export TRITON_CACHE_DIR="${TRITON_CACHE_DIR:-${DREAMVERSE_TORCH_COMPILE_CACHE_ROOT}/triton}"
-export TORCHINDUCTOR_FX_GRAPH_CACHE="${TORCHINDUCTOR_FX_GRAPH_CACHE:-1}"
-export TORCHINDUCTOR_AUTOGRAD_CACHE="${TORCHINDUCTOR_AUTOGRAD_CACHE:-1}"
-mkdir -p "${TORCHINDUCTOR_CACHE_DIR}" "${TRITON_CACHE_DIR}"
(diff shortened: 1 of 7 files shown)


SolitaryThinker:
Rebased onto main to pick up the pre-commit CI fix (#1625 — fork-PR head checkout opt-in); no content changes (range-diff clean, both commits =).

SolitaryThinker:
STG output-parity evidence (requested in review): ran `pytest fastvideo/tests/ssim/test_ltx2_similarity.py -vs` on a GB200 node (2×GPU, SP=2, FLASH_ATTN) — this test exercises the refactored STG path since its reference overrides pin `ltx2_stg_scale_video/audio: 1.0`.

**A/B on identical hardware/env:**
- PR head `757e024dc`: `{'slice_cosine_distance': 0.8544774800539017, 'full_cosine_distance': 0.8184729814529419, 'max_abs_diff': 10.375}`
- base `origin/main` (`229419208`, no PR commits): **bit-identical metrics** `{'slice_cosine_distance': 0.8544774800539017, 'full_cosine_distance': 0.8184729814529419, 'max_abs_diff': 10.375}`

Identical metrics to all 16 digits on both runs ⇒ the PR produces the same latents as main — the STG gate move and freq-grid rewrite are output-neutral in eager mode.

Both runs fail the absolute thresholds because this GB200 node diverges from the CI-seeded `B200_reference_videos` references (GB200 substring-matches the B200 device mapping, but the ARM/CUDA-13 cluster env generates different content than the CI B200 pool). That's a pre-existing environment/reference mismatch independent of this PR — filing it separately rather than blocking here.

SolitaryThinker (approved):

PR #1627, opened 2026-07-21
Fix Ulysses sequence-parallel crash in self-forcing causal Wan DiT

pkisfaludi-nv:
### Problem
Running a self-forcing causal Wan model (e.g. `SFWan2.1-T2V-A14B`, `SFWan2.1-T2V-1.3B`) with Ulysses sequence parallelism (`num_gpus>=2`) crashes:

```
RuntimeError: The size of tensor a (4680) must match the size of tensor b (9360)
at non-singleton dimension 1   # in _apply_rotary_emb
```

### Root cause
`causal_wanvideo.py` builds RoPE with `rope_num_frames = post_patch_num_frames * get_sp_world_size()` (sp× too long), but the causal pipeline **replicates** the full token sequence and a full-size KV cache on every rank (it does not shard the sequence). The attention then applies that over-long RoPE to the full local query → shape mismatch (`9360 = 2 × 4680` at sp=2).

### Fix
- Remove `* get_sp_world_size()` from the RoPE length in all branches (relativistic / standard / train) so RoPE matches the full local sequence (no-op at sp=1).
- Parallelize the causal attention **across heads** (Ulysses over heads): each rank runs attention on its head slice `[rank*H/sp : (rank+1)*H/sp]` over the full sequence, then `sequence_model_parallel_all_gather(dim=2)` reconstructs all heads. Applied uniformly to all three attention paths (flex_attention, KV-cache `self.attn`, relativistic window). A sequence all-to-all (à la `wanvideo.py`) is incompatible with the causal KV-cache windowing (expressed in global token coords), so head-parallel is the exact-matching approach; heads are independent → numerically identical to single-GPU.

### Verification
`wlsaidhi/SFWan2.1-T2V-1.3B-Diffusers`, 480×832, 21 frames (7 autoregressive KV-cache blocks), seed 1234:

| | mean | std |
|---|---|---|
| num_gpus=1 | 116.8295684 | 90.8565301 |
| num_gpus=2 (fixed) | 116.8295684 | 90.8565301 |

`torch.equal=True`, `max_abs_diff=0.0` — **byte-identical**; 2-GPU no longer crashes.

Note: verified on the standard KV-cache path (this model uses `rope_cache_policy="absolute"`); the `relativistic` path gets the identical head-slice+all-gather transform for consistency.

### DIFF
--- a/fastvideo/models/dits/causal_wanvideo.py
+++ b/fastvideo/models/dits/causal_wanvideo.py
@@ -20,7 +20,10 @@ import fastvideo.envs as envs
 from fastvideo.attention import (DistributedAttention,
                                  LocalAttention)
 from fastvideo.configs.models.dits import WanVideoConfig
-from fastvideo.distributed.parallel_state import get_sp_world_size
+from fastvideo.distributed.communication_op import (
+    sequence_model_parallel_all_gather)
+from fastvideo.distributed.parallel_state import (get_sp_parallel_rank,
+                                                  get_sp_world_size)
 from fastvideo.forward_context import get_forward_context
 from fastvideo.layers.layernorm import (FP32LayerNorm, LayerNormScaleShift,
                                         RMSNorm, ScaleResidual,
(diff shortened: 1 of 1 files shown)


kevin314 (approved):
Tested on nvl

SolitaryThinker:
Pure-rebased onto current `main` (`a534ba20`) to clear `needs-rebase`; only mechanical YAPF-format conflict resolutions were required in `causal_wanvideo.py`, and the reviewed range-diff preserves the contributor delta.

PR #1598, opened 2026-07-13
[kernel] Build + allow attn_qat_infer FP4 attention on sm_121a (DGX Spark)

Mister-Raggs:
## Summary

The `attn_qat_infer` FP4 attention kernels (modified SageAttention3) were pinned
to **sm_120a**, with the build/docs treating them as "sm_120a-only." On real
hardware they're also **numerically correct on sm_121a (DGX Spark / GB10)** — the
"sm_120a-only" was an untested release-scope assumption (the kernels shipped on
RTX 5090), not an intrinsic limit. This makes them **buildable + opt-in-selectable
on sm_121**, with the correctness evidence below. **Default behavior is unchanged
on every existing GPU** — this is opt-in (`FASTVIDEO_ATTENTION_BACKEND=ATTN_QAT_INFER`).

> **Update (validated on the QAD checkpoint):** on the checkpoint this backend is
> meant for (`FastVideo/FastWan-QAD-1.3B`), FP4 attention is **quality-neutral vs
> bf16**, and the **full 4-bit path (FP4 linear + FP4 attention) runs end-to-end
> on sm_121**. See the [validation comment](#issuecomment-4972811259) for the
> quality read, the determinism-controlled reason SSIM is inapplicable, and the
> A/B/C/D timing. The two notes about "stock weights look worse" and "FP4 linear
> crashes" below are superseded by that comment and annotated inline.

## Evidence — numerical correctness on sm_121 (GB10)

Built for `sm_121a` (`TORCH_CUDA_ARCH_LIST=12.1a`, CUDA 13; `cuobjdump` confirms
`sm_121` SASS), then `sageattn_blackwell` vs bf16 SDPA on a GB10:

| config | cos vs bf16 |
|---|---|
| non-causal, D=128, L=256/512 | 0.981–0.982 |
| non-causal, D=64, L=256/512 | 0.981–0.982 |
| causal, D=128, L=256/512 | 0.985 |
| causal, D=64, L=256/512 | 0.985 |

All ≥ 0.981 (min over 3 seeds each), no NaNs. For reference, an idealized
correct-math NVFP4 attention on the same random input caps ~0.975 — so this is at
the FP4 correctness ceiling. Robust across cross-attention (unequal q/kv), the
full Wan seqlen (32760), and `sm_scale=None`.

> Note: an earlier internal measurement had this kernel wrong on sm_121 (cos
> ~0.52). That was on a pre-#1493/#1506 build; those two FP4-attn correctness
[...]

PR #1631, opened 2026-07-21
[docs] DGX Spark (GB10) performance & tuning guide + reproduction examples

Mister-Raggs:
## Summary

The Spark install guide (`installation/spark.md`) gets you running on a GB10, but
stops there. This adds the layer above it: **which models are practical on the
GB10, what actually makes them faster, and what gives little or nothing on this
unified-memory hardware (and why)** — so users don't spend a night tuning knobs
that can't move here.

New page `docs/getting_started/installation/spark_performance.md`:
- Use distilled few-step models (~40 s/video vs 12–47 min full-step); few-step is
  **VAE-decode-bound** on the ~270 GB/s unified LPDDR5X.
- A **"what helps vs what doesn't" matrix**: bf16 VAE decode ✅ (already the Wan
  default), distilled models ✅, VSA ✅ — vs building FlashAttention ❌ (SDPA is
  already flash-efficient on sm_121), `torch.compile` of the VAE ❌, and
  linear-only quantization ❌ on long-sequence models (attention dominates the
  denoise; linears are a few %). FP4 attention is covered as an opt-in path.
- A **memory** section (unified 128 GB pool, `nvidia-smi` N/A, `peak_memory_mb`
  as the real per-run figure) and a **gotchas** section.

New examples under `examples/inference/optimizations/`:
- `spark_benchmark.py` — reproduces the headline claims (few-step median timing +
  fp32-vs-bf16 decode A/B + peak memory), in-process to control for the GB10's
  few-step run-to-run variance.
- `qad_fp4_ab.py` — FP4 attention quality/speed A/B on the QAD checkpoint
  (`FastVideo/FastWan-QAD-1.3B`), one arm per process.

Plus cross-links from `installation.md`, `spark.md`, and a note at the top of
`inference/optimizations.md`.

## Test evidence — all numbers measured on a DGX Spark (GB10)

Environment: GB10 (cc 12.1), torch 2.12.0+cu130, transformers 5.14.0, CUDA 13.

| Claim | Measured |
|---|---|
| FastWan2.1-1.3B, 3-step | ~40 s/video, decode the largest stage |
| distilled vs full-step (same arch) | ~18× |
| bf16 VAE decode vs fp32 | 1.14× faster, MS-SSIM 0.9999 (lossless), on the identical latent |
| FP4 attention (opt-in) | −6% denoise (40.87 s → 38.45 s), visually equivalent to bf16 |
| Wan2.1-1.3B few-step peak memory | ~8.4 GB (of the 128 GB pool) |

Every number in the guide is one of these measured values (no estimates carried
over). Sample stills from the FP4-attention A/B (bf16 vs FP4) are visually
equivalent; I'll attach them to the PR.

## Depends on (merge order)

This PR is written assuming two other Spark PRs are on `main`, and should merge
**after** both:

- **#1598** (build + allow `ATTN_QAT_INFER` FP4 attention on sm_121a) — `qad_fp4_ab.py`
  and the guide's FP4-attention section rely on the sm_121 kernel + allowlist it
  adds. #1598 also documents the sm_121 backend in `optimizations.md`, so this PR
  deliberately does **not** touch that section (only adds a cross-link).
- **#1607** (reason1 `BatchEncoding` fix) — the guide references running Cosmos-2.5;
  without this, Cosmos-2.5 inference crashes at text-encode on recent transformers.

Both PRs also touch `installation/spark.md` / `inference/optimizations.md`, in
different sections from this one — landing them first avoids a both-touched merge.

## Related (not duplicated)

- The LTX2 FP4-linear reproduction lives in **#1594**; this PR cites it rather than
  duplicating the harness.
- A GB10 record for the performance dashboard is being coordinated separately (a
  single-GPU GB10 benchmark config needs CI gating first).

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -5,6 +5,7 @@ FastVideo supports the following hardware platforms:
 
 - [NVIDIA CUDA](installation/gpu.md)
 - [NVIDIA DGX Spark / GB10 (ARM64 + CUDA 13)](installation/spark.md)
+  ([performance & tuning](installation/spark_performance.md))
 - [Apple silicon](installation/mps.md)
 
 ## Quick Installation
--- a/docs/getting_started/installation/spark.md
+++ b/docs/getting_started/installation/spark.md
@@ -137,6 +137,13 @@ If you hit other issues, please open an issue on our
 our [Slack community](https://join.slack.com/t/fastvideo/shared_invite/zt-3f4lao1uq-u~Ipx6Lt4J27AlD2y~IdLQ)
 for additional support.
 
+## Next: performance & tuning
+
+Installed and verified? See [DGX Spark: Performance & Tuning](spark_performance.md)
+for which models are practical on the GB10, what makes them faster, and what
+won't help on this hardware (and why) — so you don't spend a night tuning knobs
+that can't move here.
+
 ## Development Environment Setup
 
 If you're planning to contribute to FastVideo please see the
--- a/docs/getting_started/installation/spark_performance.md
+++ b/docs/getting_started/installation/spark_performance.md
@@ -1,0 +1,153 @@
+# DGX Spark (GB10): Performance & Tuning
+
+You have FastVideo [installed on a DGX Spark](spark.md) — this page is what to
+run next. It covers **which models are practical on the GB10, what actually
+makes them faster, and what won't help (and why)**, so you don't burn a night
+tuning knobs that can't move on this hardware.
+
+!!! tip "TL;DR"
+    - **Use distilled few-step models** (e.g. `FastVideo/FastWan2.1-T2V-1.3B-Diffusers`).
+      They run in ~30 s/video. Full-step models are 12–47 min on the GB10.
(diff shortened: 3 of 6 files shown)


SolitaryThinker:
Pushed two maintainer commits to this branch: a rebase onto current main and a small fix commit.

- rebase picks up #1690 — selecting `ATTN_QAT_INFER` without the kernel now hard-fails with a receipt instead of silently running flash, so the A/B can't quietly benchmark bf16 vs bf16 anymore. Note the FP4 arms will error on any box until the sm_121a kernel build (#1598) lands.
- `generate_video` returns a plain dict, so the `getattr(result, ...)` reads always hit the fallback — generation_time was silently wall time and peak_memory_mb always None. Switched to dict access in both scripts.
- relabeled "denoise" → generation time (it's text-encode + denoise + decode), and fixed the #1598 citation (runtime allowlist landed via #1647).

Verified the docs' checkable claims and the scripts compile; GPU numbers in the guide are yours to re-confirm with the fixed timing since generation_time may shift them slightly.

PR #1638, opened 2026-07-23
[feat]: Apple Silicon MLX runtime — INT8 Wan2.1 and Wan2.2 inference

aryan5v:
Adds a native Apple Silicon path to FastVideo: the DiT, DMD sampler, and decoder run on Metal through MLX, with INT8 weights by default. Covers the FastWan-QAD 1.3B and 14B (Wan2.1) and the TI2V 5B (Wan2.2) release checkpoints.

## What's here

- **MLX runtime** (`fastvideo/mlx_runtime/`) — DiT forward with dense `mx.fast.scaled_dot_product_attention`, three-step DMD sampling, affine INT8 via `mx.quantized_matmul`, pre-quantized checkpoint save/load, memory limits, and `mx.compile` with an eager fallback.
- **Two entrypoints.** `mlx_wan_prompt_to_video.py` for Wan2.1 (1.3B, 14B). `mlx_wan22_generate.py` for Wan2.2 5B, which needs per-token timestep conditioning, its own VAE and latent layout, and a different flow shift.
- **Generation modes**, each behind a quality gate: `--fast` (RIFE frame interpolation), `--refine` (two-pass, same DiT), `--fast-spatial`, `--enhance-prompt`, `--decode-backend wan-vae`, and experimental draft attention.
- **Decode.** TAEHV by default, full Wan VAE when fidelity matters.
- **Vendored third-party code** under `fastvideo/third_party/`: TAEHV and the MLX RIFE backend, both MIT, both with LICENSE and provenance.
- Regression tests (`fastvideo/tests/mlx/`) and a macOS + Linux-CPU CI workflow.

## Install and run

```bash
brew install ffmpeg
uv venv --python 3.12 --seed && source .venv/bin/activate
uv pip install -e '.[mlx]'
```

```bash
# Wan2.1 (1.3B / 14B)
python examples/inference/basic/mlx_wan_prompt_to_video.py \
  --model-root <release-dir> --mlx-checkpoint <release-dir> \
  --height 480 --width 832 --num-frames 81 \
  --prompt 'A fox runs through a misty pine forest.'

# Wan2.2 (5B)
python examples/inference/basic/mlx_wan22_generate.py \
  --mlx-checkpoint <release-dir> --text-encoder-root <release-dir> \
  --vae-root <release-dir>/vae \
  --height 704 --width 1280 --num-frames 81 --fps 16 \
  --prompt 'A cinematic landscape with mountains and a river.'
```

## Measured

Mac Studio, M4 Max, 36 GB, three-step DMD, INT8 DiT, TAEHV decode, 81 frames:

| Model | Output | End-to-end | Denoise | Peak |
|---|---|---:|---:|---:|
| 1.3B | 832×480 | 110.1s | 89.8s | 3.87 GiB |
| 5B | 1280×704 | 151.4s | 98.5s | 9.34 GiB |
| 14B | 832×480 | 601.8s | 554.2s | 21.68 GiB |

End-to-end includes a cold umT5 prompt encode (18s at 1.3B/14B, 47s at 5B); repeat prompts hit the cache and start at denoising. `--fast` cuts denoising 2.4–2.9× at every size.

## Validation

`fastvideo/tests/mlx/` — 91 passed on an M4 Max (MLX 0.32.0). CI covers a Metal subset on macos-15 and the same tests on Linux CPU.

## Notes for reviewers

Larger than the original 1.3B-only PR: it now carries the runtime, Wan2.2 support, the generation modes, and tests. Code with no path from either entrypoint (probes, causal/streaming attention, W8A8, MXFP4 QAT, I2V) has been pulled out for a follow-up so this stays reviewable.

See the changelog comment below for the fixes made during review, including an `mx.compile` bug that doubled DiT memory in the two-pass modes.


aryan5v:
## Changelog since review started

**`51ede39e9` — `mx.compile` kept a second copy of the DiT weights alive.**

One traced graph per input signature, and each graph pins its own materialization of the quantized weights. The two-pass modes (`--refine`, `--fast --refine`) denoise at two resolutions, so both graphs stayed resident. M4 Max, 832×480, 81 frames:

| config | peak before | after | total before | after |
|---|---:|---:|---:|---:|
| 14B refine | 34.67 GiB | 21.71 GiB | 652.2s | 618.1s |
| 14B fast+refine | 32.25 GiB | 18.11 GiB | 251.5s | 237.4s |
| 5B refine | 15.75 GiB | 10.80 GiB | — | — |
| 1.3B refine | 5.31 GiB | 3.90 GiB | 109.5s | 100.6s |

Memory held after denoising now equals the single-pass baseline exactly at every size — one copy instead of two. The runs also got faster: the old path sat close enough to the machine's ceiling that memory pressure cancelled out what compilation gained. The DiT now tracks the signature it compiled for and retires the previous graph before retracing.

**`b984e0c12`, `ab09ac51d` — vendored the MLX RIFE backend.**

`--fast` pulled `rife-mlx` from a git URL, so every Mac install fetched code from a single-maintainer repo with no tags and no PyPI package. Git dependencies also fail in environments that only reach a package mirror.

Now vendored under `fastvideo/third_party/rife_mlx` alongside TAEHV: 526 lines covering `build_model` and `interpolate_pair`, with upstream's video I/O and torch conversion left out so it imports without `av`, `PIL`, or torch. Both upstream projects are MIT (xocialize/rife-mlx, an MLX port of hzwer's Practical-RIFE); LICENSE records both. Weights still download from `mlx-community/RIFE-4.25` at runtime. The `mlx` extra installs only MLX now.

**`3a63e07e0` — MLX suite is green (91 passed).**

Four failures, all predating these commits and all environmental:

- *compile parity* — MLX 0.32 reassociates float ops, moving a few ULP (2.4e-07 absolute, 41 of 7680 elements). Bit-identity still holds below 0.32, so assert that where guaranteed and a tight tolerance above. A broken trace or silent eager fallback still fails.
- *wan22 real-weight parity* — the torch reference brings up FastVideo's distributed environment, which needs an `env://` rendezvous, so the test errored before reaching its skip guards. Set `MASTER_ADDR`/`MASTER_PORT` as other torch tests here do. It then passes on mean 6.1e-3 and cosine 0.999948, both matching the recorded values; only the worst single element moved, so the aggregate statistics now carry the assertion.
- *import laziness* — `fastvideo/__init__.py` imports `VideoGenerator` eagerly, so the assertion could never hold. Compares against an `import fastvideo` baseline now, which is what it means to check.

Also widened the `mlx` pin to `<0.33`: every benchmark in the release post was measured on 0.32.0, which the previous range excluded.

**`26e2d6aa1` — deferred non-shipping code.**

About 3,300 lines with no path from either entrypoint: `attn_scaling_probe.py` (no references at all), `accel_probe.py`, `wan22_i2v.py`, `mlx_mxfp4_qat.py`, `w8a8_gemm.py`, the causal/streaming stack, and five bring-up probes in `examples/`. Preserved on a branch and coming as a follow-up PR.

Also dropped `test_mlx_affine_qat_parity.py`, which imports `fastvideo.layers.quantization.mlx_affine_qat`. That module is not on this branch, so the test has never been collectable — worth a look if it was meant to ship.

**Open item:** upstream `rife-mlx` declares MIT in `pyproject.toml` and its README but has no LICENSE file. Vendoring is sound on that basis; asking the author to commit one would close it properly.

aryan5v:
## Two fixes in the generation modes

**`--fast-spatial` produced incoherent video.** It bilinearly upsampled the finished latents and decoded at the target size. A Wan latent cell is a learned code for an 8x8 (Wan2.1) or 16x16 (Wan2.2) pixel block, not a low-pass sample, so blending adjacent codes lands off the decoder's manifold: correct silhouette under a smeared, ringing veil. The 2x latent upsample removes 62% of the latent's high-frequency energy and leaves its magnitude unchanged. The upsample now runs on decoded frames, so the latent stays on the grid it was denoised on. This is the same failure that had `--fast-spatial` rejected for Wan2.2-5B; that path is re-enabled.

**`--refine` discarded its own first pass.** The hand-off is `(1 - sigma) * upsampled + sigma * noise`, with sigma taken from the first stage-2 timestep. That grid defaulted to the stage-1 grid, which opens at `t=1000`, where sigma is exactly 1.0. Stage 1 was multiplied by zero, so refine was a plain full-resolution generation at twice the cost. Leading full-noise steps are now dropped (`1000,757,522` -> `757,522`, sigma 0.757), and an explicit full-noise grid raises instead of silently discarding the pass. Same fix in the 5B path.

Smaller items: post-decode passes share one h264 re-encode, so `--fast --fast-spatial` no longer encodes twice; `gelu_tanh` calls `mlx.nn.gelu_approx`, which is bit-identical and 3.3x faster on the FFN shape; `FASTVIDEO_MLX_WINDOW` now warns, because sliding-window attention is fast and unusable on a dense-trained checkpoint; geometry errors name the calling mode instead of always saying "Refine".

## Validation

M4 Max 36 GB, 1.3B INT8, 832x480x81, seed 0, TAEHV decode, same prompt throughout:

| | denoise | peak |
|---|---:|---:|
| baseline | 86.1s | 3.87 GiB |
| `--fast-spatial` | 10.3s | 2.75 GiB |
| `--fast-spatial --fast` | 4.5s | 2.78 GiB |
| `--refine` before -> after | 95.1s -> 66.5s | 3.90 -> 3.89 GiB |

- `pytest fastvideo/tests/mlx/` — 110 passed.
- `pre-commit run --files <changed>` — passed.
- Refine now follows its stage-1 draft: structural agreement 0.838, against 0.046 for an unrelated run at the same settings.
- Before/after generations inspected for every mode above, plus `--refine` on the pre-fix head.

## Review notes

- Refine peak memory is unchanged, so the `mx.compile` graph-retirement fix in 51ede39e9 still holds.
- `--fast-spatial` output is soft by construction: it carries stage-1's detail budget and no more. `--fast-spatial-sharpen` defaults to 0.4.
- 5B refine has a lower ceiling than 1.3B. Its warped grid gives a best stage-1 weight of 0.060 against 0.243. Documented in the design doc, unresolved.
- No change to sampler math, quantization, checkpoint format, or the default flag set. `--fast-spatial` now decodes at stage-1 resolution, so `--save-latents` writes stage-1 latents in that mode.

SolitaryThinker:
Pushed maintainer repair commit 7cc41126 addressing the review blockers across MLX attention/sampling, transactional checkpoints, generation defaults and geometry, frame planning, artifact integrity, CI coverage, and vendored license packaging. Local verification: 133 passed, 2 skipped; pre-commit checks and direct mypy passed; the built wheel contains both vendored MIT licenses. GitHub CI has been retriggered.

aryan5v:
The Wan2.2-5B entrypoint had the prompt-embedding cache implemented and tested, but `--prompt-embeds-cache` defaulted to `None`, so both `_load_prompt_cache_if_valid` and `_save_prompt_cache` returned immediately and every run paid a full UMT5 encode. The Wan2.1 entrypoint has cached by default all along.

`_default_prompt_cache_path` derives a filename from the fingerprint already being computed, which covers the prompt, the enhanced prompt, the enhancer backend, encoder root, sequence length and dtype. `--prompt-cache` mirrors the Wan2.1 flag and defaults to on; `--prompt-embeds-cache` still overrides it.

Measured on 5B `--fast --fast-spatial`, 1280x704x81, M4 Max:

| | cold | warm |
|---|---:|---:|
| prompt encode | 46.8s | 0.0s |
| denoise | 8.0s | 8.0s |
| end to end | 58.5s | 10.3s |

Denoise is unchanged, so this only removes a fixed cost. Two tests added for the default path and for fingerprint separation. `pytest fastvideo/tests/mlx/` — 136 passed.

SolitaryThinker:
Pushed `428e94f7` to close the prompt-cache review findings: Wan2.1/Wan2.2 now share a best-effort, integrity-checked atomic cache; explicit caches reject stale prompts/models; incomplete model fingerprints fail closed; and the Wan2.2 default/disabled/explicit wiring has a load-bearing regression. Local Mac verification: full MLX suite 141 passed, 2 skipped; PyTorch/SDPA/config smoke set 30 passed; all pre-commit hooks passed.

SolitaryThinker (approved):

PR #1641, opened 2026-07-26
[docs]: cover all registered models in the support matrix

SolitaryThinker:
## Problem

`docs/inference/support_matrix.md` lists ~23 models while `fastvideo/registry.py` on main registers **82** Hugging Face model IDs. Entire families (HunyuanVideo 1.5, Kandinsky 5.0, LTX-2/2.3, FLUX.2, Cosmos-Predict 2.5, LingBot, SFWan, LongCat IDs, Z-Image, GLM-Image, stable-audio, ...) are absent, and the README points here for the full supported-model list. The page also had two broken footnotes: the LongCat row said "See note**" but no `**` note exists (its HF IDs appeared nowhere), and `*` markers on the FastWan2.2-FullAttn row and STA cells reference a footnote that never existed in the page's git history.

## Solution

- New **Registered Model IDs** section: one row per `register_configs(...)` entry, generated from `fastvideo/registry.py` at main commit `8d89f30d` (the state this PR was audited against) — family, all 82 HF IDs, declared `workload_types`, and a link to the matching runnable script in `examples/inference/basic/` where one exists (each linked file verified present on disk).
- Honest scope line on the existing Models x Optimization table: registered models absent from it are **untested** against those optimizations, not incompatible (per-cell validation stays exactly as it was — no cell was upgraded).
- LongCat row: dangling "See note**" replaced with the registered T2V ID.
- Orphaned `*` markers removed (no historical footnote text to restore; the STA column header already carries the Legacy Branch qualifier).
- New **Hardware and OS** section sourced from the three installation guides (x86_64 CUDA 12.6/13.0, DGX Spark aarch64 CUDA 13, Apple silicon macOS 14+), so the README's pointer to this page for hardware assumptions is now true.
- Footnote for stable-audio: the pipelines are text-to-audio (`StableAudioT2AConfig`) though registered under the T2V workload option.

Every new cell is derivable from the repo: IDs/families/workloads from `registry.py` (AST-extracted, not hand-transcribed), example links from files on disk, hardware statements from the installation guides. No optimization compatibility was claimed for any newly listed model.

## Test evidence

- Extraction script asserts: 82 total IDs, every example link exists in `examples/inference/basic/`.
- `pre-commit run --files docs/inference/support_matrix.md`: codespell + PyMarkdown pass.

### DIFF
--- a/docs/inference/support_matrix.md
+++ b/docs/inference/support_matrix.md
@@ -13,6 +13,92 @@ For the canonical, code-level list of model IDs recognized by
     We do this because we believe VSA is strictly better than STA for the
     actively maintained `main` inference path.
 
+## Registered Model IDs
+
+Every Hugging Face model ID registered in `fastvideo/registry.py` on `main`
+(commit `8d89f30d`), grouped by family. Any ID below can
+be passed to `VideoGenerator.from_pretrained(...)`; FastVideo resolves the
+matching pipeline and sampling defaults. The **Family** column is a
+documentation grouping: it follows each registration's declared `model_family`,
+except `black-forest-labs/FLUX.1-dev`, which declares none and is listed under
+`flux` for readability. The **Workloads** column shows each
+registration's declared `workload_types`; `—` means the entry is registered
+without a UI workload option but is still loadable by ID. The **Example**
+column links a runnable script in `examples/inference/basic/` where one exists.
+
+| Family | HuggingFace Model ID | Workloads | Example |
+|--------|----------------------|-----------|---------|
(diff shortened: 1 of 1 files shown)


SolitaryThinker:
Added to this PR (it owns the support-matrix surface): an explicit note that Wan-VACE is not currently supported on main — no VACE pipeline or registered ID exists (verified: zero matches in `fastvideo/`) — with the registered Wan2.1-Fun control pipeline as the closest supported path. Addresses the docs half of #1435.

PR #1657, opened 2026-07-28
[refactor]: resolve attention backend once per component at load time

SolitaryThinker:
## Purpose

Resolve each component's attention backend **once, at load time**, and carry the
decision on the component — instead of every attention layer re-deriving it from
process-wide state while the model is being built.

Today a layer's backend is decided by reading `FASTVIDEO_ATTENTION_BACKEND` at
the moment the layer is constructed, and by a process-global override that has
to flush a cache to take effect. That makes kernel selection an ambient property
of the process rather than a declared property of the component: two components
in one process cannot request different backends without mutating and restoring
global state around each construction, which is what the DMD distillation path
does today (it pops the env var, clears the selector cache, builds the dense
teacher/critic, then puts it all back).

## Design

**Resolve once. Carry the decision on the component.**

1. **One parse of the environment.** `FastVideoArgs.attention_backend` is the
   typed request for a run. When it is unset, `__post_init__` folds
   `FASTVIDEO_ATTENTION_BACKEND` into it — so the environment is an *input*,
   read once, rather than something library code consults later. An explicit
   request always wins and fails fast on a typo; the env var keeps its
   historically permissive parse, where an unknown name falls through to
   automatic selection.

2. **The loader resolves per component.** `PipelineComponentLoader.load_module`
   applies that request to each component it loads. A role-level request already
   resolved upstream (the training stack's per-role `attention_backend`) takes
   precedence over the run-wide one.

3. **The decision is recorded on the component's own config.** Each loader
   writes what its component resolved to
   `ModelConfig._resolved_attention_backend`, on the config object the model
   keeps as `self.config`. After load it is readable from the component:

   ```python
   transformer.config._resolved_attention_backend
   ```

   The loader is the right place to record it rather than the caller above:
   a loader may narrow the request for one component — the DMD teacher/critic
   transformers build dense — and the recorded value is what that component
   actually resolved, not what the run asked for globally.

4. **Every selection input is in the cache key.** The resolution cache is keyed
   on the request, the global force, the env value, component identity and the
   active device index, so a changed request lands on a different cache entry.
   No `cache_clear()` is needed anywhere, and none should be added. The device
   index matters because platform resolution probes the *current* device's
   capability — `ATTN_QAT_INFER` picks sm_12x CUTLASS, sm_100/sm_103 FP4, or a
   FlashAttention fallback per arch — so a resolution taken for one device is
   never handed to another.

This lets the distillation path drop its env-mutate / cache-flush dance: the
teacher and critic declare a dense request for their own construction, and the
generator keeps the quantized one.

### What is not in this PR

Layer constructors nested inside a model receive only a bare
`supported_attention_backends` tuple, threaded through block constructors up to
four levels deep — 43 construction sites across 26 model families. They cannot
yet read the decision off their own config, so the request is delivered from
loader to layers by a private ContextVar,
`_component_attention_backend_scope`. It is plumbing with a stated end, not API:
it is private, absent from `fastvideo/attention/__init__.py`, and
`fastvideo/attention/AGENTS.md` documents the deletion path — thread the request
alongside that tuple one family at a time, starting with the families that carry
per-role requests (Wan, LTX-2, Kandinsky5), and the scope goes away when the
last family lands. Doing all 26 families in one PR would be a large diff whose
failure mode is silent: a missed site drops the per-role request and the
teacher/critic builds quantized, which shows up as numerics rather than an
error.

`global_force_attn_backend_context_manager` is untouched. Migrating the training
stack off it is separate work.

## Blast radius

Shared-forward: the selector is read by both the inference and training stacks.
Behavior is intended to be identical for every component / architecture /
request combination — that is the contract these tests exist to hold. Layers
built outside a loader (the denoising stages, and direct model construction in
tests) still read the env var through the selector, unchanged.

## Test Plan

`fastvideo/tests/api/test_attention_selector_resolution.py` is a golden
precedence matrix: an independent oracle spells out the documented precedence,
and the test asserts the selector agrees across the full cross-product of
(global force x request x env x layer default), plus the fallback when a
requested backend is not in a layer's supported set. Any precedence change fails
it. Alongside it:

- env is folded into the typed request exactly once; an explicit request wins;
  an unparseable env value falls through instead of raising, while an
  unparseable explicit request raises;
- the recorded decision is readable from the component after load, and a
  narrowed request (teacher/critic) records the narrowed value;
- component identity and the active device index are part of the cache key;
- scopes are exception-safe and nest.

CPU-only — the platform is faked, no kernels and no GPU:

```bash
pytest fastvideo/tests/api/test_attention_selector_resolution.py \
       fastvideo/tests/attention/test_selector_role_override.py \
       fastvideo/tests/train/utils/test_moduleloader_attention_backend.py -q
```

These run in the `microscope-unit-tests` lane.

## Checklist

- [x] I updated documentation if needed (`fastvideo/attention/AGENTS.md`)
- [x] I considered GPU memory impact of my changes (none — selection only)

### DIFF
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -76,6 +76,7 @@ surfaces:
       lora_target_modules: "Legacy LoRA configuration surface pending dedicated component API."
       output_type: "Legacy output formatting surface pending GenerationResult cleanup."
       VSA_sparsity: "Model-specific inference optimization not yet represented in the typed public schema."
+      attention_backend: "Process-wide default attention-backend request applied per component at load time; kernel-selection knob not yet represented in the typed public schema."
       moba_config_path: "Model-specific MoBA optimization surface not yet represented in the typed public schema."
       master_port: "Executor/bootstrap compatibility field; not part of the canonical inference schema."
       refine_transformer_path: "Generic stage-2 refine transformer override; no typed equivalent yet."
--- a/fastvideo/attention/AGENTS.md
+++ b/fastvideo/attention/AGENTS.md
@@ -10,7 +10,7 @@ Backend registry + selector wrapping FlashAttn / SageAttn / SageAttn3 / SDPA / V
 attention/
 ├── __init__.py            # Exports DistributedAttention, LocalAttention, get_attn_backend
 ├── layer.py               # DistributedAttention, DistributedAttention_VSA, LocalAttention
-├── selector.py            # get_attn_backend (cached) + env-var override
+├── selector.py            # get_attn_backend (cached) + _component_attention_backend_scope
 ├── backends/
 │   ├── abstract.py        #   AttentionBackend / AttentionMetadata / AttentionMetadataBuilder
 │   ├── flash_attn.py      #   FA2/FA3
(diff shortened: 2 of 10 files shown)


SolitaryThinker:
Lane state at `5b9a6112`: every lane green except
`pr-fastcheck/microscope-transformer-tests`, which failed with **exit status
124** — a timeout, not a test failure.

Attribution, on differential evidence:

- The **same** transformer lane on the `ci` pipeline passed at this exact
  commit, in 10m44s. The two pipelines run the same tests; a real
  backend-selection regression would fail on both, deterministically.
- Exit 124 means the lane was killed on its wall clock, so no assertion was
  reported. That rules out the failure mode this change could plausibly cause
  (a mis-resolved backend surfaces as an assertion or a device-side assert
  under exit 1, not as a timeout).
- Every other lane passes on both pipelines, and the `fastcheck-passed`
  aggregate is green.

Which timeout it is — image-rebuild storm, model-download stall, or GPU
capacity wait — I can't say: the `pr-fastcheck` pipeline is private, so the
log isn't readable from here. Calling this infra rather than PR-caused rests
on the same-commit cross-pipeline pass, not on a log.

PR #1659, opened 2026-07-28
[bugfix]: harden FastVideo Studio UI reliability and accessibility

RazorCrest00:
## Problem

FastVideo Studio has some UI reliability/accessibility problems:

- the application shell and detail sidebars were not usable on narrow viewports;
- the "Create Job" menu depended on hover
- gallery and dataset media failures rendered as empty content.
- API failures and initial loads could look like real empty states
- create-job, GPU refresh, and caption-save failures were not actionable
- job and dataset cards nested actions inside button-like cards
- routes emitted duplicate "main" landmarks
- sliders lacked names on the actual controls
- shared controls missed 44 px touch targets and sufficiently visible focus indicators.

## Solution

- Add a responsive navigation and full-width mobile job/dataset detail drawers
- Cover 320, 375, 414, and 768 px layouts without horizontal overflow
- Replace the hover-only Create Job trigger with an accessible radix menu
- Add media controls and explicit media-error fallbacks
- Separate loading, empty, error, retry, and stale-data states
- Surface create-job and caption-save errors with recovery paths
- Restructure cards so selection and action buttons are semantic siblings
- Enforce one `main` landmark, accessible slider names, 44 px targets, and high-contrast focus rings

## Test evidence

- `npm test` 16 files, 75 tests passed.
- `npm run typecheck` passed.
- `npm run e2e -- --reporter=line` 10/10 Chromium scenarios passed.
- `npm run build` optimized Next.js production build passed; 12 static pages generated.
- Rendered inspection desktop shell, Create Job menu, settings sliders, 375 px navigation, and full-width dataset drawer verified.
- Rendered semantics: one `main`, no horizontal overflow, 11 named 44×44 sliders, and mobile drawer `aria-modal="true"` verified.
- Fresh browser console: no warnings or errors.

### DIFF
--- a/apps/fastvideo_studio/e2e/create-job.spec.ts
+++ b/apps/fastvideo_studio/e2e/create-job.spec.ts
@@ -13,10 +13,9 @@ test.describe('create inference job', () => {
   test('creates a T2V job and shows it in the queue', async ({ page }) => {
     await page.goto('/inference');
 
-    // The "Create Job" button reveals a workload menu on hover; wait for the
-    // T2V item to become visible before clicking so the CSS hover transition
-    // can't race the click.
-    await page.getByRole('button', { name: /create job/i }).hover();
+    // The trigger opens a real menu on click, so this path works for touch,
+    // mouse, and keyboard users.
+    await page.getByRole('button', { name: /create job/i }).click();
     const t2vItem = page.getByRole('menuitem', { name: /T2V/i });
     await expect(t2vItem).toBeVisible();
     await t2vItem.click();
--- a/apps/fastvideo_studio/e2e/gallery.spec.ts
+++ b/apps/fastvideo_studio/e2e/gallery.spec.ts
@@ -4,7 +4,7 @@ import { API_BASE, skipWithoutMock } from './helpers';
 
 /**
  * Gallery page: the seeded completed inference job surfaces as a media tile
- * (an <article> wrapping a <video>) captioned with its prompt.
+ * with playback controls or an explicit media-error fallback.
  */
 test.describe('gallery', () => {
   skipWithoutMock();
@@ -30,12 +30,15 @@ test.describe('gallery', () => {
       page.getByRole('heading', { level: 1, name: 'Gallery' }),
     ).toBeVisible();
 
-    // The completed job renders as an <article> containing a <video> tile.
-    const tile = page
-      .locator('article')
-      .filter({ has: page.locator('video') });
-    await expect(tile.first()).toBeVisible();
+    const tile = page.locator('article').filter({ hasText: completed!.prompt });
+    await expect(tile).toBeVisible();
+    await expect(
+      tile.locator('video').or(tile.getByText('Preview unavailable')),
+    ).toBeVisible();
 
-    await expect(page.getByText(completed!.prompt)).toBeVisible();
+    const video = tile.locator('video');
+    if (await video.isVisible()) {
+      await expect(video).toHaveAttribute('controls', '');
+    }
   });
 });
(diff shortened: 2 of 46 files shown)


SolitaryThinker:
Thanks @RazorCrest00, and sorry for the slow first response on this. Good pass, and the sweeps hold up: one `main` landmark app-wide, no `100vh` left, no leftover low-contrast rings, nothing under 44px in `src/`. The slider naming fix is right, and I confirmed the names actually come through in a real browser's accessibility tree, not just jsdom.

I pushed four commits to your branch rather than sending you back a list, since three of them were already written. Revert or rework anything you disagree with.

**`radix-ui` wasn't declared.** `CreateJobButton.tsx:5` imported the monolithic package, but `package.json` only had the scoped `@radix-ui/react-*`. It was resolving from your local `node_modules`. On a clean `npm ci` this broke typecheck and build, and `CreateJobButton.test.tsx` failed to resolve, so 73 tests ran instead of 75 and the two covering this PR's main fix were silently skipped. E2e ran 0 of 10 because the dev server never came up. Now declares `@radix-ui/react-dropdown-menu` and imports the scoped package.

**`aria-modal` without modality.** `JobDetailsSidebar.tsx` and `DatasetSidebar.tsx` set `aria-modal="true"` but nothing was inert. At 375px, six controls outside the open drawer stayed focusable and tab walked out of it, so screen readers were told those controls were hidden while keyboard users could still reach them. Added a `useDrawerFocus` hook that sets `inert` on the background and moves focus in and out.

**`Badge` inside a `<button>`.** `badge.tsx` rendered a `<div>`, and the card restructure put it inside the new select button. It's already `inline-flex`, so it renders as a `<span>` now with no visual change.

**Gallery error state.** It was a plain paragraph while the job queue, datasets and GPU surfaces all got `role="alert"` and a retry. Now matches them.

Verified on a clean install: typecheck passes, 75/75 unit tests, production build succeeds, e2e 10/10.

Four things I left alone:

- `DownloadCaptions.tsx:79` still uses the hover idiom you replaced elsewhere. I checked it on a touch viewport and it does open, so it isn't broken the way the old Create Job menu was, but there's no `aria-expanded` or `aria-controls`, no open state, and Escape closes the whole drawer instead of the menu. Converting it to the same `DropdownMenu` would fix all three. Worth a follow-up, not this PR.
- `UploadZone.tsx:100` has the `role="button"` container with nested interactive children that you removed from the cards. Same class of issue, but it predates this PR and the drag-and-drop path needs care.
- `button.tsx:24` now gives `icon` and `icon-sm` the same `size-11`. Collapsing them is your call on the component API, not a defect.
- `globals.css:167` uses `!important` inside `:where()`, where rule ordering would do the same. I left it because `globals.test.ts` string-matches that exact line and I didn't want to touch focus styling without visual review.

Separately, and not on you: no CI lane typechecks, builds or tests `apps/fastvideo_studio`, which is why an undeclared dependency reached review with green checks. `python-multipart` is also undeclared in `pyproject.toml` while `server.py` and `mock_server.py` both define upload routes, so both crash at import on a clean env. I'll open a PR for those.

RazorCrest00:
Thanks for the thorough review and for pushing these fixes directly. I pulled and reviewed all four commits and agree with the scoped Radix dependency, drawer focus management, Badge semantics, and Gallery error-state changes. I also reran verification after a clean `npm ci`: typecheck passed, all 75 unit tests passed, the production build generated all 12 static pages, and all 10 Chromium E2E tests passed.\n\nThe follow-up scope makes sense. I’ll keep the Download Captions menu and UploadZone interaction issues separate, and leave the icon sizing and focus-style cleanup for intentional component and visual review. Thanks as well for handling the missing Studio CI coverage and `python-multipart` dependency separately.

PR #1660, opened 2026-07-29
[perf]: reduce per-step overhead in causal Wan denoising, plus optional CUDA graph capture

Ishxn20:
## Summary
 
Ports worthwhile performance optimizations from NVIDIA's FlashDreams (https://github.com/NVIDIA/flashdreams) — a project built specifically for chunk-by-chunk ("causal") video generation — into FastVideo's equivalent pipelines. Of the five candidates identified, four are addressed here:
 
| # | Optimization | Status in this PR |
|---|---|---|
| 1 | Stop syncing to the CPU to read two KV-cache position counters | Done, always on |
| 2 | Record the GPU step sequence once and replay it (CUDA graph capture) | Done, opt-in, off by default |
| 3 | Run KV-cache refresh in the background | Not included — see "Dropped" below |
| 4 | Copy finished frames off the GPU in one batched transfer | Done, always on |
| 5 | Stop reconverting the noise schedule to float64 every step | Done, always on |
 
## Problem
 
FastVideo's chunk-by-chunk Wan pipelines (`WanCausalDMDPipeline`, `WanCausalPipeline`) generate video in blocks, calling the transformer many times per generation. Several places in that hot path were doing more work than necessary on every single call: forcing the CPU to wait on GPU-resident counters, copying finished frames back to host memory one at a time, recomputing values that don't change between steps, and — for item 2 — dispatching hundreds of small GPU operations individually instead of recording them once and replaying the whole sequence.
 
### Item 1 — Stop waiting on the GPU to read cache position counters
 
**Before:** Two counters (`global_end_index`, `local_end_index`) that track how much of the KV cache is filled were stored as GPU tensors. The attention code called `.item()` on them every layer, every step — each call forces the CPU to stop and wait for the GPU to catch up.
 
**After:** These are now plain Python integers, initialized as such instead of as `torch.tensor([0], ...)`. The read/write sites already handled both forms, so only the initialization needed to change.
 
**File:** `fastvideo/pipelines/stages/causal_denoising.py`, `CausalDMDDenosingStage._initialize_kv_cache`.
 
**Deliberately left alone:** `fastvideo/train/models/wan/wan_causal.py` initializes these same counters as GPU tensors and raises an error if they aren't — its gradient-checkpointing snapshot logic depends on them being tensors. Training was not touched.
 
### Item 2 — CUDA graph capture for the denoising step (opt-in, default off)
 
**The idea:** Once the KV-cache window fills up, the physical read/write positions inside the attention code become constant on every subsequent call (proven by direct derivation from the actual code, not just assumed by analogy to FlashDreams — see the extended notes in `FlashDreams.md`). Once positions are constant, the entire sequence of GPU operations for one denoising step is identical every time, which is exactly the condition needed to record it once (a CUDA graph) and replay it instead of dispatching each operation individually.
 
**What was required to make this safe:**
 
- Split the KV-cache eviction/write logic in `CausalWanSelfAttention.forward` into named steps, driven by an explicit `is_chunk_start` flag passed down from the denoising stage's own loop position, instead of re-derived from tensor comparisons inside the attention module (a value derived inside a captured region can't be recomputed on replay).
- Two separate captured graphs — one for the step that evicts old frames, one for the step that only overwrites — since they touch memory differently and replaying one for the other would silently corrupt state.
- A new `CausalCudaGraphWrapper` / `CausalCudaGraphDispatch` (`fastvideo/pipelines/stages/causal_cuda_graph.py`), adapted from FlashDreams' `infra/cuda_graph.py`: runs eagerly for a couple of warmup calls, captures on the next call, replays after that.
- A new setting, `enable_causal_cuda_graph` (default `False`), plus the matching `--enable-causal-cuda-graph` CLI flag (`fastvideo/fastvideo_args.py`).
- **Three unrelated pre-existing bugs had to be fixed for capture to work at all**, since CUDA graph capture forbids copying from ordinary (unpinned) CPU memory:
  - The cached rotary-position-embedding table was being moved to the GPU with a plain `.to(device)` from an unpinned CPU tensor, every call. Fixed by pinning the cached table and using `non_blocking=True` (`fastvideo/layers/rotary_embedding.py`, `fastvideo/models/dits/causal_wanvideo.py`).
  - The timestep-embedding frequency table was being rebuilt and re-transferred from scratch on every single call, never cached. Fixed by caching it, pinned (`fastvideo/layers/visual_embedding.py`).
  - A `torch.tensor([0], device=...)` literal — which actually builds on CPU and copies over, contrary to how it reads — was replaced with `torch.zeros(1, device=..., dtype=torch.long)`, which allocates directly on-device (`fastvideo/models/dits/causal_wanvideo.py`).
- **A real correctness bug found during testing:** the KV-cache counter update lives inside the exact function that gets captured. On replay, that Python-side update never runs — only the recorded GPU kernels do — so the counters would silently freeze after the first replay. Fixed by adding `_finalize_kv_cache_counters`, called from the denoising stage (outside the captured region) after every steady-state call, setting both counters from values the caller already knows.
- **Refuses to engage, with a clear log message, rather than crash or silently produce wrong output**, unless all of:
  - `enable_causal_cuda_graph` is explicitly set to `True`
  - `local_attn_size != -1` (a fixed-size window exists)
  - `rope_cache_policy == "relativistic"` (position embeddings don't depend on the per-chunk `start_frame`)
  - the pipeline is single-transformer, not the dual-transformer MoE path (`boundary_timestep is None`) — the "which model is at chunk-start" bookkeeping for the two-transformer case wasn't worked out and is explicitly out of scope for this PR
  - neither `dit_cpu_offload` nor `dit_layerwise_offload` is enabled — both move weight storage during the forward pass, which breaks the stable memory addresses a captured graph depends on
**Files:** `fastvideo/pipelines/stages/causal_cuda_graph.py` (new), `fastvideo/pipelines/stages/causal_denoising.py`, `fastvideo/models/dits/causal_wanvideo.py`, `fastvideo/fastvideo_args.py`, `fastvideo/layers/rotary_embedding.py`, `fastvideo/layers/visual_embedding.py`.
 
**No shipped model preset currently sets `rope_cache_policy` to `"relativistic"`** — this setting exists in the code but isn't used by any default config, so this path was tested by manually overriding the config, not by running an existing preset.
 
### Item 4 — Batch the GPU→CPU frame copy after decoding
 
**Before:** After a video finishes generating, each frame was copied from GPU to host memory one at a time in a Python loop. Each copy blocks until it completes, so the next frame's work can't be queued up in advance.
 
**After:** All frames are stacked into one tensor and copied in a single transfer.
 
**File:** `fastvideo/entrypoints/video_generator.py`, the frame post-processing block in `VideoGenerator.generate_video`.
 
**Trade-off:** briefly holds all frames in GPU memory at once instead of one at a time. Measured cost for an 81-frame, 480×832 clip: about 0.1GB — small next to what generation already uses, but not measured at higher resolutions or frame counts.
 
### Item 5 — Cache the float64 noise schedule instead of rebuilding it every step
 
**Before:** `pred_noise_to_pred_video` and `pred_noise_to_x_bound` each converted the scheduler's sigma/timestep tables to float64 and re-uploaded them to the GPU on every single denoising step, even though the schedule doesn't change during one generation.
 
**After:** A new `get_float64_schedule` (`fastvideo/models/utils.py`) converts the tables once and memoizes the result on the scheduler object, keyed by object identity so a genuinely replaced schedule (via `set_timesteps`) is still detected and rebuilt correctly.
 
**Test added:** `fastvideo/tests/ops/test_flow_schedule_cache.py` — 10 cases covering cache reuse, invalidation on schedule replacement, independence between separate scheduler instances, correct precision, and exact numerical agreement with the pre-change code across several schedule lengths. All 10 pass.
 
### Item 3 — dropped, not included
 
FlashDreams runs KV-cache refresh in the background so it overlaps with converting frames to video, which happens continuously in their pipeline. FastVideo only converts to video once, after the entire denoising loop finishes (confirmed by reading `wan_causal_dmd_pipeline.py` — the decode stage is added after the full denoising stage completes). At the point this refresh would run, the very next thing that happens is the next chunk, which needs the memory this refresh is updating — there's nothing to overlap it with. Implementing it would add complexity for no benefit unless FastVideo's pipeline structure changes to decode per-chunk instead of at the end.
 
## Models affected
 
- **Items 1 and 2** reach `WanCausalDMDPipeline` and `WanCausalPipeline` only (`fastvideo/pipelines/basic/wan/`).
- **Item 5** reaches any pipeline using the shared flow-matching conversion helpers in `fastvideo/models/utils.py`.
- **Item 4** reaches all generation that produces pixel-space video frames.
- **`matrixgame2`, `dreamx_world`, `lingbotworld2`** keep their own, separate copies of the attention and KV-cache code and are **not** affected by items 1 or 2 (confirmed: they define their own attention classes rather than importing the Wan one, despite sharing a class name). Bringing items 1/2 to those models would be separate follow-up work.
- Standard (non-causal) Wan, Hunyuan, Cosmos, SD3.5, Flux, and other models don't use the chunk-by-chunk path and are unaffected by items 1, 2, or 4.
## Testing
 
All testing used `wlsaidhi/SFWan2.1-T2V-1.3B-Diffusers`, the same model the repo's own `test_causal_similarity.py` uses for this pipeline, with the same prompt that test uses ("Will Smith casually eats noodles..."), a fixed seed (1024), 480×832 resolution, 4 inference steps. Output was compared as raw pixel arrays (not re-encoded video), checking for exact equality, not similarity.
 
Hardware: single GPU (NVIDIA GB10, aarch64, CUDA 13.0). The repo's SSIM quality-regression test (`fastvideo/tests/ssim/test_causal_similarity.py`) could not be run on this hardware — it hardcodes support to A40/L40S/H100/H200/B200 and fails at collection time on any other GPU name, unrelated to this change.
 
### Correctness
 
- **Items 1/4/5 vs. unmodified code, 6 clip lengths** (9, 21, 33, 45, 69, 81 frames — the two longer lengths originally attempted, 129 and 201 frames, hit a pre-existing, unrelated cap: with the default `local_attn_size=-1` the model keeps only a 21-latent-frame compatibility window and raises a `ValueError` past that, identically on both unmodified code and this branch, confirming it's not caused by this change). All 6 pairs byte-for-byte identical, 0 differing pixels out of ~10M–97M per pair. Also 3 independent repeat runs at 81 frames alone, all 3 pairs identical.
- **Item 2, refactor only, capture still off:** re-ran the exact-frame comparison before vs. after the KV-cache restructuring alone. Byte-for-byte identical — confirms the restructuring itself changed nothing.
- **Item 2, capture on vs. off, 4 clip lengths:** 81, 129 (8 chunks past steady state), 201 (14 chunks past), and 393 frames (30 chunks past steady state). All 4 byte-for-byte identical, 0 differing pixels out of up to ~470M per pair. The longer lengths specifically check the failure mode the design has to guard against — a wrong recording only shows up well past the point the cache first fills, so this was deliberately pushed far beyond a short clip.
- **Item 2 fallback behavior**, confirmed each refuses cleanly with a logged reason and completes normally (not tested: the dual-transformer MoE case, no such test model was available):
  - `local_attn_size == -1` → logs and falls back, completes normally.
  - `rope_cache_policy == "absolute"` → logs and falls back, completes normally.
  - `dit_layerwise_offload == True` → logs and falls back, completes normally.
- `pytest fastvideo/tests/ops/test_flow_schedule_cache.py`: 10/10 pass.
- `pre-commit run --files <changed paths>`: passes on every changed file outside the pre-commit-excluded `fastvideo/models/` path (which was matched to surrounding style by hand, per repo convention).
- Full diff manually re-reviewed line by line after initial implementation (checked for leftover debug code, matched refactored math against the original branch-by-branch), plus fresh `pre-commit`, syntax, and import checks — all clean.
### Performance
 
Stage-level timing via `FASTVIDEO_STAGE_LOGGING=1`, comparing the denoising stage specifically (the part these changes touch).
 
**Items 1+4+5 vs. unmodified code, 6 clip lengths** (single run each):
 
| Frames | Unmodified | This PR (1+4+5 only) | Change |
|---|---|---|---|
| 9 | 5.17 s | 5.13 s | ~0.8% faster |
| 21 | 10.10 s | 10.03 s | ~0.7% faster |
| 33 | 15.27 s | 15.12 s | ~1.0% faster |
| 45 | 20.75 s | 20.56 s | ~0.9% faster |
| 69 | 32.35 s | 32.12 s | ~0.7% faster |
| 81 | 38.64 s | 38.32 s | ~0.8% faster |
 
Consistent, real improvement at every length tested, tightly clustered between 0.7% and 1.0%, no outliers. (Median-of-3 at 81 frames alone gave 39.04s → 38.78s, ~0.7% — consistent with the single-run numbers above.) Item 4's own effect could not be reliably separated from run-to-run noise at these clip lengths — its stage timing is only single-digit-to-low-tens of milliseconds here, small enough that measurement jitter dominates; per the design doc, its benefit is expected to grow with frame count beyond what was tested.
 
**Item 2, capture off vs. on, 4 clip lengths:**
 
| Clip length | Capture off | Capture on | Change |
|---|---|---|---|
| 81 frames | 40.44 s | 41.87 s | ~3.5% slower |
| 129 frames (8 chunks past steady state) | 65.16 s | 67.01 s | ~2.8% slower |
| 201 frames (14 chunks past steady state) | 101.58 s | 104.10 s | ~2.5% slower |
| 393 frames (30 chunks past steady state) | 199.22 s | 203.16 s | ~2.0% slower |
 
Item 2 was measured slower, not faster, at every scale tested, though the gap consistently narrows as the clip gets longer (3.5% → 2.8% → 2.5% → 2.0%) — consistent with a fixed one-time warmup/capture cost being spread over more replayed steps as the video grows, without ever crossing over to a net win in the range tested. Plausible explanation for why it doesn't cross over: this model's per-layer GPU compute time is large enough that the per-operation dispatch overhead CUDA graphs eliminate is already a small fraction of total time, while the input-staging and output-cloning the wrapper requires (needed for correctness with a stateful KV cache) adds its own overhead. No attempt has yet been made to reduce that overhead; this is the natural next step if item 2 is worth pursuing further. Given this, item 2 is included as correct, safe, off-by-default groundwork — not as a demonstrated speedup — and should not be represented as one when this merges.
 
## Known limitations / follow-up
 
- Item 2 doesn't yet improve speed on the hardware tested (see above).
- Item 2's dual-transformer (Wan2.2 MoE) case isn't supported and falls back to eager execution.
- Item 2 hasn't been tried in combination with `torch.compile` (`enable_torch_compile`) — the wrapper doesn't yet have a "drain" step for Inductor's lazy autotuning, which FlashDreams' own version handles and this port deliberately left out for now.
- Items 1/2 don't reach `matrixgame2`, `dreamx_world`, or `lingbotworld2`, which keep separate copies of this code.
- Item 4's memory cost hasn't been measured at resolutions or frame counts higher than what was tested here.
## GPU testing status
 
Done — see "Testing" above. (Not pending.)

### DIFF
--- a/fastvideo/entrypoints/video_generator.py
+++ b/fastvideo/entrypoints/video_generator.py
@@ -860,12 +860,15 @@ class VideoGenerator:
             frames = None
         else:
             videos = rearrange(samples, "b c t h w -> t b c h w")
-            frames = []
+            device_frames = []
             for x in videos:
                 x = torchvision.utils.make_grid(x, nrow=6)
                 x = x.permute(1, 2, 0).squeeze(-1)
-                x = (x * 255).to(torch.uint8)
-                frames.append(x.contiguous().cpu().numpy())
+                device_frames.append((x * 255).to(torch.uint8).contiguous())
+            # Transfer the frames in one copy. Reading them back individually
+            # blocks on the device once per frame, which serializes the
+            # remaining frames' device-side work behind each transfer.
+            frames = (list(torch.stack(device_frames).cpu().numpy()) if device_frames else [])
         postprocess_time = time.perf_counter() - postprocess_start
         logger.info("PostDecodeFrameProcessStage completed in %.3f s", postprocess_time)
         if logging_info is not None:
--- a/fastvideo/fastvideo_args.py
+++ b/fastvideo/fastvideo_args.py
@@ -151,6 +151,14 @@ class FastVideoArgs:
     torch_compile_kwargs_vae: dict[str, Any] = field(default_factory=dict)
     torch_compile_kwargs_audio_vae: dict[str, Any] = field(default_factory=dict)
 
+    # CUDA graph capture for the causal Wan denoising loop (item 2 of the
+    # FlashDreams port). Only takes effect once the KV cache window has
+    # filled and its write positions become fixed; falls back to eager
+    # execution with a log message if the pipeline/config doesn't meet the
+    # preconditions (local_attn_size == -1, rope_cache_policy != "relativistic",
+    # or a dual-transformer MoE boundary_timestep).
+    enable_causal_cuda_graph: bool = False
+
     disable_autocast: bool = False
 
     # VSA parameters
@@ -550,6 +558,14 @@ class FastVideoArgs:
             default=None,
             help=
             "JSON string of kwargs to pass to torch.compile. Example: '{\"backend\":\"inductor\",\"mode\":\"reduce-overhead\"}'",
+        )
+        parser.add_argument(
+            "--enable-causal-cuda-graph",
+            action=StoreBoolean,
+            default=FastVideoArgs.enable_causal_cuda_graph,
+            help="Capture the causal Wan denoising step into a CUDA graph once its KV cache "
+            "window reaches steady state, to cut per-op dispatch overhead. Falls back to eager "
+            "execution automatically if the pipeline doesn't meet the preconditions.",
         )
 
         parser.add_argument(
--- a/fastvideo/layers/rotary_embedding.py
+++ b/fastvideo/layers/rotary_embedding.py
@@ -553,6 +553,13 @@ def get_rotary_pos_embed(
         start_frame=start_frame,
         use_real=use_real,
     )
+    # Pin CPU tables so callers' later `.to(device)` is a pinned-memory H2D
+    # copy, not a pageable one. Callers that run inside CUDA graph capture
+    # (e.g. the causal denoising loop, once its KV cache reaches steady
+    # state) need this: capture forbids copying from pageable CPU memory.
+    if freqs_cos.device.type == "cpu" and torch.cuda.is_available():
+        freqs_cos = freqs_cos.pin_memory()
+        freqs_sin = freqs_sin.pin_memory()
     # The returned tensors are shared cache entries: callers must never mutate
     # them in place. Note .to(device) is an identity alias when the tensor is
     # already on the target device (e.g. CPU runs), so it does NOT guarantee a
--- a/fastvideo/layers/visual_embedding.py
+++ b/fastvideo/layers/visual_embedding.py
@@ -134,23 +134,38 @@ class TimestepEmbedder(nn.Module):
         return t_emb
 
 
+_TIMESTEP_EMBED_FREQS_CACHE: dict[tuple, torch.Tensor] = {}
+
+
 def timestep_embedding(t: torch.Tensor,
                        dim: int,
                        max_period: int = 10000,
                        dtype: torch.dtype = torch.float32) -> torch.Tensor:
     """
     Create sinusoidal timestep embeddings.
-    
+
     Args:
         t: Tensor of shape [B] with timesteps
         dim: Embedding dimension
         max_period: Controls the minimum frequency of the embeddings
-        
+
     Returns:
         Tensor of shape [B, dim] with embeddings
     """
     half = dim // 2
-    freqs = torch.exp(-math.log(max_period) * torch.arange(start=0, end=half, dtype=dtype) / half).to(device=t.device)
+    # The frequency basis only depends on (half, max_period, dtype), so cache
+    # it instead of rebuilding + re-transferring a CPU tensor every call.
+    # Pinned so the transfer is also valid inside CUDA graph capture (e.g.
+    # the causal denoising loop once its KV cache reaches steady state),
+    # which forbids copies from pageable CPU memory.
+    cache_key = (half, max_period, dtype)
+    freqs = _TIMESTEP_EMBED_FREQS_CACHE.get(cache_key)
+    if freqs is None:
+        freqs = torch.exp(-math.log(max_period) * torch.arange(start=0, end=half, dtype=dtype) / half)
+        if torch.cuda.is_available():
+            freqs = freqs.pin_memory()
+        _TIMESTEP_EMBED_FREQS_CACHE[cache_key] = freqs
+    freqs = freqs.to(device=t.device, non_blocking=True)
     args = t[:, None].float() * freqs[None]
     embedding = torch.cat([torch.cos(args), torch.sin(args)], dim=-1)
     if dim % 2:
--- a/fastvideo/models/dits/causal_wanvideo.py
+++ b/fastvideo/models/dits/causal_wanvideo.py
@@ -74,7 +74,7 @@ class CausalWanSelfAttention(nn.Module):
             supported_attention_backends=(AttentionBackendEnum.FLASH_ATTN,
                                           AttentionBackendEnum.TORCH_SDPA))
 
-    def forward(self, 
+    def forward(self,
                 q: torch.Tensor,
                 k: torch.Tensor,
                 v: torch.Tensor,
@@ -83,7 +83,8 @@ class CausalWanSelfAttention(nn.Module):
                 kv_cache: dict | None = None,
                 current_start: int = 0,
                 cache_start: int | None = None,
-                frame_seqlen: int = 1560):
+                frame_seqlen: int = 1560,
+                is_chunk_start: bool | None = None):
         r"""
         Args:
             x(Tensor): Shape [B, L, num_heads, C / num_heads]
@@ -92,6 +93,14 @@ class CausalWanSelfAttention(nn.Module):
             freqs(Tensor): Rope freqs, shape [1024, C / num_heads / 2]
             frame_seqlen (int): Number of tokens per latent frame,
                 e.g. 1560 for 480x832 resolution.
+            is_chunk_start: Whether this call is the first step of a new
+                autoregressive chunk (as opposed to a later denoising step
+                refining the same chunk). Callers that already know this
+                (e.g. the causal denoising stage) should pass it explicitly,
+                which lets CUDA-graph capture avoid re-deriving it from the
+                cache counters on every replay. Callers that don't pass it
+                (e.g. training) fall back to deriving it from
+                ``current_end > global_end_index``, exactly as before.
         """
         if cache_start is None:
             cache_start = current_start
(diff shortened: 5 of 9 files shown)


SolitaryThinker (changes requested):
Reviewed at `ab5d8a02e5516af453c7584cfe06dc1cb0f88265`.

Ishan — thanks for reporting the CUDA-graph result as a slowdown instead of burying it. The KV-cache counter bug you found (Python-side update sitting inside the region a replay skips) is a nasty one. I reproduced your byte-identical claims on different hardware, both halves: 0 differing values out of 97,044,480.

Main ask is to split this. The two halves go in opposite directions.

### Always-on changes: much bigger win than you measured

GB200, CUDA 13.0, torch 2.12, `wlsaidhi/SFWan2.1-T2V-1.3B-Diffusers`, 480x832, 81 frames, 4 DMD steps, seed 1024. Alternating base and branch processes in one session, 3 generations each, 4 rounds:

- denoise: -16% to -34%, median -26%
- e2e: -11% to -25%, median -19%

Branch holds 4471-4659 ms across all 12 generations. Base swings 5316-7010 ms. That's ~2,100 host stalls per generation going away.

Items 1/4/5 plus the `torch.tensor([0], device=...)` -> `torch.zeros(...)` change should be their own PR. Byte-identical, large, no reason to wait behind the capture work.

### Capture: has a bug that stops it ever helping

`causal_denoising.py:332` builds `t_expanded_noise` as `t_cur * torch.ones(..., dtype=torch.long)`. `warp_denoising_step` is true for every Wan DMD preset, so `t_cur` is float32 and the product promotes to float32. `t_expanded_context` at `:417` is int64. Both go through `_continuation`, `_slot_compatible` compares dtype, so every denoise/context alternation calls `reset()` and drops the graph.

That works out to one capture per autoregressive block. I derived 6 at 81 frames and 10 at 129 from the block count, then measured exactly 6 (three times) and exactly 10. Each capture does a full `synchronize()` + `empty_cache()` via `torch.cuda.graph.__enter__`, and the count scales with clip length, so it never amortizes. Capture on: +7.4% e2e, +11.7% denoise, against the same branch with it off.

Give the context step its own wrapper, or make the two dtypes agree. Then re-measure — no idea yet whether it wins once it stops thrashing.

### Rest

- `rotary_embedding.py:560-562` pins host memory in a helper 13 model files call. Exactly one (`causal_wanvideo.py:692`) passes `non_blocking=True`. A 21-frame 480x832 table pair is 67.1 MB and the cache holds 16, so ~1.07 GB pinned per process. Hunyuan, HyWorld, MatrixGame, DreamX and LingbotWorld all pay that for nothing. Lifetime hazard too: a captured async copy bakes in the host pointer, and that cache evicts. Make it opt-in, or cache device-side in the one caller that benefits.
- `visual_embedding.py:164` — the real defect is the missing `device=` on `torch.arange`. `longcat.py:110`, `zimage.py:65`, `kandinsky5.py:34`, `ltx2.py:191` and `visual_embedding.py:252` all pass it. Put `t.device` in the cache key and both the transfer and the module-global pinned buffer go away.
- `causal_denoising.py:428` and `:444` hard-code `is_chunk_start=False` where line 340 passes `None`. Two transformers with a timestep list entirely on one side of the boundary: one cache goes a block untouched, the context call is its first write of that block, `_write_in_place` overshoots. Pass `None`. Derived, not executed — no MoE causal checkpoint here.
- `FlashDreams.md` is referenced 5x (`causal_denoising.py:75,104`, `causal_cuda_graph.py:36,169`, `causal_wanvideo.py:288`) and isn't in the repo or its history. `:104` hands the whole capture safety derivation to it.
- 8 of 9 files have no test, including the capture module and the KV-cache restructure that changes the default path. Cheapest useful one: drive `_evict_and_write`/`_write_in_place` over a synthetic cache dict for a multi-block rollout and assert the index pair against the pre-refactor expressions. No model, no GPU.
- Nothing shipped can reach the capture path. `local_attn_size` and `rope_cache_policy` are arch fields with no user-facing surface and no override in the tree, and `dit_layerwise_offload` defaults to on, so our own causal example gets refused. The docstring lists three preconditions and omits that one. I had to patch a config to reach the code at all.

Minor: the dispatch takes `transformer` and ignores it on the graph path; the two graphs don't share a memory pool; `call_current_end` is used outside the loop that binds it; the batched frame copy holds the per-frame list and the stacked tensor at once, so peak is ~2x the quoted number.

One correction to the bot comment above: capture doesn't run the recorded kernels, so the immediate `graph.replay()` isn't a double-apply. Checked directly — an in-place increment inside the capture leaves the buffer untouched, then increments once per replay. Your comment at `causal_cuda_graph.py:138-141` is correct.

PR #1665, opened 2026-07-30
[new-model] Added LingBot-World-Fast image-to-video support

Ishxn20:
## Summary

- Wires up `FastVideo/LingBot-World-Fast-Diffusers`, the old LingBot-World repo's distilled causal I2V checkpoint (`generate_fast.py` / `LingBot-World-Fast`) that was never ported.
- Reuses the existing LingBot World 2 causal-fast DiT unchanged — verified all 1421 parameter names and shapes match the released checkpoint exactly. Only the arch config differs: `local_attn_size=-1` (global attention, no eviction), `chunk_size=3`, `timesteps_index=(0,179,358,679)`.
- Adds the arch/pipeline configs, registry entries, preset, pipeline subclass, example script, and SSIM test needed to run it.
- Fixes a model-family detector collision: the LingBot World (Cam) detector only excluded the `"causal-fast"` spelling, so it would have wrongly claimed this checkpoint's path and its `lingbotworldcausaldmdpipeline` model_index name. Registered LingBot-World-Fast first and tightened the exclusion.
- Fixes `WanCrossAttention` in the shared LingBot World 2 causal-fast DiT to route through the `attention()` dispatcher instead of calling `flash_attention()` directly, so the already-advertised `TORCH_SDPA` backend actually works when flash-attn isn't installed. Self-attention already used this dispatcher; cross-attention did not.

## Test plan

- [x] Verified all 1421 transformer parameter names and shapes against the real downloaded checkpoint (name + shape match, exact)
- [x] Verified effective config after `update_model_arch` matches the reference `generate_fast.py` (`chunk_size=3`, `timesteps_index=(0,179,358,679)`, `sample_shift=10.0`)
- [x] Verified registry resolution for all three LingBot checkpoints (Base-Cam, Fast, v2 causal-fast) — no cross-contamination
- [x] `pre-commit run --files ...` clean (yapf, ruff, codespell, mypy)
- [x] Loaded real weights end-to-end (18.54B params, 16 shards) and generated a short video on a single GPU (no FSDP, no flash-attn — SDPA fallback) — produced a coherent, prompt-following video with correct camera motion
- [ ] SSIM regression test — needs 2 GPUs, unavailable in dev environment; will run in CI. First reference video should upload as a draft since this is tagged `[new-model]` (bootstrap mode)
- [ ] Activation-trace parity against the original `wan/modules/model_fast.py` — the original also requires flash-attn, unavailable in dev environment

### DIFF
--- a/examples/inference/basic/basic_lingbotworld_fast.py
+++ b/examples/inference/basic/basic_lingbotworld_fast.py
@@ -15,14 +15,12 @@ def main():
     )
 
     prompt = (
-        "The video presents a soaring journey through a fantasy jungle. The "
-        "wind whips past the rider's blue hands gripping the reins, causing "
-        "the leather straps to vibrate. The ancient gothic castle approaches "
-        "steadily, its stone details becoming clearer against the backdrop of "
-        "floating islands and distant waterfalls.")
-    image_path = ("https://raw.githubusercontent.com/Robbyant/lingbot-world/"
-                  "main/examples/00/image.jpg")
-    action_path = "examples/inference/basic/lingbotworld_examples/00"
+        "A serene lakeside scene with a lone tree standing in calm water, "
+        "surrounded by distant snow-capped mountains under a bright blue sky "
+        "with drifting white clouds — gentle ripples reflect the tree and sky, "
+        "creating a tranquil, meditative atmosphere.")
+    action_path = "examples/datasets/lingbotworld2"
+    image_path = f"{action_path}/image.jpg"
 
     generator.generate_video(
         prompt,
--- a/examples/inference/basic/basic_lingbotworld_fast.py
+++ b/examples/inference/basic/basic_lingbotworld_fast.py
@@ -1,0 +1,40 @@
+from fastvideo import VideoGenerator
+
+OUTPUT_PATH = "video_samples_lingbotworld_fast"
+
+
+def main():
+    generator = VideoGenerator.from_pretrained(
+        "FastVideo/LingBot-World-Fast-Diffusers",
+        num_gpus=1,
+        use_fsdp_inference=False,  # set to True if GPU is out of memory
+        dit_cpu_offload=True,
+        vae_cpu_offload=False,
+        text_encoder_cpu_offload=True,
+        pin_cpu_memory=True,
+    )
+
+    prompt = (
+        "The video presents a soaring journey through a fantasy jungle. The "
+        "wind whips past the rider's blue hands gripping the reins, causing "
+        "the leather straps to vibrate. The ancient gothic castle approaches "
+        "steadily, its stone details becoming clearer against the backdrop of "
+        "floating islands and distant waterfalls.")
+    image_path = ("https://raw.githubusercontent.com/Robbyant/lingbot-world/"
+                  "main/examples/00/image.jpg")
+    action_path = "examples/inference/basic/lingbotworld_examples/00"
+
+    generator.generate_video(
+        prompt,
+        image_path=image_path,
+        action_path=action_path,
+        output_path=OUTPUT_PATH,
+        save_video=True,
+        num_frames=81,
+        height=480,
+        width=832,
+    )
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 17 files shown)


SolitaryThinker (changes requested):
Reviewed exact head `2c24e0ec29d4b917b946d068b753556b35e1f764` — changes requested.

Thanks for keeping this variant small by reusing the LingBot World 2 causal-fast implementation. I verified the released checkpoint class/config, the FastVideo registry and pipeline-class resolution, and the reference sampler's fixed chunk/timestep contract. The registration and scheduler replacement are pointed at the right underlying problem.

Blocking findings:

1. `TORCH_SDPA` is still ignored when FlashAttention is installed (`fastvideo/models/dits/lingbotworld2/causal_fast.py:154-195,399`). The changed cross-attention call now reaches `attention()`, but that shared helper chooses FlashAttention whenever either package is importable; it never consults FastVideo's environment/global backend selection. With `FASTVIDEO_ATTENTION_BACKEND=TORCH_SDPA` and FlashAttention availability mocked true, an exact-head check fails with `AssertionError: TORCH_SDPA selection dispatched to FlashAttention`. Fix the shared helper once so both self- and cross-attention honor the existing selector, and leave a small regression test covering explicit SDPA while FlashAttention is present.

2. The new preset accepts two overrides that the sampler discards (`fastvideo/pipelines/basic/lingbotworld_fast/presets.py:10-13`). `LingBotWorld2CausalFastGenerationStage.forward()` always indexes the four `cfg.timesteps_index` entries (`causal_fast_pipeline.py:260-261`) and never reads `batch.num_inference_steps` or `batch.guidance_scale`. An exact-head contract check reports `accepted denoise overrides are ignored: ['guidance_scale', 'num_inference_steps']`. The released fast sampler is fixed-step and has no CFG argument, so the minimal fix is to stop advertising these as allowed stage overrides; otherwise wire and test real semantics.

3. This new model still lacks numerical acceptance proof. The PR records name/shape equality and one coherent generation, but explicitly has no official-vs-FastVideo activation/latent parity, and its only authored regression is a two-GPU FlashAttention SSIM test that has not run. Locally it collects but skips on CPU; a bootstrap reference generated by this implementation cannot establish official parity. Please add a non-skipping parity test/evidence for the reused DiT under this variant's global-attention/chunk/timestep config, a lightweight registry/pipeline/default smoke test, and run/seed the SSIM case. Record the reproducible commands/results under `tests/local_tests/lingbotworld_fast/` (README/port status), consistent with the other model ports.

The head also conflicts with current `main`, so it needs a rebase after the fixes.

Evidence at this head:

- `[measured]` `pre-commit run --files <12 changed paths>`: passed.
- `[measured]` exact checkpoint registry + pinned pipeline resolution: passed.
- `[measured]` authored SSIM test: `1 skipped` (CPU; no GPU/reference result).
- `[read]` sibling sweep covered all LingBot detectors/class aliases, both callers of the shared attention helper, pipeline initialization order, and preset-to-batch override flow.

SolitaryThinker:
Rebased onto current `main` to resolve the config-export conflicts; no PR content changed, and the range-diff differs only at the two mechanical union resolutions.

SolitaryThinker:
Follow-up at `1a1c77089add53dc6feb7612f49749372971b17a` addresses all non-parity/non-SSIM review findings:

- LingBot self- and cross-attention now use FastVideo's shared per-component backend selector, so an explicit or global `TORCH_SDPA` request wins even when FlashAttention is installed.
- The fixed four-step/no-CFG sampler no longer advertises `num_inference_steps` or `guidance_scale` as stage overrides.
- Added CPU-only regression coverage for SDPA dispatch and for preset/registry/exact pipeline resolution. Ishan's existing checkpoint/config/pipeline wiring remains intact.

Evidence:
- [measured] Production fixes reverted while retaining the tests: **2 failed, 65 passed**.
- [measured] Fixes restored: **67 passed**; selector/config sibling suites: **43 passed**.
- [measured] Local pre-commit chain passed; GitHub pre-commit and trigger checks are green.
- [read] Sibling sweep confirms both LingBot attention callers route through the same selected-backend helper.

Component/pipeline numerical parity and the SSIM reference run remain intentionally outstanding.

PR #1667, opened 2026-07-30
[misc]: consolidate dataset download scripts under examples/datasets/

SuhaanCoding:
## Problem

Dataset download scripts were scattered across model specific example folders, with heavy duplication:

- crush-smol*: 6 byte identical `download_dataset.sh` copies (`wan_t2v_1.3B`, `Wan2.1-Fun-1.3B-InP`, `wan_i2v_14B_480p`, `Wan2.2-TI2V-5B-Diffusers` distill, `ltx2`, `SFWan2.1-T2V`) all fetching `wlsaidhi/crush-smol-merged`
- Wan-Syn: 2 near-duplicates (VSA's downloaded 480P+720P, distill's 480P only)
- mixkit: 1 script buried under `wan_t2v_1.3B/mixkit/`

a dedicated `examples/dataset/` folder already existed but held only `vidprom` and `lingbotworld2`.

## Solution

Move all dataset download scripts into `examples/datasets/`, keyed by dataset (matching the existing `dataset/` convention):

Design choices (happy to adjust):
- Dataset-keyed layout rather than per-model: the duplicated scripts were identical, and the existing `dataset/` folder already uses this convention. Model-specific variants, if ever needed, can be added inside the dataset's folder
- Clean break: old scripts deleted and all 20 referencing files updated
- Scripts only: `validation.json` / `validation_dataset` files stay with their model examples, so nothing under `fastvideo/tests/` is touched
- Renamed `examples/dataset/` → `examples/datasets/`

Also fixes in passing: `Wan2.1-Fun-1.3B-InP/crush_smol/README.md` pointed at the wrong `wan_i2v_14b_480p` download path.

## Test evidence

- `pre-commit run --files <all changed>`: all hooks pass (yapf, ruff, codespell, PyMarkdown, mypy)
- Repo-wide grep sweep confirms zero stale references to any old script path
- No behavior changes: the consolidated scripts are byte equivalent to the copies they replace (Wan-Syn's combined script split into its two constituent commands verbatim)


SolitaryThinker (approved):
thanks for the contribution!

PR #1677, opened 2026-08-03
[kernel] attention: sm_100a CUDA forward for block-causal-sink (Blackwell)

lpc0220:
## Purpose

Speeds up the full-sequence block-causal + sink + sliding-window attention on Blackwell
(sm_100a) by replacing **only the forward** of `_BlockCausalSinkAttention` with a CUDA kernel.
The Triton backward and its kernels are untouched: the CUDA forward emits `lse` in exactly the
form the backward already consumes -- `max(qk*qk_scale) + log2(l)` in the exp2 domain, matching
`qk_scale = sm_scale * LOG2E` -- so the two halves compose unchanged.

**Performance.** Wan config (H=12, `frame_seqlen`=1456, `num_frame_per_block`=3, `sink_size`=1,
`local_attn_size`=6), visited-pair TFLOPS, 1 CTA:

| frames | seq len | CUDA | Triton | speedup |
|-------:|--------:|-----:|-------:|--------:|
|      6 |   8,736 | 1162 |    443 |  2.62x  |
|     12 |  17,472 | 1281 |    495 |  2.59x  |
|     18 |  26,208 | 1361 |    523 |  2.60x  |
|     36 |  52,416 | 1412 |    546 |  2.59x  |
|     72 | 104,832 | 1400 |    557 |  2.51x  |

## Changes

- `csrc/attention/block_causal_sink_sm100a.cu` -- torch binding, returns `(out, lse)`.
- `csrc/attention/block_causal_sink_{kernel,launch}_sm100a.cuh` -- warp-specialized kernel
  (load / tcgen05 MMA / softmax / correction / epilogue / scheduler) and its launch entry.
- `csrc/attention/primitives.cuh` -- the tcgen05 / TMA / mbarrier primitives it needs, flattened
  into one header so no external tree has to be vendored.
- `csrc/common_extension.cpp`, `CMakeLists.txt` -- pybind registration, sm_100a build gate.
- `fastvideo/attention/kernels/block_causal_sink_cuda.py` -- backend + `is_supported()`.
- `fastvideo/attention/kernels/block_causal_sink.py` -- **+18 lines**: a guarded branch in
  `forward`. Everything after it, including `ctx.save_for_backward(...)`,

**No copies for the supported layout.** q/k/v/o are consumed as contiguous `[B,H,L,D]` --
the layout FastVideo already passes -- so nothing is reordered on either side. V in
particular is read **MN-major** rather than requiring a pre-transposed `V_T`, which would
have cost ~80us of copies against a ~37us kernel.

**Two build requirements that fail confusingly if missed** (both handled

- `-gencode arch=compute_100a,code=sm_100a`. Plain `-arch=sm_100a` silent
  target and ptxas rejects every `tcgen05` / `setmaxnreg` / `.cta_group` instruction.
- link `libcuda`: `cuTensorMapEncodeTiled` is a driver-API symbol, so the
  builds and links fine and then fails at *import* with `undefined symbol`.

### 1. Build the op and route FastVideo's own entry point through it, then path on identical inputs and diff. H=12, L=8736 (nf=6, tpf=1456, nfpb=3, sink=1, win=6).

```bash
python - <<'EOF'
import torch
from fastvideo.attention.kernels import block_causal_sink as bcs, block_c
from fastvideo.models.dits._causal_train_attention import CausalTrainAttentionPlan
B,H,L,D = 1,12,1456*6,128
torch.manual_seed(0)
q,k,v = (torch.randn(B,H,L,D, device="cuda", dtype=torch.bfloat16) for _
plan = CausalTrainAttentionPlan(kind="blockwise", impl="triton", num_frames=6,
        frame_seqlen=1456, num_frame_per_block=3, local_attn_size=6, sink
        sm_scale=1/(D**0.5))
print("is_supported:", cu.is_supported(plan, q))
o_cuda = bcs.block_causal_sink_attention(q,k,v,plan)
cu._HAS_CUDA_BCS = False                      # force Triton
o_tri  = bcs.block_causal_sink_attention(q,k,v,plan)
print("max |diff|:", (o_cuda.float()-o_tri.float()).abs().max().item())
EOF
```

Test Results

<details>

is_supported: True
max |diff|: 0.00098          # bf16 output precision

lse vs Triton forward:  max |diff| = 6e-6   (default sm_scale and sm_scal
  negative control (kernel denied sm_scale): 0.796  -> the check has teeth

fallback with extension absent: is_supported() False, Triton path runs, output finite

Performance -- Wan sweep, H=12 tpf=1456 nfpb=3 sink=1 win=6, visited-pair TFLOPS:
  nf          6      12      18      36      72
  CUDA     1162    1281    1361    1412    1400
  Triton    443     495     523     546     557
  ratio    2.62x   2.59x   2.60x   2.59x   2.51x

</details>

Checklist

Notes for reviewers

- Memory: no additional allocations beyond out and lse ([B*H, L] fp32 = 1/64 the
bytes of out). No .contiguous() copies of q/k/v.
- Not yet done, flagged deliberately: I have not run pre-commit, added tests to this
repo, or built through build.sh/CMake -- the kernel was compiled with the flags
CMakeLists.txt specifies, so the CMake wiring itself is unexercised. The backward was not
run end-to-end either; only lse parity (6e-6) was verified. Happy to close any of these.
- SSIM / model tests not run: this changes a kernel behind an existing op, and falls back to
Triton wherever is_supported() is False.

### DIFF
--- a/fastvideo-kernel/CMakeLists.txt
+++ b/fastvideo-kernel/CMakeLists.txt
@@ -308,6 +308,21 @@ if(BUILD_CXX_KERNELS)
         csrc/turbodiffusion/quant/quant.cu
     )
     
+    # Blackwell block-causal + sink + sliding-window attention (sm_100a only).
+    # NOTE the "a" suffix: -arch=sm_100a is NOT enough -- it emits a plain sm_100 target and
+    # ptxas rejects every tcgen05 / setmaxnreg / .cta_group instruction. The explicit gencode
+    # spelling below is required, and matches the 10.0a entry in TORCH_CUDA_ARCH_LIST.
+    set(ENABLE_BCS_SM100A OFF)
+    if(TORCH_CUDA_ARCH_LIST MATCHES "(^|[; ,])(10\\.0a|100a|sm_100a)([; ,]|$)")
+        set(ENABLE_BCS_SM100A ON)
+    endif()
+    if(ENABLE_BCS_SM100A)
+        message(STATUS "fastvideo-kernel: building block_causal_sink_sm100a (Blackwell)")
+        list(APPEND EXTENSION_SOURCES csrc/attention/block_causal_sink_sm100a.cu)
+        set_source_files_properties(csrc/attention/block_causal_sink_sm100a.cu PROPERTIES
+            COMPILE_OPTIONS "-gencode;arch=compute_100a,code=sm_100a")
+    endif()
+
     # Conditionally add TK kernels
     if(ENABLE_TK_KERNELS)
         list(APPEND EXTENSION_SOURCES
@@ -336,6 +351,9 @@ if(BUILD_CXX_KERNELS)
     if(ENABLE_TK_KERNELS)
         list(APPEND COMPILE_DEFS TK_COMPILE_ST_ATTN TK_COMPILE_BLOCK_SPARSE)
     endif()
+    if(ENABLE_BCS_SM100A)
+        list(APPEND COMPILE_DEFS TK_COMPILE_BLOCK_CAUSAL_SINK_SM100A)
+    endif()
 
     target_compile_definitions(fastvideo_kernel_ops PRIVATE ${COMPILE_DEFS})
 
@@ -346,6 +364,13 @@ if(BUILD_CXX_KERNELS)
     # Link against Torch libraries to avoid undefined symbols at import time
     # (e.g., torch::autograd vtables) when loading the extension module.
     target_link_libraries(fastvideo_kernel_ops PRIVATE ${TORCH_LIBRARIES})
+
+    # The Blackwell kernel builds its TMA descriptors with cuTensorMapEncodeTiled, a CUDA
+    # DRIVER API entry point -- it is not in libcudart, so the module fails to import with
+    # "undefined symbol: cuTensorMapEncodeTiled" unless libcuda is linked explicitly.
+    if(ENABLE_BCS_SM100A)
+        target_link_libraries(fastvideo_kernel_ops PRIVATE cuda)
+    endif()
 
     # Also link against libtorch_python to satisfy Python-binding symbols
     # (e.g., torch::PyWarningHandler) required by torch/extension.h.
--- a/fastvideo-kernel/csrc/attention/block_causal_sink_kernel_sm100a.cuh
+++ b/fastvideo-kernel/csrc/attention/block_causal_sink_kernel_sm100a.cuh
@@ -1,0 +1,1012 @@
+// block_causal_sink_kernel_sm100a.cuh -- block-causal + sink + sliding-window FMHA forward, sm_100a.
+// Warp-specialized: load / MMA (tcgen05) / softmax / correction / epilogue / scheduler.
+// Writes O and, when asked, the log-sum-exp the backward consumes.
+//
+// Generated (comments stripped). Do not edit by hand.
+#pragma once
+
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <cuda_bf16.h>
+#include <cstdio>
+#include <cstdlib>
+#include <cstdint>
+#include <type_traits>
+#include <cstring>
+#include <cmath>
+#include <vector>
+#include "primitives.cuh"
+
+constexpr int M_TILE = 128;
+constexpr int M_TILES_PER_CTA = 2;
+constexpr int K_TILE = 128;
+constexpr int HEAD_DIM = 128;
+
+constexpr int SUB_COLS_BF16 = 64;
+constexpr int SUB_COLS_BYTES = SUB_COLS_BF16 * (int)sizeof(__nv_bfloat16);
+constexpr int Q_SUBTILES = HEAD_DIM / SUB_COLS_BF16;
+constexpr int K_SUBTILES = HEAD_DIM / SUB_COLS_BF16;
+constexpr int V_SUBTILES = K_TILE / SUB_COLS_BF16;
+constexpr int P_SUBTILES = K_TILE / SUB_COLS_BF16;
+constexpr int Q_SUB_COLS_BYTES = M_TILE * SUB_COLS_BYTES;
+constexpr int K_SUB_COLS_BYTES = K_TILE * SUB_COLS_BYTES;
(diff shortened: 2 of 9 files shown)


shaoxiongduan:
Hi! I took a look through the kernel and most of it looks fine, but `rolling_window_tokens` is `sink_size` frames too wide.

Specifically,

`block_causal_sink_cuda.py` passes

```python
plan.local_attn_size * plan.frame_seqlen,  # rolling_window_tokens
```

but `local_attn_size` is the **total** KV budget with the sink counted inside it. `_plan_scalars`
and `reference_causal_train_attention` both use
`ROLLING_TOK = max(0, local_attn_size - sink_size) * frame_seqlen`
(`_causal_train_attention.py:268`; `validate_causal_attention_geometry` enforces the accounting by
rejecting `sink_size + num_frame_per_block > local_attn_size`). Since the kernel's mask is
`keep_hi & (keep_sink | keep_win)` with `window_start = block_end - rolling_window_tokens`, the
CUDA path attends to `sink_size` extra frames per query block.

**Why the parity test can't see it.** At `nf=6, local_attn_size=6` the window never scrolls past
the sink, so both formulations collapse to `[0, block_end)` and the two masks are *identical*.
Divergence starts at `num_frames > local_attn_size`. That config also makes
`build_sink_delta_tables` return `None`, so `HAS_SINK_ROPE_DELTA=true` is never instantiated —
the re-rope path isn't covered either.

**Measured.** Built the branch for sm_100a (`cpp_extension.load`, `TORCH_CUDA_ARCH_LIST=10.0a`,
`-gencode arch=compute_100a,code=sm_100a`, `-lcuda`) and ran `block_causal_sink_attention` both
ways on one GB200, `[1, 12, L, 128]` bf16:

| config | CUDA vs Triton |
|---|---|
| `nf=6, frame_seqlen=1456` (this PR's config) | max\|Δ\| = 0.000977 — reproduces the reported 0.00098 |
| `nf=21, frame_seqlen=1560` (the causal training configs) | max\|Δ\| = 0.098, mean 0.0037 |
| per-block max at `nf=21` | `[0.001, 0.0005, 0.075, 0.051, 0.053, 0.098, 0.051]` |

Blocks 0–1 sit at the bf16 floor; every block from 2 on is wrong — exactly the point where the
window scrolls past the sink.

**The kernel itself is correct.** Matching the window on the Triton side (same δ tables,
`local' = local + sink`, which reproduces the pair of numbers the wrapper passes) gives
`max|Δ| = 0.000977`. So the tile ranges, the `q`/`q_sink` swap and the twice-visited boundary tile
are all fine — it is only the argument.

**Forward and backward also disagree.** `ctx.plan` still carries the narrow window, so the Triton
backward re-derives `p = exp2(qk*scale - lse)` under the narrow mask from an `lse` produced under
the wide one. With `nf=21` and the relativistic correction on, gradients land ~13% of `max|dq|`
away from the all-Triton result, and are not the derivative of the forward that ran.

fix is just: 

```diff
-        plan.local_attn_size * plan.frame_seqlen,  # rolling_window_tokens
+        max(0, plan.local_attn_size - plan.sink_size) * plan.frame_seqlen,
```

lpc0220:
> Hi! I took a look through the kernel and most of it looks fine, but `rolling_window_tokens` is `sink_size` frames too wide.
> 
> Specifically,
> 
> `block_causal_sink_cuda.py` passes
> 
> ```python
> plan.local_attn_size * plan.frame_seqlen,  # rolling_window_tokens
> ```
> 
> but `local_attn_size` is the **total** KV budget with the sink counted inside it. `_plan_scalars` and `reference_causal_train_attention` both use `ROLLING_TOK = max(0, local_attn_size - sink_size) * frame_seqlen` (`_causal_train_attention.py:268`; `validate_causal_attention_geometry` enforces the accounting by rejecting `sink_size + num_frame_per_block > local_attn_size`). Since the kernel's mask is `keep_hi & (keep_sink | keep_win)` with `window_start = block_end - rolling_window_tokens`, the CUDA path attends to `sink_size` extra frames per query block.
> 
> **Why the parity test can't see it.** At `nf=6, local_attn_size=6` the window never scrolls past the sink, so both formulations collapse to `[0, block_end)` and the two masks are _identical_. Divergence starts at `num_frames > local_attn_size`. That config also makes `build_sink_delta_tables` return `None`, so `HAS_SINK_ROPE_DELTA=true` is never instantiated — the re-rope path isn't covered either.
> 
> **Measured.** Built the branch for sm_100a (`cpp_extension.load`, `TORCH_CUDA_ARCH_LIST=10.0a`, `-gencode arch=compute_100a,code=sm_100a`, `-lcuda`) and ran `block_causal_sink_attention` both ways on one GB200, `[1, 12, L, 128]` bf16:
> 
> config	CUDA vs Triton
> `nf=6, frame_seqlen=1456` (this PR's config)	max|Δ| = 0.000977 — reproduces the reported 0.00098
> `nf=21, frame_seqlen=1560` (the causal training configs)	max|Δ| = 0.098, mean 0.0037
> per-block max at `nf=21`	`[0.001, 0.0005, 0.075, 0.051, 0.053, 0.098, 0.051]`
> Blocks 0–1 sit at the bf16 floor; every block from 2 on is wrong — exactly the point where the window scrolls past the sink.
> 
> **The kernel itself is correct.** Matching the window on the Triton side (same δ tables, `local' = local + sink`, which reproduces the pair of numbers the wrapper passes) gives `max|Δ| = 0.000977`. So the tile ranges, the `q`/`q_sink` swap and the twice-visited boundary tile are all fine — it is only the argument.
> 
> **Forward and backward also disagree.** `ctx.plan` still carries the narrow window, so the Triton backward re-derives `p = exp2(qk*scale - lse)` under the narrow mask from an `lse` produced under the wide one. With `nf=21` and the relativistic correction on, gradients land ~13% of `max|dq|` away from the all-Triton result, and are not the derivative of the forward that ran.
> 
> fix is just:
> 
> ```diff
> -        plan.local_attn_size * plan.frame_seqlen,  # rolling_window_tokens
> +        max(0, plan.local_attn_size - plan.sink_size) * plan.frame_seqlen,
> ```

Done. Added a unit test file.

SolitaryThinker (approved):
Thanks for the contribution!

PR #1682, opened 2026-08-05
[ci]: add golden-gate lane — single-layer bitwise DiT fingerprints for all SSIM-covered families

SolitaryThinker:
## What

A **golden-gate** test loads ONE transformer layer of a DiT — located via the safetensors index, so ~1–2 GB moves instead of the full checkpoint — runs a fixed seeded packed batch through it, and compares the output **bitwise** against a device-keyed golden latent.

**Coverage: 17 gates across all 16 SSIM-covered DiT families** — wan (arch shared by SFWan/TurboWan), minimax_h3, gamecraft, flux, flux2-klein, sd35, ltx2, longcat, kandinsky5, lingbot, matrixgame2 + matrixgame3, dreamx-cam, glm-image, zimage, stable-audio, gen3c. A shared harness (`_harness.py`) owns env pinning, selective shard loading with per-model param renames (index-sharded and single-file checkpoints), golden resolution (local dir → HF dataset → mint-and-fail), the environment-fingerprint assertion, and the bitwise compare; each model contributes a ~90-line `GateSpec`.

New lane: `/test golden-gate` → Buildkite `golden_gate` step → `pr_test.py::run_golden_gate_tests` (L40S:1).

## Why zero tolerance is earned, not hoped for

Verified on GB200 (2026-08-05):
- **all 17 gates minted and re-verified at drift 0.0** across independent processes — every covered architecture is bit-stable on one device,
- the full 50-layer H3 forward holds `atol=0, rtol=0` against pinned Diffusers (`tests/local_tests/transformers/test_minimax_h3_transformer_parity.py`),
- two end-to-end seeded H3 T2V generations produced **md5-identical videos**.

Any drift at a gate is a real change to the compute path. A warm gate verifies in seconds (the two matrixgame gates verify together in 4.4 s) vs minutes-to-an-hour for the SSIM generations they front-run: a green gate proves the DiT compute path is bit-identical, so the SSIM lane cannot have regressed for that model; a red gate names either the exact drift or — via the embedded environment fingerprint (torch/CUDA/cuDNN/flash-attn/TF32/attention backend/`FASTVIDEO_FA4`) — the environment change that requires a deliberate golden re-mint. The fingerprint exists because an unpinned `FASTVIDEO_FA4` cost a full reference-reseed cycle this week.

## Goldens

Device-keyed like SSIM references, same HF dataset, `golden_gates/<device>/`. **All 17 GB200 goldens are uploaded.** L40S goldens mint themselves on the first CI run and need a one-time upload (the SSIM reference seeding flow). Local `goldens/` dir or `FASTVIDEO_GOLDEN_GATE_DIR` wins over the download.

Determinism hardening baked in: `torch.use_deterministic_algorithms(True)` tripwire, `cudnn.benchmark=False`, `CUBLAS_WORKSPACE_CONFIG=:4096:8`, single GPU, attention backend pinned per family (matching each SSIM test).

## Per-model notes (encoded in the specs)

- **matrixgame2 runs fp32 end-to-end**: its int64 null-modulation promotes activations to fp32 mid-block (`causal_model.py:569` + `layernorm.py` scale-shift promotion), which would crash any bf16 deployment — the gate carries `GateSpec.dtype=float32`.
- **stable_audio** gates in bf16 while its pipeline runs fp16 — self-consistent, flagged for a deliberate follow-up.
- **dreamx** gates the Cam variant; the AR variant's KV-cache forward needs its own recipe. **lingbot** gates the high-noise expert only.

## Files

- `fastvideo/tests/golden_gate/_harness.py` — shared machinery
- `fastvideo/tests/golden_gate/test_*.py` — 16 per-family gate files (17 gates)
- `fastvideo/tests/modal/pr_test.py` — `run_golden_gate_tests` (L40S:1)
- `.github/workflows/ci-slash-commands.yml` — `golden-gate` in VALID + MAP
- `.buildkite/pipeline.yml` + `.buildkite/scripts/pr_test.sh` — dispatch

Possible follow-up (not in this PR): add `golden-gate` to the `fastcheck` meta so every PR gets the fingerprint sweep by default.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -68,6 +68,17 @@ steps:
             limit: 2
       agents:
         queue: "default"
+    - label: ":vertical_traffic_light: Golden-Gate Tests"
+      if: build.env("TEST_SCOPE") == "direct" && build.env("TEST_TYPE") == "golden_gate"
+      command: "timeout 90m .buildkite/scripts/pr_test.sh"
+      retry:
+        automatic:
+          - exit_status: 128
+            limit: 3
+          - exit_status: -1
+            limit: 2
+      agents:
+        queue: "default"
     - label: ":microscope: Unit Tests"
       if: build.env("TEST_SCOPE") == "direct" && build.env("TEST_TYPE") == "unit_test"
       command: "timeout 90m .buildkite/scripts/pr_test.sh"
--- a/.buildkite/scripts/pr_test.sh
+++ b/.buildkite/scripts/pr_test.sh
@@ -187,6 +187,10 @@ case "$TEST_TYPE" in
         log "Running transformer tests..."
         MODAL_COMMAND="$MODAL_ENV HF_API_KEY=$HF_API_KEY python3 -m modal run $MODAL_TEST_FILE::run_transformer_tests"
         ;;
+    "golden_gate")
+        log "Running golden-gate tests..."
+        MODAL_COMMAND="$MODAL_ENV HF_API_KEY=$HF_API_KEY python3 -m modal run $MODAL_TEST_FILE::run_golden_gate_tests"
+        ;;
     "ssim")
         log "Running SSIM tests..."
         SSIM_BOOTSTRAP_ARGS=$(ssim_bootstrap_args)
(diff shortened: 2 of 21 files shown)


SolitaryThinker:
Expanded from one model to the full SSIM matrix: 17 gates across all 16 covered DiT families, all minted and re-verified at bitwise drift 0.0 on GB200. Goldens for every gate are on the dataset under `golden_gates/NVIDIA_GB200/`. A warm verify of a gate runs in seconds (the two matrixgame gates verify together in 4.4s). Per-model caveats are in the commit message; L40S goldens still mint themselves on first CI run per the seeding flow.

PR #1683, opened 2026-08-05
[new-model] Add Wan2.2-S2V-14B speech-to-video (model port + checkpoint converter)

SuhaanCoding:
## Purpose

Adds **Wan2.2-S2V-14B**, a speech-to-video model.

You give it a picture of a person and a voice recording. It makes a video of
that person moving and talking in time with the sound. The text prompt still
sets the scene and the camera.

This model was on the project wish list. There was no GitHub issue for it, so
this PR is the first item.

## Changes

**What is in this PR:** the model, the pipeline, the audio handling, the tests,
and a script that repacks the official download into the folder layout this
repo can load.

**Update — official weights are live.** The canonical repo this PR's
registry entry points at now exists:
**https://huggingface.co/FastVideo/Wan2.2-S2V-14B-Diffusers**. It was produced
with the script in this PR (58GB, Apache-2.0, no tensors modified), and the
script's built-in check passed on the real files: all 1260 weights land in
the right place with the right shape. My original conversion remains at
https://huggingface.co/SuhaanCoding/Wan2.2-S2V-14B-Diffusers as a cross-check.
The three weight-gated tests in this PR now run against the official path
with no `WAN_S2V_MODEL_PATH` override needed, and the quickstart below works
as written (verified end to end on GPU — all 32 tests pass, sample video
in the thread).

```python
# this works now:

gen = VideoGenerator.from_pretrained("FastVideo/Wan2.2-S2V-14B-Diffusers")
gen.generate_video(
    prompt="...",
    image_path="face.jpg",
    audio_path="speech.wav",
    output_path="out.mp4",
    save_video=True,
)
```

**New files**

| File | Lines | What it does |
|---|---:|---|
| `fastvideo/models/dits/wan_s2v.py` | 575 | The model itself |
| `fastvideo/models/dits/wan_s2v_audio.py` | 259 | Turns sound into something the model can read |
| `fastvideo/tests/transformers/test_wan_s2v.py` | 580 | Tests (29 run everywhere, 3 more need weights + GPU) |
| `scripts/checkpoint_conversion/wan_s2v_to_diffusers.py` | 195 | Repacks the official download into the loadable layout |
| `fastvideo/pipelines/stages/audio_encoding.py` | 134 | Pipeline step that reads the audio file |
| `fastvideo/configs/models/dits/wan_s2v.py` | 112 | Model size settings, copied from the official download |
| `fastvideo/pipelines/basic/wan/s2v_stages.py` | 92 | Two steps this model does differently (see notes) |
| `fastvideo/pipelines/basic/wan/wan_s2v_pipeline.py` | 75 | Puts the pipeline steps in order |
| `examples/inference/basic/basic_wan_s2v.py` | 44 | Example script |

**Changed files** — all small additions

| File | What changed |
|---|---|
| `fastvideo/pipelines/stages/denoising.py` | Hands this model its sound and picture, and mutes the sound on the "no guidance" pass, the way the official code does |
| `fastvideo/models/loader/component_loader.py` | Loads the speech encoder |
| `fastvideo/configs/pipelines/wan.py` | Settings for this model |
| `fastvideo/pipelines/basic/wan/presets.py` | Default size, frame count, steps (matching the official defaults) |
| `fastvideo/registry.py`, `fastvideo/models/registry.py` | Makes the model name resolve |
| `fastvideo/api/sampling_param.py`, `fastvideo/api/schema.py`, `fastvideo/pipelines/pipeline_batch_info.py` | New `audio_path` option, carried through the pipeline |
| `docs/inference/support_matrix.md` | Adds the model to both tables |
| `docs/design/inference_schema_parity_inventory.yaml` | Lists the new settings, as the contract test requires |
| `fastvideo/tests/api/test_parser.py` | Existing test now expects the new option |
| `pyproject.toml` | One spell-checker exception: the official download has a setting whose value is literally spelled `"padd"`, and it must match |
| `fastvideo/pipelines/stages/__init__.py` | Exports the new step |

### Notes for reviewers

**Why this model gets its own file** instead of reusing the existing Wan model.
Three things are genuinely different:

1. The model works on a mixed list of tokens: the video being made, plus the
   reference picture, plus (for long videos) a few earlier frames. Only the
   video part comes back out.
2. The picture and earlier frames are already clean, so they are treated as
   "time step zero" while the video part gets the real time step. Every
   block in the model handles the two groups separately.
3. Sound is fed in partway through, into 12 of the 40 layers, rather than
   mixed into the input.

**The two custom steps** in `s2v_stages.py`, both copied from the official
runner:

- The reference picture is encoded alone, as a single frame. The existing
  image step builds a padded fake video instead, which is a different format.
- When decoding, the reference frame is put in front so the decoder has
  context, and the warm-up frames are cut off afterwards.

**Other things worth knowing:**

- There is no version of this model in the diffusers library. The pull request
  that would have added one (huggingface/diffusers#12258) was closed without
  merging. So the official Wan2.2 repo is the only reference, and this port
  was checked against it line by line.
- The model needs the **large** wav2vec2 speech encoder (it expects 25 layers
  of sound features; the small one only has 13). The right one is bundled
  inside the official download, and the repack script carries it over.
- One number in the settings is worth flagging: the official code runs with a
  73-frame memory window for long videos, while the model file's own default
  says 17. We follow the official runtime value (73). Both happen to line up
  mathematically, which is why no test can tell them apart — only a real run
  can.
- Making several GPUs share one video is not wired up yet. One GPU works.
  Noted in the support matrix.

## Test Plan

```bash
# style, types, spelling — full repo
pre-commit run --all-files

# unit tests. Not the whole transformers folder: two unrelated test files
# there download multi-GB checkpoints the moment they are imported, which
# hangs any machine without them. That is pre-existing and not from this PR.
pytest fastvideo/tests/api/ fastvideo/tests/transformers/test_wan_s2v.py -q
```

Two extra checks were run locally that are **not part of this PR**, because
they need a clone of the official Wan2.2 repo next to them:

1. A small stand-in model (64 wide, 4 layers — instead of 5120 wide, 40
   layers) was built in both this port and the official code, given identical
   random weights on both sides, and run on the same inputs on CPU. This
   proves the math is the same — every add and multiply in the same order —
   but not the GPU speed path or the low-precision number format used in
   production.
2. One full-size layer (5120 wide) was loaded with **real weights** from the
   official download (700MB fetched, not the whole 28GB) and compared the same
   way.

Happy to attach both scripts if that helps review.

## Test Results

<details>
<summary>pre-commit (full repo)</summary>

```
PyMarkdown...............................................................Passed
mypy.....................................................................Passed
Check for spaces in all filenames........................................Passed
```

</details>

<details>
<summary>unit tests</summary>

```
320 passed, 3 skipped
```

The 3 skips need the downloaded weights and a GPU; they print how to enable
them and will run in CI.

</details>

<details>
<summary>matches the official code — small stand-in model, same weights both sides, 12 input variations</summary>

```
weights: ref 168 | port 168 | missing 0 | unexpected 0
  OK   baseline                 max|delta| = 0.000e+00
  OK   drop_motion_frames       max|delta| = 0.000e+00
  OK   add_last_motion=1        max|delta| = 0.000e+00
  OK   add_last_motion=0        max|delta| = 0.000e+00
  OK   motion_frames=(3,1)      max|delta| = 0.000e+00
  OK   t=0                      max|delta| = 0.000e+00
  OK   t=999                    max|delta| = 0.000e+00
  OK   8 latent frames          max|delta| = 0.000e+00
  OK   non-square 16x24         max|delta| = 0.000e+00
  OK   nonzero cond_states      max|delta| = 0.000e+00
  OK   cond None == ref zeros   max|delta| = 0.000e+00
  OK   motion None == ref drop  max|delta| = 0.000e+00
12/12 cases bit-matched; worst delta 0.000e+00
```

The outputs are identical down to the last digit. The last two rows check
that this port's defaults behave exactly like the official code's explicit
values.

</details>

<details>
<summary>matches the official code — one real full-size layer, real downloaded weights</summary>

```
loaded 27 real tensors for block 0
reference block: missing 0, unexpected 0
port block     : missing 0, unexpected 0
block output (1, 96, 5120)
max |delta| = 0.000e+00  (relative 0.000e+00)
```

</details>

<details>
<summary>every weight in the download has a home</summary>

All 1260 tensors in the official download map onto a model parameter with the
right shape — nothing missing, nothing left over, no shape differences.
Checked without downloading the weights (the file headers carry the shapes)
and re-checked automatically by the repack script before it will let anyone
upload its output.

</details>

Checked:

- The math matches the official code exactly, in 12 settings, including this
  port's default behaviours.
- One real full-size layer with real weights also matches exactly.
- Every weight has a correctly-shaped home.
- The pipeline's hand-off into the model is tested end to end on a small
  model: the exact call the denoising loop makes, through the guidance
  arithmetic, through a real scheduler step.

**checked on a GPU.** I ran everything on a DGX Spark:
all 32 tests pass (including a full forward pass of the loaded model), and a
first real video was generated -- speech sync confirmed by watching it. Details
and the clip are in the comments below. Remaining items from the original list:

https://github.com/user-attachments/assets/58fd3fe7-0545-4110-b1d9-c4dc5a35a8f7

## Checklist

**Aug 28:** rebased onto main (v0.2.1, `c7da2f5d`), no conflicts. All 32 tests pass on the rebased branch on a DGX Spark, including the three weight-gated ones against the official checkpoint.


SuhaanCoding:
Said this in the PR Body: Someone with about 40GB of disk needs to run the script once and upload the result to the FastVideo page on Hugging Face (as was done for GameCraft, MagiHuman, LingBot and others). Until that upload exists, the model cannot be loaded by users. This is the same order the MagiHuman port used: code first, weights upload as the
follow-up.

Checking it on my own HF account right now to verify ~ like 6-12 hours worth of download/upload for extra verification and will confirm it here once done

SuhaanCoding:
Converted weights are now up: https://huggingface.co/SuhaanCoding/Wan2.2-S2V-14B-Diffusers (58GB, made with the script in this PR, no tensors modified — details in the updated description). The script's built-in check passed on the real files: all 1260 weights, correct names and shapes. The three weight-gated tests in this PR can now be run against it by anyone with a GPU: `WAN_S2V_MODEL_PATH` points at a local download of that repo. Happy to mirror it to the FastVideo org or rerun the script if a maintainer prefers.

SuhaanCoding:
I tested it, the port does the exact same math as the official model: identical outputs on a small copy and on one real full-size layer and every one of the 1260 weights from the official download loads into the right place, verified on the real files.

Need a maintainer copying the weights to the project's official page

SuhaanCoding:
**Update: ran the whole thing on a GPU and it works end to end.**

- All 32 tests in `fastvideo/tests/transformers/test_wan_s2v.py` pass on GPU, including loading every weight through the real loader and a full forward pass of the 16.3B model.

-Got a funny video out of it 


https://github.com/user-attachments/assets/dde1594b-3b46-4a44-949d-f44a5fdf1775

fastvideo/pipelines/stages/audio_encoding.py
```diff
@@ -0,0 +1,134 @@
...
+        waveform = self._load_waveform(batch.audio_path)
+        inputs = self.audio_processor(waveform, sampling_rate=WAV2VEC_SAMPLE_RATE, return_tensors="pt")
+
+        with torch.no_grad():
+            outputs = self.audio_encoder(inputs.input_values.to(device), output_hidden_states=True)
+        features = torch.stack(outputs.hidden_states).squeeze(1)  # [num_layers, T, C]
+        features = _resample_features(features, WAV2VEC_FEATURE_RATE, REFERENCE_VIDEO_RATE)
+
+        assert batch.num_frames is not None and batch.fps is not None
+        batch.audio_embeds = self._bucket_to_frames(features, batch.num_frames, batch.fps).unsqueeze(0)
```
H1yori233, line 117: The source audio is used for conditioning but is not propagated to the save path. run produced MP4 with zero audio stream.

fastvideo/pipelines/basic/wan/s2v_stages.py
```diff
@@ -0,0 +1,92 @@
...
+    _WARMUP_FRAMES = 3
+
+    def forward(self, batch: ForwardBatch, fastvideo_args: FastVideoArgs) -> ForwardBatch:
+        assert batch.image_latent is not None, "S2V decode needs the reference latent for temporal context"
+        num_frames = batch.num_frames
+        batch.latents = torch.cat([batch.image_latent.to(batch.latents.device, batch.latents.dtype), batch.latents],
+                                  dim=2)
+        batch = super().forward(batch, fastvideo_args)
+        if fastvideo_args.output_type != "latent" and batch.output is not None and num_frames is not None:
+            batch.output = batch.output[:, :, -num_frames:][:, :, self._WARMUP_FRAMES:]
```
H1yori233, line 91: `num_frames=80` returns 77 frames. It matches official Wan but violates FastVideo’s public frame-count contract. Maybe separate internal `infer_frames` from the final `4k+1` output count?

fastvideo/pipelines/basic/wan/wan_s2v_pipeline.py
```diff
@@ -0,0 +1,75 @@
...
+        self.add_stage(stage_name="image_latent_preparation_stage",
+                       stage=S2VRefImageEncodingStage(vae=self.get_module("vae")))
+
+        self.add_stage(stage_name="denoising_stage",
+                       stage=DenoisingStage(transformer=self.get_module("transformer"),
+                                            scheduler=self.get_module("scheduler")))
+
+        # Official recipe: decode with the reference latent prepended so the
+        # causal VAE has temporal context, then trim the warm-up frames.
+        self.add_stage(stage_name="decoding_stage", stage=S2VDecodingStage(vae=self.get_module("vae")))
```
H1yori233, line 72: I think multi-clip continuation is missing? Official Wan repeats generation for longer audio and carries motion latents between clips. This pipeline performs one pass and silently truncates output to the first 77 frames.

fastvideo/pipelines/stages/audio_encoding.py
SuhaanCoding: fixed this, the audio stage now keeps the source track on batch.extra["audio"] / ["audio_sample_rate"], which is the same convention LTX-2 and MagiHuman use

fastvideo/pipelines/basic/wan/s2v_stages.py
SuhaanCoding: Okayy num_frames is now the 4k+1 count the caller gets back

fastvideo/pipelines/basic/wan/wan_s2v_pipeline.py
SuhaanCoding: Yeah you are right, added the continuation loop from speech2video.py

PR #1691, opened 2026-08-06
[bugfix] profiler region system: record + export actually work, usability roll-up

SolitaryThinker:
The region profiler (`FASTVIDEO_TORCH_PROFILER_DIR` + `FASTVIDEO_TORCH_PROFILE_REGIONS`) recorded nothing and exported nothing. This fixes it and makes it actually usable.

bugs fixed:
- profiler was built with a wait/warmup schedule but nothing ever calls `profiler.step()`, so it never left the wait state and recorded zero events
- collection-enabled flag started False so the initial toggle-off short-circuited and everything before the first region leaked into the trace
- traces only export at `stop()`, which inference never calls — added an atexit hook (idempotent)
- the region marker opened before collection was enabled, so `fastvideo.region::*` never appeared in traces
- `WITH_STACK` declared default False but the env lambda defaulted True; now actually off (~1.5x runtime overhead when on)

usability:
- `FASTVIDEO_TORCH_PROFILE_REGIONS=inference_denoising` works — short names resolve the `profiler_region_` prefix
- typo'd region names warn once instead of silently profiling nothing forever
- `stop()` writes a per-rank `key_averages` summary (txt + json) next to the trace, so operator splits don't require parsing multi-GB chrome traces
- regions also emit nvtx ranges, so the same names show up in nsys timelines
- module-level `profiler_region("name")` context manager (no-op without a profiler); h3 denoising uses it

deleted dead code: `pipeline.profile()` / `pipeline.profiler`, `get/set_global_profiler`, `default_enabled`, the unused `WAIT/WARMUP/ACTIVE_STEPS` env knobs.

CPU contract test proves the whole contract in a subprocess: ops inside regions captured, ops outside excluded, short names resolve, warn-once, atexit export, summary files written.

### DIFF
--- a/examples/training/finetune/wan_t2v_1.3B/crush_smol/profile_finetune_t2v.sh
+++ b/examples/training/finetune/wan_t2v_1.3B/crush_smol/profile_finetune_t2v.sh
@@ -19,16 +19,13 @@ export FASTVIDEO_TORCH_PROFILER_RECORD_SHAPES=1
 export FASTVIDEO_TORCH_PROFILER_WITH_STACK=1  
 export FASTVIDEO_TORCH_PROFILER_WITH_FLOPS=1
 export FASTVIDEO_TORCH_PROFILER_WITH_PROFILE_MEMORY=1
-export FASTVIDEO_TORCH_PROFILER_WAIT_STEPS=2
-export FASTVIDEO_TORCH_PROFILER_WARMUP_STEPS=1
-export FASTVIDEO_TORCH_PROFILER_ACTIVE_STEPS=1
 export FASTVIDEO_TORCH_PROFILER_DIR="../profiler_traces/wan_t2v_finetune/"
 
 # Training arguments
 training_args=(
   --tracker_project_name "wan_t2v_finetune"
   --output_dir "checkpoints/wan_t2v_finetune"
-  --max_train_steps 5000
+  --max_train_steps 1  # Profiler captures every selected region until shutdown.
   --train_batch_size 1
   --train_sp_batch_size 1
   --gradient_accumulation_steps 8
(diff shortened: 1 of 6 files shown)


SolitaryThinker:
Rebased onto latest main to resolve the MiniMax H3 conflict and pick up main’s pre-commit formatting fix; no PR content changes (range-diff clean).

SolitaryThinker (commented):
Verdict: CHANGES_REQUESTED at `3cc2420e841fc1238c8ad21d0b769e8632dd9f87`.

SolitaryThinker’s shared-controller changes address the actual profiler lifecycle failures, and the CPU subprocess contract is a useful end-to-end guard.

- Root cause `[read]`: the wrong behavior originated in `fastvideo/profiler.py`: `get_or_create_profiler()` installed a wait/warmup schedule although no FastVideo controller path called `profiler.step()`; `TorchProfilerController` initialized `_collection_enabled=False` even though `profiler.start()` begins collecting; and inference had no path that called `stop()`. This patch fixes those shared lifecycle points rather than adding caller-specific guards.
- Red proof `[measured]`: `python -m pytest -q fastvideo/tests/contract/test_profiler_regions.py` gives `2 failed` with the production sources restored to base and `2 passed` at this head.
- Sibling sweep `[read/measured]`: all controller callers, decorators, and `train`/`train_one_step` overrides were traced. Two blockers remain:
  1. `pre-commit run --files <changed paths>` still fails because YAPF rewrites `fastvideo/profiler.py:365-367` and `fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_denoising.py:151-152`. Commit the hook output and rerun the check.
  2. `examples/training/finetune/wan_t2v_1.3B/crush_smol/profile_finetune_t2v.sh:22-24` still sets the three step-window variables this PR deletes, while line 31 runs 5,000 steps and `TrainingPipeline.train_one_step()` is profiled on every iteration until the final `stop()`. With stack, memory, and FLOP capture enabled, that becomes an unbounded 5,000-step trace. Remove the obsolete variables and bound the profiling example to a small run, or retain a working bounded-collection mechanism.

SolitaryThinker:
Addressed both review findings in e097050d8: applied the project formatter and bounded the profiling example to one step after removing the dead schedule knobs; pre-commit, bash syntax, and the profiler contract test all pass.

PR #1693, opened 2026-08-07
[bugfix]: condition HunyuanVideo 1.5 i2v on the reference image

### LINKED ISSUE #1692: [Bug] HunyuanVideo 1.5 I2V cannot run: image_embeds is never populated, stage verification fails
### Describe the bug

`HunyuanVideo15ImageToVideoPipeline` cannot generate from a reference image at
all. Passing `image_path` raises `StageVerificationError` before denoising,
because nothing ever encodes the image and `image_embeds` reaches the stage's
own output check as an empty list. Both HY1.5 I2V checkpoints are registered
and selectable, so this is reachable through the normal API.

The SigLIP encoder that would do the work is shipped with the checkpoint but
never loaded. Loading
`hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-480p_i2v_step_distilled`
logs its `model_index.json`:

```
'feature_extractor': ['transformers', 'SiglipImageProcessor'],
'image_encoder':     ['transformers', 'SiglipVisionModel'],
```

FastVideo requests neither.

**1. Neither component is in the required module list**

`fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py:23-25`

```python
_required_config_modules = [
    "text_encoder", "text_encoder_2", "tokenizer", "tokenizer_2", "vae", "transformer", "scheduler"
]
```

**2. The image encoding stage is constructed with no encoder**

`fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py:49-50`

```python
self.add_stage(stage_name="image_encoding_stage",
               stage=Hy15ImageEncodingStage(image_encoder=None, image_processor=None))
```

**3. The stage only writes `image_embeds` on the T2V path**

`fastvideo/pipelines/stages/image_encoding.py:106-116`

```python
def forward(self, batch, fastvideo_args):
    if batch.pil_image is None:
        batch.image_embeds = [torch.zeros(1, 729, 1152, device=get_local_torch_device())]

    raw_latent_shape = list(batch.raw_latent_shape)
    raw_latent_shape[1] = 1
    batch.video_latent = torch.zeros(tuple(raw_latent_shape), device=get_local_torch_device())
    return batch
```

When `pil_image` is present the branch is skipped and `image_embeds` keeps its
`ForwardBatch` default of `[]` (`fastvideo/pipelines/pipeline_batch_info.py:79`).
`video_latent` is the 1 channel conditioning mask and is zeroed unconditionally,
so the mask is 0 on every frame including the first.

**4. The inherited output check then fails**

`fastvideo/pipelines/stages/image_encoding.py:90-94`

```python
def verify_output(self, batch, fastvideo_args) -> VerificationResult:
    result = VerificationResult()
    result.add_check("image_embeds", batch.image_embeds, V.list_of_tensors_dims(3))
    return result
```

`[]` does not satisfy `list_of_tensors_dims(3)`, so the run aborts here.

**5. Even past that point the layout would be T2V**

The 65 channel input is assembled at `denoising.py:405-414` as
`cat([latents, video_latent, v2v_zero_pad], dim=1)`, which is 32 + 1 + 32. With
`video_latent` zeroed and `v2v_zero_pad = torch.zeros_like(latents)`
(`denoising.py:319`), the last 33 channels are all zero, which is the T2V
layout. `denoising.py:156-164` only forwards `encoder_hidden_states_image` when
`len(image_embeds) > 0`, and the transformer decides the branch with
`is_t2v = torch.all(encoder_hidden_states_image == 0)`
(`fastvideo/models/dits/hunyuanvideo15.py:505`). So the reference image has no
route into the model even if the check were relaxed.

**6. The two pipeline files are identical apart from the class name**

```
$ diff fastvideo/pipelines/basic/hunyuan15/hunyuan15_pipeline.py \
       fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py
21c21
< class HunyuanVideo15Pipeline(ComposedPipelineBase):
---
> class HunyuanVideo15ImageToVideoPipeline(ComposedPipelineBase):
59c59
< EntryClass = HunyuanVideo15Pipeline
---
> EntryClass = HunyuanVideo15ImageToVideoPipeline
```

Same stages, same required modules, same order. There is no I2V specific
behavior anywhere in the pipeline.

For reference, `HYWorldImageEncodingStage` in the same file does the work that
is missing here: SigLIP for `image_embeds`, VAE for the conditioning latent,
and a mask channel set to 1 on the first frame. Its docstring describes that
layout as "following the original HunyuanVideo-1.5 implementation", so the
intended contract appears known and simply not wired up for HY1.5.

### Reproduction

```python

gen = VideoGenerator.from_pretrained(
    "hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-480p_i2v_step_distilled",
    num_gpus=1,
    use_fsdp_inference=False,
    dit_cpu_offload=True,
    vae_cpu_offload=True,
    text_encoder_cpu_offload=True,
    pin_cpu_memory=True,
)

gen.generate_video(
    "The scene comes alive with gentle motion, cinematic lighting.",
    image_path="assets/images/bus_terminal.jpg",
    seed=42,
    output_path="i2v_probe",
    save_video=True,
    negative_prompt="",
    num_frames=33,
    fps=16,
)
```

Result:

```
File "fastvideo/pipelines/composed_pipeline_base.py", line 510, in forward
    batch = stage(batch, fastvideo_args)
File "fastvideo/pipelines/stages/base.py", line 176, in __call__
    self._run_verification(output_result, stage_name, "output")
File "fastvideo/pipelines/stages/base.py", line 98, in _run_verification
    raise StageVerificationError(error_msg)
fastvideo.pipelines.stages.base.StageVerificationError: Output verification failed
for image_encoding_stage|Hy15ImageEncodingStage: Failed fields: image_embeds
Details: Validation failures:
  Field 'image_embeds':
    1. Validator 'validator' failed. Actual: []
```

The same call without `image_path` runs to completion, which is consistent with
the T2V path being the only one implemented.

A cheaper check that needs no generation: the loaded module list never mentions
`image_encoder` or `feature_extractor` even though `model_index.json` declares
both.

### Environment

DGX Spark (GB10, aarch64), CUDA 13.0, torch 2.12, single GPU, FastVideo at
99d04a7f. The defect is in pipeline wiring rather than anything device specific.

### Why this went unnoticed

Not a criticism of any PR, just context for where coverage is missing.
`hunyuan15_i2v_pipeline.py` was added in #1046, whose subject is the 1080p
super resolution pipeline. That PR shipped no I2V example and no I2V test, and
`examples/inference/basic/` still has only the two T2V scripts, `basic_hy15.py`
and `basic_hy15_1080p.py`. The original model PR #943 added encoder and VAE unit
tests but nothing that exercises a pipeline end to end, and there is no HY1.5
entry under `fastvideo/tests/ssim/`. No automated or manual path has run HY1.5
I2V with an actual image.

### Suggested fix

Add `image_encoder` and `feature_extractor` to the I2V pipeline's
`_required_config_modules`, pass them into `Hy15ImageEncodingStage`, and give
that stage the I2V path it lacks: SigLIP encode into `image_embeds`, VAE encode
the reference image into the conditioning latent, and set the mask channel to 1
on the first frame and 0 elsewhere.

Worth noting for whatever test comes with the fix: an SSIM reference alone would
not have caught this once the crash is gone, since a T2V fallback still produces
a plausible video. Asserting that two different reference images yield different
output would catch it directly.

Happy to send a PR if that direction looks right. I hit this while establishing
what the inference side contract is, ahead of adding HY1.5 I2V training on top
of #1662.

KyleNeverGivesUp:
## Purpose

`HunyuanVideo15ImageToVideoPipeline` cannot generate from a reference image.
Passing `image_path` raises `StageVerificationError` before denoising, and even
past that point the image has no route into the transformer.

Fixes #1692

## Changes

The i2v checkpoints ship a SigLIP pair that FastVideo never asked for. Their
`model_index.json` declares:

```
'feature_extractor': ['transformers', 'SiglipImageProcessor'],
'image_encoder':     ['transformers', 'SiglipVisionModel'],
```

Neither was in `_required_config_modules`, so `Hy15ImageEncodingStage` was
constructed with `image_encoder=None`. The stage only wrote `image_embeds` on
the `pil_image is None` branch, so supplying an image left it at its
`ForwardBatch` default of `[]` and the inherited output check rejected it.

Three changes:

1. `Hunyuan15I2V480PStepDistilledConfig` and `Hunyuan15I2V720PConfig` declare
   `image_encoder_config = SiglipVisionConfig`, which is what makes the loader
   build one. FastVideo already has the SigLIP implementation
   (`fastvideo/models/encoders/siglip.py`, registered at
   `fastvideo/models/registry.py:106`), so no new model code.
2. The i2v pipeline adds `image_encoder` and `feature_extractor` to its required
   modules and passes them, plus the VAE, into the stage.
3. `Hy15ImageEncodingStage` gains the i2v path: SigLIP hidden states into
   `image_embeds`, VAE latents of the reference image into the first frame of
   the conditioning stream, and a mask channel set to 1 on that frame.

### Channel order

This is the part that is invisible until an image is actually supplied.
`DenoisingStage` appends `video_latent` and a zero pad separately
(`denoising.py:405-414`), producing

```
[latents 32][mask 1][conditioning 32]
```

but the model expects

```
[latents 32][conditioning 32][mask 1]
```

For t2v both trailing blocks are zero, so the two orders yield an identical
tensor and the discrepancy never surfaces. The i2v path therefore populates
`image_latent` as one 33 channel block and leaves `video_latent` unset, so
denoising takes the `elif batch.image_latent is not None` branch that appends it
whole.

The expected layout is not a guess. The diffusers reference implementation,
`diffusers/pipelines/hunyuan_video1_5/pipeline_hunyuan_video1_5_image2video.py`,
builds `cat([latents, cond_latents_concat, mask_concat], dim=1)`, applies
`scaling_factor` to the image latents, repeats them over the temporal dimension
and zeroes every frame after the first, and sets the mask to 1 on frame 0.
`HYWorldImageEncodingStage` in this repo already assembles `image_latent` the
same way.

### Blast radius

t2v and the two super-resolution pipelines share this stage but never carry a
`pil_image`, so they keep the existing branch untouched, including
`video_latent`, which `SRDenoisingStage` reads directly
(`sr_denoising.py:168`). A pipeline built without the encoder that is handed an
image now raises a message naming the missing pieces rather than failing an
opaque field check.

## Test Plan

```bash
pytest fastvideo/tests/stages/test_hy15_image_encoding_stage.py -q
pre-commit run --files fastvideo/pipelines/stages/image_encoding.py \
                       fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py \
                       fastvideo/configs/pipelines/hunyuan15.py \
                       fastvideo/tests/stages/test_hy15_image_encoding_stage.py
```

CPU only, no checkpoint needed. The tests fake SigLIP and the VAE, since what
needs guarding is the layout rather than the numerics.

Covered:

- t2v still emits all zero `image_embeds`, which is what keeps the transformer's
  `is_t2v` check true, and still populates `video_latent` for the SR stages
- i2v `image_embeds` are non-zero, which is what turns that check off
- conditioning occupies channels 0 to 31 and the mask channel 32, with the mask
  on frame 0 only and the conditioning zero after frame 0
- `scaling_factor` is applied
- `video_latent` stays unset on the i2v path, so denoising cannot take the
  branch that would swap the blocks
- a pipeline without the encoder raises a clear error

## Test Results

<details>

```
$ pytest fastvideo/tests/stages/test_hy15_image_encoding_stage.py -q
.......
7 passed, 14 warnings in 0.05s

$ pre-commit run --files ...
```

</details>

Before, from #1692:

```
fastvideo.pipelines.stages.base.StageVerificationError: Output verification failed
for image_encoding_stage|Hy15ImageEncodingStage: Failed fields: image_embeds
Details: Validation failures:
  Field 'image_embeds':
    1. Validator 'validator' failed. Actual: []
```

## Checklist

Memory: the i2v pipelines now load a SigLIP vision tower they previously did
not. It honours `image_encoder_cpu_offload` and is moved back to CPU after the
encode, same as the other image encoding stages.

**Not verified here: an end to end i2v run on a GPU.** I am doing that next and
will post before and after in #1692. Marking this ready rather than draft since
the change is self contained and the layout question is settled against the
diffusers reference, but happy to hold it until that lands if you would rather
see the run first.

There is also no i2v example script for HY1.5 and no HY1.5 entry under
`fastvideo/tests/ssim/`, which is why this went unnoticed. Worth noting that an
SSIM reference alone would not catch a regression here once the crash is gone,
since a t2v fallback still produces a plausible video. Happy to add an example
script in a follow up.

### DIFF
--- a/fastvideo/configs/pipelines/hunyuan15.py
+++ b/fastvideo/configs/pipelines/hunyuan15.py
@@ -112,7 +112,7 @@ class Hunyuan15T2V480PConfig(PipelineConfig):
 
     vae_tiling: bool = True
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         self.vae_config.load_encoder = False
         self.vae_config.load_decoder = True
         if self.text_encoder_configs:
@@ -127,6 +127,15 @@ class Hunyuan15I2V480PStepDistilledConfig(Hunyuan15T2V480PConfig):
     # The i2v checkpoints ship a SigLIP vision tower; declaring it here is what
     # makes the loader build one.
     image_encoder_config: EncoderConfig = field(default_factory=SiglipVisionConfig)
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.vae_config.load_encoder = True
+
+    def check_pipeline_config(self) -> None:
+        super().check_pipeline_config()
+        if not self.vae_config.load_encoder:
+            raise ValueError("HunyuanVideo 1.5 I2V requires the VAE encoder.")
 
 
 @dataclass
@@ -145,6 +154,15 @@ class Hunyuan15I2V720PConfig(Hunyuan15T2V720PConfig):
     flow_shift: int = 7
 
     image_encoder_config: EncoderConfig = field(default_factory=SiglipVisionConfig)
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.vae_config.load_encoder = True
+
+    def check_pipeline_config(self) -> None:
+        super().check_pipeline_config()
+        if not self.vae_config.load_encoder:
+            raise ValueError("HunyuanVideo 1.5 I2V requires the VAE encoder.")
 
 
 @dataclass
--- a/fastvideo/configs/pipelines/hunyuan15.py
+++ b/fastvideo/configs/pipelines/hunyuan15.py
@@ -8,7 +8,7 @@ import torch
 
 from fastvideo.configs.models import DiTConfig, EncoderConfig, VAEConfig
 from fastvideo.configs.models.dits import HunyuanVideo15Config
-from fastvideo.configs.models.encoders import (BaseEncoderOutput, Qwen2_5_VLConfig, T5Config)
+from fastvideo.configs.models.encoders import (BaseEncoderOutput, Qwen2_5_VLConfig, SiglipVisionConfig, T5Config)
 from fastvideo.configs.models.vaes import Hunyuan15VAEConfig
 from fastvideo.configs.models.upsamplers import SRTo720pUpsamplerConfig, SRTo1080pUpsamplerConfig
 from fastvideo.configs.pipelines.base import PipelineConfig, UpsamplerConfig
@@ -124,6 +124,10 @@ class Hunyuan15T2V480PConfig(PipelineConfig):
 class Hunyuan15I2V480PStepDistilledConfig(Hunyuan15T2V480PConfig):
     flow_shift: int = 7
 
+    # The i2v checkpoints ship a SigLIP vision tower; declaring it here is what
+    # makes the loader build one.
+    image_encoder_config: EncoderConfig = field(default_factory=SiglipVisionConfig)
+
 
 @dataclass
 class Hunyuan15T2V720PConfig(Hunyuan15T2V480PConfig):
@@ -139,6 +143,8 @@ class Hunyuan15I2V720PConfig(Hunyuan15T2V720PConfig):
 
     # HunyuanConfig-specific parameters with defaults
     flow_shift: int = 7
+
+    image_encoder_config: EncoderConfig = field(default_factory=SiglipVisionConfig)
 
 
 @dataclass
--- a/fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py
+++ b/fastvideo/pipelines/basic/hunyuan15/hunyuan15_i2v_pipeline.py
@@ -20,8 +20,12 @@ logger = init_logger(__name__)
 
 class HunyuanVideo15ImageToVideoPipeline(ComposedPipelineBase):
 
+    # image_encoder and feature_extractor are the SigLIP pair the i2v
+    # checkpoints ship; without them the reference image has no route into the
+    # transformer and the run is text-to-video wearing an i2v label.
     _required_config_modules = [
-        "text_encoder", "text_encoder_2", "tokenizer", "tokenizer_2", "vae", "transformer", "scheduler"
+        "text_encoder", "text_encoder_2", "tokenizer", "tokenizer_2", "vae", "transformer", "scheduler",
+        "image_encoder", "feature_extractor"
     ]
 
     def create_pipeline_stages(self, fastvideo_args: FastVideoArgs):
@@ -47,7 +51,9 @@ class HunyuanVideo15ImageToVideoPipeline(ComposedPipelineBase):
                                                     transformer=self.get_module("transformer")))
 
         self.add_stage(stage_name="image_encoding_stage",
-                       stage=Hy15ImageEncodingStage(image_encoder=None, image_processor=None))
+                       stage=Hy15ImageEncodingStage(image_encoder=self.get_module("image_encoder"),
+                                                    image_processor=self.get_module("feature_extractor"),
+                                                    vae=self.get_module("vae")))
 
         self.add_stage(stage_name="denoising_stage",
                        stage=DenoisingStage(transformer=self.get_module("transformer"),
--- a/fastvideo/pipelines/stages/image_encoding.py
+++ b/fastvideo/pipelines/stages/image_encoding.py
@@ -119,9 +119,10 @@ class Hy15ImageEncodingStage(ImageEncodingStage):
         Encode the prompt into image encoder hidden states.
         """
         device = get_local_torch_device()
+        batch_size = batch.raw_latent_shape[0]
 
         if batch.pil_image is None:
-            batch.image_embeds = [torch.zeros(1, 729, 1152, device=device)]
+            batch.image_embeds = [torch.zeros(batch_size, 729, 1152, device=device)]
             raw_latent_shape = list(batch.raw_latent_shape)
             raw_latent_shape[1] = 1
             batch.video_latent = torch.zeros(tuple(raw_latent_shape), device=device)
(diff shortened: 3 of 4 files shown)


KyleNeverGivesUp:
The failing pre-commit check here is not from this PR. `yapf --all-files` currently reformats `fastvideo/utils.py` and `fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_denoising.py` on main, neither of which this branch touches, so the job fails on anything opened against current main.

Fixed separately in #1700 to keep this diff clean. Once that lands I will rebase and the check should go green.

KyleNeverGivesUp:
@SolitaryThinker rebased onto main now that #1700 and #1702 landed, pre-commit
is green. This fixes #1692: HY1.5 i2v raises StageVerificationError on any run
that passes an image.

KyleNeverGivesUp:
Rebased onto current main, now zero commits behind. The earlier failure was
Build #4856 from Aug 14, on a base that was 23 commits stale.

Ran the full `unit_test` lane on a DGX Spark against this branch: 880 passed,
8 failed. All eight also fail on clean `origin/main` on the same machine with an
identical list, so none come from this PR. They are environment dependent: three
in `test_attn_qat_infer_capability_gate` gate on sm90, sm100 and consumer
Blackwell while this box is sm_121a; two in `test_modal_fa4_policy` need a Modal
environment; the remaining three reproduce on main unchanged.

PR #1694, opened 2026-08-07
[feat] Add Ring Attention support

klhhhhh:
## Motivation

Video diffusion models process long spatiotemporal token sequences, making attention increasingly memory- and communication-intensive as video resolution and duration grow. Existing Ulysses sequence parallelism distributes sequence tokens across GPUs by exchanging sequence and attention-head dimensions, but its parallel degree is constrained by the number of attention heads.

This PR adds Ring Attention as an additional sequence-parallel strategy in FastVideo. Ring Attention keeps queries local while circulating sharded keys and values across a Ring process group, allowing every query shard to attend to the complete global sequence without materializing the full sequence on each GPU.

The PR supports both:

- **Pure Ring Attention**, where the entire sequence-parallel group forms one Ring group (`ring_size == sp_size`).
- **Unified Sequence Parallelism (USP)**, which combines Ulysses and Ring Attention in a two-dimensional process mesh:

  ```text
  sp_size = ulysses_size × ring_size
  ```

USP enables more flexible sequence-parallel configurations and avoids requiring the full sequence-parallel degree to be provided exclusively by either Ulysses or Ring Attention.

The implementation integrates Ring Attention with FastVideo's distributed process-group management, attention layer, and local RoPE handling. It also adds tests covering configuration validation, RoPE slicing, Ring subgroup and Ulysses subgroup construction, the single-GPU Ring fallback, blockwise log-sum-exp merging, and two-GPU pure-Ring numerical parity against full FlashAttention.

The current numerical parity test validates the pure-Ring kernel. The hybrid USP path is currently covered by process-group layout tests and multi-GPU integration runs.


klhhhhh:
# Run ring-atten and usp algorithm successfully.

## 4-GPU Ring Attention Inference Test

This PR was manually validated with four GPUs using the following sequence-parallel configuration:

- `num_gpus=4`
- `sp_size=4`
- `ring_size=2`
- Attention backend: `FLASH_ATTN`
- FSDP inference: disabled
- DiT/VAE CPU offloading: disabled
- Text encoder CPU offloading: enabled

Because `ring_size=2` is smaller than `sp_size=4`, this configuration exercises the hybrid USP/Ring Attention path rather than pure Ring Attention.

### Inference example

`examples/inference/basic/basic.py` was configured as follows:

```python

# from fastvideo.api.sampling_param import SamplingParam

OUTPUT_PATH = "video_samples"


def main():
    # FastVideo will automatically use the optimal default arguments for the
    # model.
    # If a local path is provided, FastVideo will make a best-effort attempt
    # to identify the optimal arguments.
    generator = VideoGenerator.from_pretrained(
        "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
        # FastVideo will automatically handle distributed setup.
        num_gpus=4,
        sp_size=4,
        ring_size=2,
        use_fsdp_inference=False,
        dit_cpu_offload=False,
        vae_cpu_offload=False,
        text_encoder_cpu_offload=True,
        # Set to False when CPU RAM is limited or an obscure
        # "CUDA error: Invalid argument" is encountered.
        pin_cpu_memory=True,
        # image_encoder_cpu_offload=False,
    )

    # sampling_param = SamplingParam.from_pretrained(
    #     "Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
    # )
    # sampling_param.num_frames = 45
    # sampling_param.image_path = (
    #     "https://huggingface.co/datasets/huggingface/"
    #     "documentation-images/resolve/main/diffusers/astronaut.jpg"
    # )

    # Generate a video using the same API regardless of the GPU count.
    prompt = (
        "A curious raccoon peers through a vibrant field of yellow sunflowers, "
        "its eyes wide with interest. The playful yet serene atmosphere is "
        "complemented by soft natural light filtering through the petals. "
        "Mid-shot, warm and cheerful tones."
    )
    generator.generate_video(
        prompt,
        output_path=OUTPUT_PATH,
        save_video=True,
    )

    # Generate another video without reloading the model.
    prompt2 = (
        "A majestic lion strides across the golden savanna, its powerful frame "
        "glistening under the warm afternoon sun. The tall grass ripples gently "
        "in the breeze, enhancing the lion's commanding presence. The tone is "
        "vibrant, embodying the raw energy of the wild. Low angle, steady "
        "tracking shot, cinematic."
    )
    generator.generate_video(
        prompt2,
        output_path=OUTPUT_PATH,
        save_video=True,
    )


if __name__ == "__main__":
    main()
```

### Running script

The following script was used to run the four-GPU inference test:

```bash
#!/usr/bin/env bash
set -euo pipefail

cd /FastVideo

export CUDA_VISIBLE_DEVICES=0,1,2,3
export FASTVIDEO_ATTENTION_BACKEND=FLASH_ATTN
export NCCL_DEBUG=INFO
export PYTHONUNBUFFERED=1
export NCCL_CUMEM_ENABLE=0
export NCCL_IB_DISABLE=1
export NCCL_P2P_DISABLE=1
export TORCH_NCCL_ASYNC_ERROR_HANDLING=1

echo "===== GPU information ====="
nvidia-smi \
    --query-gpu=index,name,memory.total \
    --format=csv

echo "===== FastVideo source ====="
python -c 'import fastvideo; print(fastvideo.__file__)'

echo "===== FlashAttention ====="
python -c 'import flash_attn; print("flash_attn:", flash_attn.__version__)'

echo "===== Start 4-GPU USP/Ring Attention inference ====="
echo "num_gpus=4, sp_size=4, ring_size=2"

mkdir -p video_samples logs

python examples/inference/basic/basic.py \
    2>&1 | tee logs/ring_4gpu_pipeline.log

echo "===== Finished ====="
echo "Videos: /FastVideo/video_samples"
echo "Log: /FastVideo/logs/ring_4gpu_pipeline.log"
```

### Run command

```bash
chmod +x run_ring_4gpu.sh
./run_ring_4gpu.sh
```

### Outputs

Generated videos:

```text
/FastVideo/video_samples
```

Execution log:

```text
/FastVideo/logs/ring_4gpu_pipeline.log
```

klhhhhh:
And the running log is here.

[ring_4gpu_pipeline.log](https://github.com/user-attachments/files/31305970/ring_4gpu_pipeline.log)

Davids048 (changes requested):
fastvideo/attention/ring/ I see that this folder is basically a direct copy of the repo files at https://github.com/feifeibear/long-context-attention/tree/main/yunchang/kernels

I don't think this is helpful because 
1. This mix the role of FastVideo and the Kernel package, making FastVideo hard to maintain. 
2. I don't see a clear justification of why we need all the files just to implement ring attention. E.g. are we actually going to support ring attn on NPU?

klhhhhh:
@Davids048 Thanks for pointing this out. I agree that vendoring the whole kernel package makes the boundary between FastVideo and the kernel implementation less clear.

The main reason I initially kept the upstream structure was to preserve the existing Ring Attention functionality with minimal modification, and potentially leave room for other kernel backends in the future. For example, FastVideo’s roadmap also includes support for additional attention kernels such as FlashInfer, so I was not sure whether some of these implementations might become useful later.

That said, for the current PR, the actual Ring Attention path only relies on the FlashAttention-related implementation, mainly the functionality used by ring_flash_attn_func. The other kernels are not required for the current integration.

I think it would be cleaner to trim the vendored code to only the components needed for the current Ring Attention implementation, and add other kernels back later if there is a concrete use case. This should keep the FastVideo side easier to maintain while preserving the current functionality.

Davids048:
https://github.com/hao-ai-lab/FastVideo/pull/1694#issuecomment-5369555270

@klhhhhh Could you clarify how this test verify the correctness of the correctness of the implementation? E.g. does it compare the output with that generated with naive 1 GPU regular Flash attention? 

---

> Because ring_size=2 is smaller than sp_size=4, this configuration exercises the hybrid USP/Ring Attention path rather than pure Ring Attention.

I think to verify the correctness we should have a baseline with just pure ring attention. Also, it said "this configuration exercises the hybrid USP/Ring Attention" but in the PR description it mentioned that Unified Sequence Parallelism (USP) is not implemented yet. Are you mixing it up with Ulysses?

klhhhhh:
@Davids048 
> Thanks for pointing this out. I have implemented both pure Ring Attention and the hybrid USP path (Ulysses + Ring Attention). The PR description is outdated and still reflects an earlier stage when USP had not yet been implemented; I will update it accordingly.
>
> I agree that pure Ring Attention should be used as the baseline for correctness verification. I will make the distinction between the two configurations explicit:
>
> - Pure Ring: `sp_size == ring_size`
> - Hybrid USP: `sp_size > ring_size`, combining Ulysses and Ring Attention
>
> So the 4-GPU configuration with `sp_size=4` and `ring_size=2` is the hybrid USP path, not pure Ring. I will also include a separate pure-Ring configuration in the validation results.

klhhhhh:
> [#1694 (comment)](https://github.com/hao-ai-lab/FastVideo/pull/1694#issuecomment-5369555270)
> 
> @klhhhhh Could you clarify how this test verify the correctness of the correctness of the implementation? E.g. does it compare the output with that generated with naive 1 GPU regular Flash attention?
> 
> > Because ring_size=2 is smaller than sp_size=4, this configuration exercises the hybrid USP/Ring Attention path rather than pure Ring Attention.
> 
> I think to verify the correctness we should have a baseline with just pure ring attention. Also, it said "this configuration exercises the hybrid USP/Ring Attention" but in the PR description it mentioned that Unified Sequence Parallelism (USP) is not implemented yet. Are you mixing it up with Ulysses?

The distributed correctness test creates deterministic global Q/K/V tensors, shards them along the sequence dimension across two ranks, runs the actual Ring Attention kernel on each local shard, gathers the local outputs, and compares the reconstructed global output against regular FlashAttention computed from the unsharded global Q/K/V on rank 0. The comparison uses BF16 inputs with rtol=3e-2 and atol=3e-2.
I also separately test the blockwise log-sum-exp merge against full FlashAttention. However, the current numerical parity test covers the two-GPU pure-Ring path only. The existing USP test validates the 2D process-group layout, while the four-GPU sp_size=4, ring_size=2 run is currently an integration test rather than a numerical parity test. I will clarify this distinction in the PR description.

klhhhhh:
@Davids048 , already delete some files that were not used. Can you help me review it again? Thanks.

alexzms (changes requested):
Thanks for this — and for the care that went into the FastVideo-side integration. A few things I want to call out as genuinely well done, because they made this reviewable: the hot path costs one boolean (the existing Ulysses body moved into `_forward_ulysses_attention` unchanged); the USP rank layout is documented with its rationale rather than just asserted; `_slice_local_rope`'s docstring explains *why* Ulysses can apply RoPE unsliced and Ring cannot; and the comment on why the ring path bypasses `self.attn_impl.forward` answered my question before I could ask it. I traced the rank arithmetic in `_build_usp_subgroup_ranks` and the degenerate-case aliasing in `destroy_model_parallel` — both correct.

The requests below are mostly about scope and about the vendored half.

### Blocking

**1. The description and the code disagree about what ships.** The summary says pure Ring only, `ring_size` must equal `sp_size`, and "Hybrid Ulysses–Ring execution is not supported in this PR". But `_check_ring_attention_args` explicitly accepts `1 < ring_size < sp_size`, `initialize_model_parallel` builds the ring×ulysses mesh, `layer.py:361/378/403` runs the Ulysses all-to-all around the ring step, and the `--ring-size` help text documents the hybrid as supported. Meanwhile `tests/distributed/test_ring_attention.py` is driven by `WORLD_SIZE` and contains no reference to `ring_size` or `ulysses`, so the hybrid has no numerical coverage — `test_usp_group_layout.py` only checks rank arithmetic as a pure function and `test_ring_attention_config.py` only checks the config is accepted. **What's the intent here — did the hybrid land after the description was written, or is it deliberately staged?** Either gate `1 < ring_size < sp_size` off to match the description, or add a parity test that runs it.

**2. Two unrelated changes to argument validation.** `check_fastvideo_args` on main asserts `hsdp_replicate_dim <= num_gpus and num_gpus % hsdp_replicate_dim == 0` and the same for `hsdp_shard_dim` (fastvideo_args.py:896-897). Both are gone here, unmentioned. That's validation loss with no connection to Ring Attention. Removing the `if num_gpus < max(tp_size, sp_size): num_gpus = max(...)` fallback (line 899) may well be an improvement, but it turns a silent auto-bump into a hard error for anyone setting `sp_size` without `num_gpus` — worth its own PR with a note. **Was the removal intentional, or lost in a rebase?**

**3. `capabilities.py` mutates the environment at import time.** Lines 42-43 set `os.environ["TORCH_CUDA_ARCH_LIST"]` as a side effect of importing the module, conditional on flashinfer being installed, which changes the compile target for anything built later in the process.

**4. Silent precision loss.** In `_forward_ring_attention`, fp32 q/k/v are cast to bf16 and the output cast back. Every other unsupported condition in that same function raises loudly (training, causal, replicated tokens, padding) — this one quietly degrades numerics. Please raise or at minimum warn, for consistency.

### Scope — the vendored half

This is my main concern, and it's a design question more than a defect list.

`kernels/__init__.py` brings in `AttnType` with 15 members and a `select_flash_attn_impl()` dispatcher; `capabilities.py` probes 7 optional packages at import. FastVideo already has `AttentionBackendEnum`, `attention/selector.py`, `attention/backends/` and `platforms/` doing exactly this. So the package now carries two parallel backend-dispatch systems — and because `DistributedAttention.__init__` requires `backend == FLASH_ATTN`, exactly one of those 15 paths is reachable. The file header says it "keeps only the dispatch paths used by the vendored Ring Attention implementations", but all 15 are there.

Similarly, `ring/__init__.py` states the zigzag / striped / varlen / NPU / FlashInfer / PyTorch variants are intentionally not imported — that's 2,280 lines that cannot execute, some of which still depend on `yunchang`, flashinfer or torch_npu.

**What's the plan for these?** If the intent is to enable the other variants in follow-ups, I'd rather they arrive with their integration than sit here unreachable. My preference would be to ship only the path this PR validates (FA, non-causal, even shards) and drop the rest.

There's a real obstacle I want to acknowledge on the alternative: ring's online merge needs `(out, softmax_lse)` per block, and `AttentionImpl.forward` only returns the output. So "just call the resolved backend" isn't free — it needs the backend interface to optionally return LSE. That seems like the right long-term direction (it's what would ever let Ring compose with FA3 or SDPA instead of being welded to FA2), but a minimal FA-only shim would also be a fine intermediate. A 15-way dispatcher is the option I'd push back on.

### Minor

- `kernels/__init__.py` dispatches `AttnType.SAGE_FP16_TRITON` at both line 205 and line 244. Python takes the first, so the second is unreachable — and the two differ (the dead one passes `pv_accum_dtype="fp32"` and `quantization_backend="cuda"`). Upstream's bug, but it ships as ours.
- `ring_flash_attn.py:11-13` points the reader at a local `_fa_kernels` shim and "`_fa_kernels.py` for details"; no such file exists here — the shim is `kernels/`.
- Raw `print()` at `capabilities.py:43` and `ring_flash_attn.py:50` (the latter once per process on the first ring forward). Repo convention is `init_logger(__name__)`.
- The `minimax_h3_denoising.py` hunk only splits one `from fastvideo.profiler import ...` into two lines with no functional change — worth dropping, since #1739 is also touching MiniMax-H3.
- `get_ring_group()` returns a raw `ProcessGroup` while `get_ulysses_group()` returns a `GroupCoordinator`. Asymmetric for two sibling accessors.
- None of the new tests run in CI: `fastvideo/tests/distributed/` is in no lane, and the unit lane names only `test_sdpa_metadata_mask_contract.py` from `fastvideo/tests/attention/`. #1658 would collect the attention directory; the distributed tests would still need a home.

alexzms (commented):
Thanks — and nice catches in the latest round, several of which I'd missed. The `type(self).forward is not DistributedAttention.forward` guard is the important one: `DistributedAttention_VSA` and `LTXDistributedAttention` override `forward()` wholesale, so those models would have silently kept running plain Ulysses while the user believed Ring was on. Skipping the `cat`/`chunk` when there's no Ulysses subgroup and the `flash_attn` version tuple parse are both real fixes too.

On the HSDP validation — thanks for restoring it and for saying in the comment that dropping the `num_gpus` auto-bump was deliberate; that answers my question. One note for the record, since I looked into it afterwards: `fsdp_load.py` computes `world_size = hsdp_replicate_dim * hsdp_shard_dim` and hands both to `init_device_mesh(mesh_shape=(replicate, shard))`, which requires the product to equal the real world size — strictly stronger than the divisibility checks, since `replicate * shard == num_gpus` already implies `num_gpus % replicate == 0`. So these are belt-and-braces rather than load-bearing. Keeping them is still the right call — failing at argument-parse time with your message beats failing inside mesh construction after the model has started loading — I just don't want you to think a correctness hole was riding on them.

**I ran this on hardware and the implementation checks out.** 4x GB200, torch 2.9.1+cu128, NCCL 2.28.9, flash-attn 2.8.3, at `b6007b30`:

- `tests/attention/test_ring_attention_{config,rope}.py` — 20 passed in 0.61s
- driving the workers directly, `--nproc_per_node=2 --ring-worker` and `--nproc_per_node=4 --hybrid-usp-worker` both complete cleanly
- the full ring suite under pytest passes, including `test_multi_gpu_hybrid_usp_matches_full_attention`

So pure Ring at 2 ranks and Ring x Ulysses at 4 ranks both match full attention numerically. That's the part I most wanted to confirm.

Two problems left, both in the test infrastructure rather than the implementation.

**1. The harness deadlocks itself.** `_run_torchrun` (test_ring_attention.py:659-661) sets `TORCH_DISTRIBUTED_DEBUG` to `DETAIL` whenever the caller hasn't. With that value both multi-GPU tests hang and die on the 120s timeout — and the hang is in the test's own `minimal raw P2P` precheck, a bare `dist.send`/`dist.recv` pair, before Ring Attention runs at all. Both ranks print `submitting raw P2P` and neither reaches `minimal raw P2P test passed`. Bisecting the two variables the helper forces:

| env | result |
|---|---|
| `NCCL_DEBUG=INFO` alone | exit 0, passes |
| `TORCH_DISTRIBUTED_DEBUG=DETAIL` alone | hangs, killed at the timeout |

Presetting `TORCH_DISTRIBUTED_DEBUG=OFF` is what makes everything above pass. I've only reproduced it on this configuration so it may be specific to NCCL 2.28 / sm_100, but a debug default that hangs its own tests is worth making opt-in regardless — perhaps behind `FASTVIDEO_RING_TEST_DEBUG=1`, next to the socket-fallback block you already have commented out just above.

**2. The multi-GPU tests leave state behind.** Running the whole `fastvideo/tests/distributed/` directory on this branch gives `1 failed, 48 passed, 16 skipped`, and the failure is `test_ulysses_fault_injection.py::test_rank_local_setup_failure_falls_back_group_wide[4-0-post_dev_comm]` — not one of yours. Run that file on its own and it's `4 passed, 16 skipped` on this branch *and* on main, so it isn't a regression in the Ring code — it's cross-test pollution from the ring workers. Worth a teardown that leaves the process-group state clean, since any lane that collects the directory would hit this.

Those two together also answer the CI-lane question from last round: wiring these in before the deadlock is fixed would just make the lane red. Once it is, `run_training_tests` already runs on `gpu="L40S:4"` and there are several `L40S:2` lanes, so the 4-GPU hybrid test has a home; the two CPU files fit the unit lane (#1658 would collect all of `tests/attention/` if it lands first).

Fix those two and I'll approve — the implementation itself I'm satisfied with.

Davids048:
I do not think we should merge this yet. Ring support has been threaded through several existing layers without first deciding which layer owns it.

One example is `attention/layer.py`. DistributedAttention now reads the Ring topology, performs Ulysses communication, slices RoPE, validates Ring-specific restrictions, and bypasses the configured backend to call the Ring FlashAttention kernel directly and etc. That is too much policy for the attention layer.

`parallel_state.py` has a similar problem. It was already responsible for general process-group state. This PR also makes it responsible for constructing the USP topology and representing the special Ring and Ulysses cases through global variables and aliased groups. The file is becoming both a process-group registry and a sequence-parallel policy module.

I think we should pause this PR, extract an ownership-boundary refactor based on the requirements that this implementation exposed, and then apply Ring Attention through that boundary.

klhhhhh:
@Davids048 Hi David, I will refactor it.

klhhhhh:
@Davids048 @alexzms , Hi David and alex, I already refactored it, please check.

PR #1698, opened 2026-08-07
[refactor] H3 pipeline cleanup: shared helpers, dead machinery, loop-invariant hoists

SolitaryThinker:
Output of a 4-angle review pass (reuse / simplification / efficiency / altitude) over the H3 inference pipeline. 9 files, +121/−185.

Perf (eager denoising step, all bitwise-neutral):
- packing used out-of-place `index_copy` — three full-buffer clones of the packed `[1, S, 5376]` states per step; now `new_empty` + `index_copy_` (the three index sets partition the sequence, verified).
- RoPE cos/sin rebuilt every step and re-cast fp32→bf16 in every block (200 casts/step); text branch (context_embedder + 2 refiner blocks + 4 collectives) rerun every step on the constant prompt. Both are position/prompt-only, so they're cached per generation — identity-keyed on the source tensor (held, so the check is safe), bypassed under grad and compile.
- Together ~3–7 ms/step eager; reference decode also stops at the trimmed maximum instead of materializing arbitrarily long sources.

Dedup/dead code:
- `MiniMaxH3TimestepEmbedding` was byte-identical to `layers.mlp.MLP(act_type="silu")` (same submodule names → param mapping/fp32 rules unaffected); deleted.
- `build_packed_sequence` now calls the `_frame_position_grid`/`_fill_audio_positions` helpers the Ref2VA packer already extracted (byte-equivalent, incl. the `num_audio_rows - num_audio_latents == num_audio_latents` identity).
- Latent prep's three hand-rolled `(latents - mean) / std` sites use the VAEs' `normalize_latents` (fp16 round-trip preserved); one shared keyframe encoder for fl2va + reference paths.
- Dead getattr waterfalls (conditioning/input prep) for attributes the components expose as typed properties; unreachable empty-rows branch in packing; save/restore scaffolding in moduleloader around a function-local args object; never-read `rope_scaling` write-back in the encoder config.

Every change is bitwise-neutral by construction (correctly-rounded elementwise ops, identical-value caching, fully-overwritten buffers) — the golden-gate lane is the proof; SSIM lanes as usual.

Not included (sized follow-up): the Qwen3-VL conditioner builds/loads all 64 LM layers but H3 reads only `hidden_states[50]` — truncating would drop ~14 GB of weights and ~22% of encode time, but needs the hidden-state indexing convention pinned and a checkpoint-key filter in `load_weights`.

### DIFF
--- a/fastvideo/configs/models/encoders/minimax_h3_qwen3_vl.py
+++ b/fastvideo/configs/models/encoders/minimax_h3_qwen3_vl.py
@@ -128,10 +128,6 @@ class MiniMaxH3Qwen3VLArchConfig(TextEncoderArchConfig):
         self.mrope_section = (int(section[0]), int(section[1]), int(section[2]))
         if sum(self.mrope_section) * 2 != self.head_dim:
             raise ValueError("MiniMax H3 Qwen3-VL mRoPE sections must cover exactly half of each attention head.")
-        rope_scaling["mrope_interleaved"] = self.mrope_interleaved
-        rope_scaling["mrope_section"] = list(self.mrope_section)
-        rope_scaling.setdefault("rope_type", "default")
-        self.rope_scaling = rope_scaling
 
         if self.vision_out_hidden_size != self.hidden_size:
             raise ValueError("MiniMax H3 Qwen3-VL vision_out_hidden_size must match the language hidden_size "
--- a/fastvideo/models/dits/minimax_h3.py
+++ b/fastvideo/models/dits/minimax_h3.py
@@ -18,6 +18,7 @@ from fastvideo.distributed.communication_op import (
 )
 from fastvideo.distributed.parallel_state import get_sp_world_size, model_parallel_is_initialized
 from fastvideo.layers.linear import ReplicatedLinear
+from fastvideo.layers.mlp import MLP
 from fastvideo.layers.quantization import QuantizationConfig
 from fastvideo.layers.visual_embedding import Timesteps
 from fastvideo.models.dits.base import BaseDiT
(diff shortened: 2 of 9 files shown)


SolitaryThinker (commented):
Review at head c8834fa (rebased onto main 6eb9569; conflict-free, ruff/yapf clean). Author self-review — treat the verdict accordingly.

**Probe 1 — root cause or symptom.** Each change removes the duplication/waste at its origin rather than patching a caller: the timestep MLP is deleted in favor of the identical `layers.mlp.MLP` (same submodule names, so the fp32-keep list at `minimax_h3.py:410` and param mapping still match by name) [read]; the per-step RoPE/text-refiner recompute is cached at the layer that produces it, identity-keyed on the exact tensors the denoising stage holds for the loop's lifetime, bypassed under grad and compile [read]; `index_copy_` replaces three full-buffer clones where the buffer is produced [read].

**Probe 2 — would a test fail without this?** For a bitwise-neutral refactor the guard is equivalence, so I executed it where CPU allows: old-vs-new `build_packed_sequence` compared bitwise on three layouts including first+last fl2va anchors — `position_ids`, `token_tags`, index sets all equal [measured]. The model-side changes are covered by the golden-gate lane (bitwise single-layer fingerprints) and SSIM lanes on this PR's CI [measured-by-CI, pending]. The normalize_latents swap rests on correctly-rounded elementwise-op equivalence plus preserved fp16 round-trip order [read] — the SSIM lanes are the executable check.

**Probe 3 — siblings.** `git grep` shows zero remaining references to the deleted `MiniMaxH3TimestepEmbedding` and `_component_value`; the `_module_dtype` hits in `minimax_h3_audio.py` are a distinct, still-used local helper, not an orphan [measured]. `construct_variable_block_sizes` gained a trailing defaulted param, so its existing Wan call sites are unaffected [read]. Known sibling NOT fixed here, by design: other DiTs share the per-step RoPE-rebuild pattern; the cache is H3-local because H3's stage guarantees stable position tensors across the loop — generalizing needs per-pipeline audits.

Caveats worth a reviewer's eye: (1) the two caches hold references to stage tensors between generations (~6 MB) — released on the next generation's first step; (2) `_refined_text` skips two refiner blocks + four collectives per step from step 1 — all ranks hit/miss together since each rank keys on its own held tensor; (3) under torch.compile both caches disable themselves, so #1689's compiled path is unchanged.

Verdict: PASS at c8834fa contingent on green golden-gate + SSIM lanes, with the standing note that this is a self-review — a second pair of eyes on the cache-lifetime reasoning in `_rotary_for`/`_refined_text` is the highest-value spot check.

SolitaryThinker:
Golden-gate executed on GB200 at this head (c8834fa): `test_minimax_h3_t2v_golden_gate` **1 passed** — bitwise fingerprint match, discharging the model-side equivalence contingency from the review (MLP swap, in-place index_copy_, RoPE/text caches). Methodological note: a first attempt failed because the runner exported FASTVIDEO_FA4=1 while the golden was minted without it — the gate's env fingerprint refused the comparison, which is exactly its job; rerun with matching env passed. [measured]

SolitaryThinker:
Diffusers bit-exactness re-verified on GB200 for this head, differentially [measured]: the pinned-reference parity suite (`tests/local_tests/minimax_h3/`, reference pinned at the port's diffusers commit) collected and passed **35/35 on base main (6eb9569)** and **35/35 at this head (c8834fa)** — scheduler, packing, ref2va, and source-parity modules all green on both sides. Together with the golden-gate bitwise pass above, component and pipeline parity with the diffusers reference is unchanged by this PR.

SolitaryThinker:
Head moved c8834fa → 65df3d8 for pre-commit: one whitespace-only fix (pinned yapf 0.43.0 collapses a double blank line in `minimax_h3_input_preparation.py` left by the helper deletion; my local yapf differed on it). The prior pre-commit failure was the stale pre-rebase run — the pinned hook is clean repo-wide on the rebased base. Golden-gate and 35/35 diffusers-parity evidence recorded at c8834fa carries over unchanged (blank-line-only delta).

PR #1699, opened 2026-08-07
[perf] MiniMax-H3: rank-reduced AdaLN pruned model option (-39% params, -23 GiB VRAM)

shaoxiongduan:
# [perf] MiniMax-H3: optional rank-reduced AdaLN (-39% params, -23 GiB VRAM)

## Purpose

MiniMax-H3's AdaLN modulation projections account for 13.01B of the DiT's 33.12B
parameters — 39% of the model — while contributing roughly 0.003% of its FLOPs. Each
projection consumes `silu(time_embedder(time_proj(t)))` where `t` is a scalar timestep, so
the set of inputs it can ever see is a smooth 1-D curve spanning a very small subspace,
even though the weight is stored as `[96768, 2688]`.

This PR adds an opt-in path that stores those projections against a shared low-rank basis
instead. At rank 16 the DiT drops to 20.17B parameters and peak VRAM falls by 22.8 GiB per
GPU, with no change in latency. Nothing changes for anyone who does not opt in.

Fixes # N/A

## Changes

- `MiniMaxH3ArchConfig.adaln_rank: int | None` (default `None`), with a range check in
  `__post_init__`. When it is set, `adaln_proj.linear` and `norm_out.linear` are built at
  `[out, adaln_rank]` and a shared `adaln_basis` projects `silu(temb)` into that subspace.
- `MiniMaxH3AdaLayerNormModulation` and `MiniMaxH3AdaLayerNormOut` gain an `apply_silu`
  flag, since SiLU is applied once before the basis rather than inside each consumer.
- `MiniMaxH3TransformerBlock.forward` narrows the six modulation tensors to the activation
  dtype so the block stays BF16 end to end.
- `_get_parameter_dtype` pins the factorized projections to FP16. This is deliberate and
  is explained under *Test Results*: BF16 is materially worse there, and FP32 is more
  accurate than the model being reproduced.
- `tools/minimax_h3/fit_adaln_basis.py` fits the basis by truncated SVD of the sampled
  timestep curve and writes a converted diffusers component directory. `adaln_rank` in the
  checkpoint's `config.json` reaches the arch config through the existing
  `update_model_arch`, so **no loader changes are required** and no runtime flag is needed.
- The three H3 examples carry a commented-out `--model-path` pointing at a converted
  checkpoint.

A converted rank-16 checkpoint for both partitions is published at
[noctuashap/MiniMax-H3-pruned-r16](https://huggingface.co/noctuashap/MiniMax-H3-pruned-r16).

## Test Plan

```bash
pre-commit run --files \
    fastvideo/configs/models/dits/minimax_h3.py \
    fastvideo/models/dits/minimax_h3.py \
    tools/minimax_h3/fit_adaln_basis.py

# fit error at candidate ranks, no checkpoint written
python tools/minimax_h3/fit_adaln_basis.py --src $H3/transformer --dst /dev/null \
    --rank 16 --report-only

# convert both partitions
python tools/minimax_h3/fit_adaln_basis.py --src $H3/transformer      --dst $OUT/transformer      --rank 16
python tools/minimax_h3/fit_adaln_basis.py --src $H3/transformer_ref  --dst $OUT/transformer_ref  --rank 16

# A/B through the stock example, 4x GB200, 50 steps, 1344x768, 124 frames, seed 0
python examples/inference/basic/basic_minimax_h3_t2v.py --model-path $H3  --steps 50 --num-gpus 4 --seed 0 --prompt "$P"
python examples/inference/basic/basic_minimax_h3_t2v.py --model-path $OUT --steps 50 --num-gpus 4 --seed 0 --prompt "$P"

# Ref2VA, same reference video and seed
python examples/inference/basic/basic_minimax_h3_ref2va.py --model-path $H3  --reference-video ref.mp4 --steps 50 --num-gpus 4 --seed 0 --prompt "$P2"
python examples/inference/basic/basic_minimax_h3_ref2va.py --model-path $OUT --reference-video ref.mp4 --steps 50 --num-gpus 4 --seed 0 --prompt "$P2"
```

## Test Results

**The released path is untouched.** With `adaln_rank` unset, a full-rank generation on this
branch is byte-for-byte identical to one on `main` — MS-SSIM 1.000000, 100% identical
pixels, zero audio difference. Repeated runs and SP=1 vs SP=4 are also bit-identical, so
the pipeline has no run-to-run noise and that comparison is meaningful.

| | official | rank-16 |
|---|---:|---:|
| parameters | 33.12B | 20.17B |
| peak VRAM / GPU | 88.3 GiB | 65.5 GiB |
| T2VA latency (warm) | 95.05 s | 95.44 s |
| Ref2VA latency | 355.93 s | 354.60 s |

Side-by-side, official weights on the left and the rank-16 checkpoint on the right. Same
prompt, same seed, both generated with the stock `basic_minimax_h3_t2v.py`:

https://github.com/user-attachments/assets/47f4b66b-41df-4fbb-b63d-dea39b46544c

<details>
<summary>Fit error and the FP16 choice</summary>

```
Fit error over all 51 projections, vs an FP64 reference:
  rank  8 : max|err| 6.678e-03   relative 5.4e-04
  rank 16 : max|err| 6.306e-08   relative 5.1e-09
  rank 32 : max|err| 1.013e-13   relative 8.2e-15

Runtime error through the whole chain, including the narrowing cast at the
apply site, per block at rank 16:
  released full-rank BF16   3.4e-03 .. 4.4e-03
  factorized BF16           6.4e-03 .. 7.6e-03    (~1.7x worse)
  factorized FP16           3.1e-03 .. 4.6e-03    (matches released)
  factorized FP32           2.4e-03 .. 3.8e-03    (better than released)
```

The BF16 error is flat from rank 8 through rank 128, so it is cancellation rather than
truncation: `u(t)` is 2688 values (max 0.806, mean 1.4e-03) summed as many small terms,
and the factorization rebuilds the same value from `r` large terms that cancel. FP16 has
two more mantissa bits than BF16, which is enough to absorb that. FP32 was rejected because
the aim is to reproduce the released model, not to improve on it. ComfyUI's independent
rank-8 variant stores these tensors as F16 for the same reason, and raises precision only
for its factorized checkpoints.

</details>

<details>
<summary>Conversion output</summary>

```
$ python tools/minimax_h3/fit_adaln_basis.py --src .../transformer --dst .../transformer --rank 16
basis: U=(4096, 2688) rank=16 relative residual ||U-U_r||/||U|| = 1.495e-09
modulation error over all projections: max|err|=6.306e-08 (|Wu|max=12.376, relative=5.095e-09)
parameters: 33.123B -> 20.166B (39.1% removed)
bf16 footprint: 66.2 GB -> ~40.3 GB

$ ... --src .../transformer_ref --dst .../transformer_ref --rank 16
basis: U=(4096, 2688) rank=16 relative residual ||U-U_r||/||U|| = 1.660e-09
modulation error over all projections: max|err|=6.435e-08 (|Wu|max=12.624, relative=5.097e-09)
```

`transformer` and `transformer_ref` need separate fits: their `time_embedder` weights
differ by 1–6% relative, so one basis cannot serve both.

</details>

## Checklist

- [x] I ran `pre-commit run --all-files` and fixed all issues — *ran on the changed files;
      yapf, ruff and codespell pass. mypy fails, but identically on untouched files, because
      it reads the worktree directory name as a package; it should be re-checked on a
      normally-named checkout.*
- [ ] I added or updated tests for my changes — **not done.** Verification used out-of-tree
      scripts (parity harness, dtype diagnostics). Happy to upstream a rank-fit unit test
      and a dtype-resolution test if wanted.
- [x] I updated documentation if needed — examples carry the alternative `--model-path`; the
      converter documents its own usage. No support-matrix change, since this adds no model.
- [x] I considered GPU memory impact of my changes — that is the point of the PR; measured
      88.3 → 65.5 GiB peak per GPU.

- [ ] I verified SSIM regression tests pass — **they will not pass with a pruned
      checkpoint, by construction.** See *Test Results*. With `adaln_rank` unset the
      released path is bit-identical, so existing references remain valid for the default
      configuration.
- [ ] I updated the support matrix if adding a new model — not applicable, no new model.

## Known limitations

- Quality was checked on one prompt per task (T2VA and Ref2VA), compared against the
  full-rank output and inspected by eye. There is no multi-prompt sweep behind these
  numbers.
- Rank 16 is a safety margin rather than a requirement. It was chosen from FP64 fit error,
  and the later dtype work showed that FP16 rounding (3.1–4.6e-03) is 6–8× larger than
  rank-8 truncation (5.4e-04), so rank 8 would likely be indistinguishable.
- Should merge other commits touching H3 dit first, will rebase after that

### DIFF
--- a/examples/inference/basic/basic_minimax_h3_fl2va.py
+++ b/examples/inference/basic/basic_minimax_h3_fl2va.py
@@ -25,6 +25,7 @@ from fastvideo.pipelines.basic.minimax_h3.packing import resolve_canvas_size
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--model-path", default="MiniMaxAI/MiniMax-H3")
+    # parser.add_argument("--model-path", default="noctuashap/MiniMax-H3-pruned-r16")  # rank-16 AdaLN: -39% params, -23 GiB VRAM
     parser.add_argument("--image", required=True, help="First-frame image path.")
     parser.add_argument("--last-image", help="Optional last-frame image path.")
     parser.add_argument("--output", default="outputs/minimax_h3_fl2va")
--- a/examples/inference/basic/basic_minimax_h3_fl2va.py
+++ b/examples/inference/basic/basic_minimax_h3_fl2va.py
@@ -25,7 +25,12 @@ from fastvideo.pipelines.basic.minimax_h3.packing import resolve_canvas_size
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--model-path", default="MiniMaxAI/MiniMax-H3")
-    # parser.add_argument("--model-path", default="noctuashap/MiniMax-H3-pruned-r16")  # rank-16 AdaLN: -39% params, -23 GiB VRAM
+    # Rank-reduced AdaLN checkpoint (-39% params, -23 GiB VRAM): pass
+    #   --model-path noctuashap/MiniMax-H3-pruned-r16
+    # (or a local dir converted with tools/minimax_h3/fit_adaln_basis.py).
+    # adaln_rank is read from the checkpoint config; no other flags needed.
+    # Rank-reduced checkpoints are inference-only: training needs the
+    # full-rank release.
     parser.add_argument("--image", required=True, help="First-frame image path.")
     parser.add_argument("--last-image", help="Optional last-frame image path.")
     parser.add_argument("--output", default="outputs/minimax_h3_fl2va")
--- a/examples/inference/basic/basic_minimax_h3_ref2va.py
+++ b/examples/inference/basic/basic_minimax_h3_ref2va.py
@@ -25,6 +25,7 @@ from fastvideo.pipelines.basic.minimax_h3 import MiniMaxH3Reference
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--model-path", default="MiniMaxAI/MiniMax-H3")
+    # parser.add_argument("--model-path", default="noctuashap/MiniMax-H3-pruned-r16")  # rank-16 AdaLN: -39% params, -23 GiB VRAM
     parser.add_argument("--reference-video", required=True)
     parser.add_argument("--reference-audio", help="Optional additional audio reference.")
     parser.add_argument("--output", default="outputs/minimax_h3_ref2va")
--- a/examples/inference/basic/basic_minimax_h3_ref2va.py
+++ b/examples/inference/basic/basic_minimax_h3_ref2va.py
@@ -25,7 +25,12 @@ from fastvideo.pipelines.basic.minimax_h3 import MiniMaxH3Reference
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--model-path", default="MiniMaxAI/MiniMax-H3")
-    # parser.add_argument("--model-path", default="noctuashap/MiniMax-H3-pruned-r16")  # rank-16 AdaLN: -39% params, -23 GiB VRAM
+    # Rank-reduced AdaLN checkpoint (-39% params, -23 GiB VRAM): pass
+    #   --model-path noctuashap/MiniMax-H3-pruned-r16
+    # (or a local dir converted with tools/minimax_h3/fit_adaln_basis.py).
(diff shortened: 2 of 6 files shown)


SolitaryThinker:
thanks! which audio is the example video playing?

shaoxiongduan:
their audio are near identical, so i took the original video's audio. here's both of them side by side:

Original H3:
https://github.com/user-attachments/assets/1798704a-d382-4d78-99bf-621a1cf2dedb

Pruned H3:
https://github.com/user-attachments/assets/73680123-5045-4030-b162-0b8e677853dc

SolitaryThinker (approved):

PR #1703, opened 2026-08-12
[perf] Reduce MiniMax-H3 VAE peak memory

H1yori233:
## Purpose

Reduce MiniMax-H3 VAE peak memory during reference-video encoding and distributed video/audio decoding without changing model numerics.

The existing path:

- materializes the full reference video on the accelerator before encoding;
- retains the complete decoded video on the accelerator before copying it to CPU;
- repeats VAE decoding on every sequence-parallel rank.

Inspired by [ComfyUI#15446](https://github.com/Comfy-Org/ComfyUI/pull/15446).

## Changes

- Encode reference videos one VAE clip at a time while keeping source pixels on CPU.
- Share one temporal chunk iterator between the existing `decode` path and the streaming decode path.
- Copy finalized decoded chunks directly into a CPU output buffer.
- Run video and audio VAE decoding only on the executor output rank.
- Add separate VAE and pipeline-stage regression tests.
- Extend official-reference parity coverage to the streamed pixel output.

The implementation remains entirely within the MiniMax-H3 model and stages. It does not add generic pipeline configuration or executor behavior.

## Performance

Measured with a 124-frame, 768×1344 video on GB200:

| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| Reference encode peak GPU memory | 7746.7 MiB | 4877.1 MiB | -37.0% |
| Reference encode time | 10.29 s | 10.05 s | -2.3% |
| Video decode peak GPU memory | 6206.7 MiB | 5832.1 MiB | -6.0% |
| Video decode time | 6.58 s | 6.67 s | +1.4% |

With SP=4:

| Metric | `main` | This PR | Change |
|---|---:|---:|---:|
| Aggregate VAE decode peak | 23362.4 MiB | 5835.5 MiB | -75.0% |
| Non-output-rank VAE allocation | 5840.6 MiB/rank | 0 MiB/rank | eliminated |

The output-rank result was numerically identical to the previous path (`max_abs = 0`).

## Test Plan

```bash
pytest \
  fastvideo/tests/vaes/test_minimax_h3_video_vae_streaming.py \
  fastvideo/tests/stages/test_minimax_h3_vae_streaming.py \
  tests/local_tests/pipelines/test_minimax_h3_pipeline_smoke.py -q

PYTHONPATH="$MINIMAX_H3_OFFICIAL_REF_DIR/src:$PWD" \
MINIMAX_H3_MODEL_ROOT=/path/to/MiniMax-H3 \
MINIMAX_H3_RUN_VIDEO_VAE_PARITY=1 \
pytest tests/local_tests/vaes/test_minimax_h3_video_vae_parity.py -v -s

pytest fastvideo/tests/ssim/test_minimax_h3_similarity.py -v -s
```

## Test Results

- Focused unit and pipeline tests: `9 passed`
- Official video VAE parity: `1 passed`
  - encode mean/logvar: exact
  - normalized latents: exact
  - raw decode: exact
  - streamed pixels: exact
- SSIM: `1 passed`
  - mean: `0.98521`
  - minimum: `0.9800`
  - required threshold: `0.95`
- Pre-commit on all changed files: passed
- `git diff --check`: passed

## Trade-offs

- Video decode latency increased by approximately 1.4% in the measured run.
- This PR intentionally avoids changing generic output-buffer ownership. A temporary generic CPU-side copy may therefore still exist outside the MiniMax-H3 stages.

## Checklist

- [x] I verified SSIM regression tests pass
- [x] Support-matrix update is not applicable; no new model or workload was added

### DIFF
--- a/fastvideo/models/vaes/minimax_h3_video.py
+++ b/fastvideo/models/vaes/minimax_h3_video.py
@@ -747,43 +747,113 @@ class AutoencoderKLMiniMaxH3(nn.Module):
             moments = moments[:, :, :-self.config.token_drop]
         return moments
 
-    def _decode(self, z: torch.Tensor) -> torch.Tensor:
-        tokens_chunk_size = self.tokens_chunk_size
+    def _encode_pixels(self, pixels: torch.Tensor) -> torch.Tensor:
+        """Encode unnormalized pixels while keeping full videos off the accelerator."""
+        clip_length = self.config.clip_length
+        moments = []
+        for frame_start in range(0, pixels.shape[2], clip_length):
+            clip = pixels[:, :, frame_start:frame_start + clip_length].to(
+                device=self.pixel_mean.device,
+                dtype=torch.float32,
+            )
+            if pixels.dtype == torch.uint8:
+                clip = clip / 255.0
+            if clip.shape[2] < clip_length:
+                pad_frames = clip[:, :, -1:].repeat(1, 1, clip_length - clip.shape[2], 1, 1)
+                clip = torch.cat([clip, pad_frames], dim=2)
+            clip = self.normalize_pixels(clip)
+            moments.append(self._encode_clip(clip))
+            del clip
+        encoded = torch.cat(moments, dim=2)
+        if self.config.token_drop > 0:
+            encoded = encoded[:, :, :-self.config.token_drop]
+        return encoded
+
+    def _temporal_decode_plan(self, latent_num_frames: int) -> tuple[int, int, int]:
+        """Return pad tokens, chunk count, and exact decoded frame count."""
+        if latent_num_frames <= 0:
+            raise ValueError(f"MiniMax-H3 latent frame count must be positive, got {latent_num_frames}.")
+
         token_drop = self.config.token_drop
+        tokens_chunk_size = self.tokens_chunk_size
         temporal_ratio = self.temporal_compression_ratio
-        chunk_num_frames = tokens_chunk_size * temporal_ratio
-        num_tokens = z.shape[2] + token_drop
+        num_tokens = latent_num_frames + token_drop
         pad_tokens = (-num_tokens) % tokens_chunk_size
         num_chunks = (num_tokens + pad_tokens) // tokens_chunk_size - int(token_drop > 0)
+        if num_chunks < 1:
+            pad_tokens += tokens_chunk_size
+            num_chunks = 1
+
+        decoded_num_frames = num_chunks * (tokens_chunk_size * temporal_ratio - self.frame_pre_padding)
+        if token_drop > 0:
+            decoded_num_frames += self.frame_overlap
+        if pad_tokens > 0:
+            intra_tail = self.config.clip_length % temporal_ratio
+            pad_frames = sum(intra_tail if intra_tail and (latent_num_frames + offset) %
+                             tokens_chunk_size == 0 else temporal_ratio for offset in range(pad_tokens))
+            decoded_num_frames -= pad_frames
+        return pad_tokens, num_chunks, decoded_num_frames
+
+    def _decode_chunks(self, z: torch.Tensor):
+        """Yield finalized temporal chunks in decode order."""
+        tokens_chunk_size = self.tokens_chunk_size
+        chunk_num_frames = tokens_chunk_size * self.temporal_compression_ratio
+        pad_tokens, num_chunks, output_num_frames = self._temporal_decode_plan(z.shape[2])
         if pad_tokens > 0:
             z = torch.cat([z, z[:, :, -1:].repeat(1, 1, pad_tokens, 1, 1)], dim=2)
 
-        decoded_chunks = []
+        output_frame_start = 0
         overlap = None
         for index in range(num_chunks):
             start = index * tokens_chunk_size
             clip = self._decode_clip(z[:, :, start:start + tokens_chunk_size + self.token_overlap])
-            for overlap_index in range(int(token_drop > 0) + 1):
-                frame_start = overlap_index * chunk_num_frames
-                chunk = clip[:, :, frame_start:frame_start + chunk_num_frames]
-                chunk = chunk[:, :, self.frame_pre_padding:]
-                if overlap_index == 0:
-                    if overlap is not None:
-                        chunk = self._blend(overlap, chunk, self.frame_overlap, dim=-3)
-                    decoded_chunks.append(chunk)
-                else:
-                    overlap = chunk
-        if overlap is not None:
-            decoded_chunks.append(overlap)
-        decoded = torch.cat(decoded_chunks, dim=2)
+            chunk = clip[:, :, self.frame_pre_padding:chunk_num_frames]
+            next_overlap = None
+            if self.config.token_drop > 0:
+                next_overlap = clip[:, :, chunk_num_frames + self.frame_pre_padding:].clone()
+            if overlap is not None:
+                chunk = self._blend(overlap, chunk, self.frame_overlap, dim=-3)
 
-        if pad_tokens > 0:
-            intra_tail = self.config.clip_length % temporal_ratio
-            num_tokens_before_pad = z.shape[2] - pad_tokens
-            pad_frames = sum(intra_tail if intra_tail and (num_tokens_before_pad + offset) %
-                             tokens_chunk_size == 0 else temporal_ratio for offset in range(pad_tokens))
-            decoded = decoded[:, :, :-pad_frames]
-        return decoded
+            num_frames = min(chunk.shape[2], output_num_frames - output_frame_start)
+            if num_frames > 0:
+                yield chunk[:, :, :num_frames]
+                output_frame_start += num_frames
+            overlap = next_overlap
+
+        if overlap is not None and output_frame_start < output_num_frames:
+            yield overlap[:, :, :output_num_frames - output_frame_start]
+
+    def decoded_pixel_shape(self, latent_shape: torch.Size | tuple[int, ...]) -> tuple[int, int, int, int, int]:
+        """Return the exact CPU pixel-buffer shape for a latent tensor shape."""
+        if len(latent_shape) != 5:
+            raise ValueError(f"MiniMax-H3 latents must be five-dimensional, got shape {tuple(latent_shape)}.")
+        batch_size, channels, latent_num_frames, latent_height, latent_width = map(int, latent_shape)
+        if channels != self.latent_channels:
+            raise ValueError(f"MiniMax-H3 latents must have {self.latent_channels} channels, got {channels}.")
+        _, _, decoded_num_frames = self._temporal_decode_plan(latent_num_frames)
+        return (
+            batch_size,
+            int(self.config.out_channels),
+            decoded_num_frames,
+            latent_height * self.spatial_compression_ratio,
+            latent_width * self.spatial_compression_ratio,
+        )
+
+    def _decode_to_pixels(self, z: torch.Tensor, output: torch.Tensor) -> None:
+        """Decode temporal chunks and immediately copy finalized pixels to CPU."""
+        output_frame_start = 0
+        for chunk in self._decode_chunks(z):
+            num_frames = chunk.shape[2]
+            pixels = self.denormalize_pixels(chunk.float()).clamp_(0, 1)
+            output[:, :, output_frame_start:output_frame_start + num_frames].copy_(pixels)
+            output_frame_start += num_frames
+        if output_frame_start != output.shape[2]:
+            raise RuntimeError(
+                f"MiniMax-H3 decode wrote {output_frame_start} frames into an output buffer expecting "
+                f"{output.shape[2]}.")
+
+    def _decode(self, z: torch.Tensor) -> torch.Tensor:
+        return torch.cat(list(self._decode_chunks(z)), dim=2)
 
     def encode(
         self,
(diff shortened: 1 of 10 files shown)


Davids048 (changes requested):
I confirmed that the PR reduces CUDA memory. I have only two requested revisions before approval.

  ### 1. Add real-weight parity coverage for encode_pixels()

  FastVideo defines encode_pixels() here
  (https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/fastvideo/models/vaes/minimax_h3_video.py#L872-L892). The production reference-video
  pipeline calls this method
  (https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_latent_preparation.py#L118-L124)
  with CPU uint8 frames.

  The real-weight parity test instead calls ordinary model.encode()
  (https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/tests/local_tests/vaes/test_minimax_h3_video_vae_parity.py#L111-L120) with an FP32 CUDA
  tensor. Therefore, the real-weight test does not exercise the production encode path introduced by this PR.

  Please extend the real-weight parity test as follows:

  1. Create deterministic CPU uint8 frames. The existing 22-frame geometry is sufficient.
  2. Pass those frames directly to FastVideo’s production-weight encode_pixels() method.
  3. Construct the official Diffusers input from the same bytes using the required FP32 conversion and pixel normalization.
  4. Compare every value in the posterior mean and log-variance tensors using a declared tolerance.
  5. Verify the Diffusers reference automatically against commit
     abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc or against hashes of all required reference source files.

  This revision is complete when the automated test verifies the CPU-uint8 encode_pixels() path against the pinned Diffusers implementation using production weights.

  ### 2. Define and publish the memory benchmark

  The PR table reports “peak GPU memory,” but it does not define the metric or measurement boundary.

  For a configuration that keeps the complete FP32 VAE on the measured GPU, the VAE parameters alone occupy approximately 9.93 GiB. Therefore, the reported 4.88–7.75 GiB values
  do not represent absolute PyTorch CUDA allocation under that configuration. They may be valid under another formula or measurement boundary, but the PR does not provide
  enough information to determine which one.

  Please publish a reproducible benchmark with:

  - Source and inputs: Base and PR Git revisions, model snapshot, input geometry, and input source.
  - Environment: GPU model, PyTorch version, CUDA version, allocator configuration, and VAE CPU-offload setting.
  - Measurement procedure: Warmup count, measured repetition count, CUDA synchronization points, peak-statistics reset point, and measured stage boundary.
  - Reported metrics: Absolute allocated and reserved peaks, plus peak growth above stage start:
    peak allocation - allocation immediately before the stage.

  - Distributed reporting: Each SP=4 rank’s measurement and any aggregate labeled as a sum of rank-local maxima rather than a simultaneous node peak.
  - Artifacts: The exact command or script, raw output logs, and terminal exit status.

  This revision is complete when another reviewer can run one documented procedure on both revisions and reproduce an unambiguous memory table.

  After these two revisions, I have no other requested changes.

H1yori233:
> I confirmed that the PR reduces CUDA memory. I have only two requested revisions before approval.
> 
> ### 1. Add real-weight parity coverage for encode_pixels()
> FastVideo defines encode_pixels() here (
> 
> https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/fastvideo/models/vaes/minimax_h3_video.py#L872-L892
> 
> ). The production reference-video
> pipeline calls this method
> (
> https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_latent_preparation.py#L118-L124
> 
> )
> with CPU uint8 frames.
> The real-weight parity test instead calls ordinary model.encode() (
> 
> https://github.com/hao-ai-lab/FastVideo/blob/528cef02c4a2723a3106079cb5d4ee9a509f7056/tests/local_tests/vaes/test_minimax_h3_video_vae_parity.py#L111-L120
> 
> ) with an FP32 CUDA
> tensor. Therefore, the real-weight test does not exercise the production encode path introduced by this PR.
> Please extend the real-weight parity test as follows:
> 
> 1. Create deterministic CPU uint8 frames. The existing 22-frame geometry is sufficient.
> 2. Pass those frames directly to FastVideo’s production-weight encode_pixels() method.
> 3. Construct the official Diffusers input from the same bytes using the required FP32 conversion and pixel normalization.
> 4. Compare every value in the posterior mean and log-variance tensors using a declared tolerance.
> 5. Verify the Diffusers reference automatically against commit
>    abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc or against hashes of all required reference source files.
> 
> This revision is complete when the automated test verifies the CPU-uint8 encode_pixels() path against the pinned Diffusers implementation using production weights.
> 
> ### 2. Define and publish the memory benchmark
> The PR table reports “peak GPU memory,” but it does not define the metric or measurement boundary.
> 
> For a configuration that keeps the complete FP32 VAE on the measured GPU, the VAE parameters alone occupy approximately 9.93 GiB. Therefore, the reported 4.88–7.75 GiB values do not represent absolute PyTorch CUDA allocation under that configuration. They may be valid under another formula or measurement boundary, but the PR does not provide enough information to determine which one.
> 
> Please publish a reproducible benchmark with:
> 
> * Source and inputs: Base and PR Git revisions, model snapshot, input geometry, and input source.
> * Environment: GPU model, PyTorch version, CUDA version, allocator configuration, and VAE CPU-offload setting.
> * Measurement procedure: Warmup count, measured repetition count, CUDA synchronization points, peak-statistics reset point, and measured stage boundary.
> * Reported metrics: Absolute allocated and reserved peaks, plus peak growth above stage start:
>   peak allocation - allocation immediately before the stage.
> * Distributed reporting: Each SP=4 rank’s measurement and any aggregate labeled as a sum of rank-local maxima rather than a simultaneous node peak.
> * Artifacts: The exact command or script, raw output logs, and terminal exit status.
> 
> This revision is complete when another reviewer can run one documented procedure on both revisions and reproduce an unambiguous memory table.
> 
> After these two revisions, I have no other requested changes.

Thanks for the review. Both requested revisions are addressed.

1. The real-weight parity test now creates deterministic CPU `uint8` frames, calls FastVideo’s production `encode_pixels()` path directly, constructs the official Diffusers input from the same bytes, and verifies the reference checkout against commit `abc5e9bf71fd38f53cd471bc3acaa84bc5ecbfdc`. Posterior mean and log-variance match element by element with `atol=0` and `rtol=0` (`max_abs=0`).
2. I added a documented, reproducible [memory benchmark](https://github.com/H1yori233/FastVideo/blob/feat/kaiqin/h3-vae-memory/tests/local_tests/vaes/benchmark_minimax_h3_video_vae_memory.py). It records the source revision, model hashes, input geometry, GPU/software/allocator configuration, offload setting, measurement boundaries, synchronization and peak-reset procedure, every repetition, per-rank absolute and incremental allocated/reserved peaks, rank-local aggregates, output shapes, and exit status.

| Workload                  | Peak allocated                | Incremental allocated         | Peak reserved                 | Incremental reserved          |
| ------------------------- | ----------------------------- | ----------------------------- | ----------------------------- | ----------------------------- |
| Encode SP1                | 18.695 → 14.461 GiB (-22.65%) | 18.695 → 14.461 GiB (-22.65%) | 19.451 → 15.525 GiB (-20.18%) | 19.451 → 15.525 GiB (-20.18%) |
| Decode SP1                | 15.875 → 15.415 GiB (-2.90%)  | 15.854 → 15.393 GiB (-2.90%)  | 16.711 → 15.988 GiB (-4.32%)  | 16.633 → 15.910 GiB (-4.34%)  |
| Decode SP4 rank-local sum | 63.500 → 15.455 GiB (-75.66%) | 63.415 → 15.393 GiB (-75.73%) | 66.844 → 16.105 GiB (-75.91%) | 66.531 → 15.910 GiB (-76.09%) |

All six base/PR benchmark invocations completed successfully with exit status 0.

SolitaryThinker:
```markdown
Reviewed in depth on top of `74b409d7c`. Both of @Davids048's requested revisions are properly addressed in that
commit (production CPU-uint8 `encode_pixels()` parity vs the pinned Diffusers commit with `atol=rtol=0` and
automatic revision verification; fully specified reproducible memory benchmark with per-rank absolute+incremental
allocated/reserved peaks and sum-of-rank-maxima labeling).

Additional verification I ran:
- Implemented the pre-PR `_decode` verbatim as an oracle and swept latent frames 1–37 (all pad cases, blended
  seams, `token_drop=0`, batched slicing) on CPU: the new chunk iterator is **bit-exact** with the old algorithm
  everywhere it worked, and additionally handles T≤2 where the old code crashed on an empty `torch.cat`. The
  0.985 SSIM is harness tolerance, not seams.
- The SP output-rank gate is safe: the gated region has no collectives (no hang risk), both executors consume
  only `responses[0]` (world rank 0), placeholders pass the stage verifiers, cleanup is rank-symmetric, and
  neither training stack nor preprocess uses these stages.

I pushed two additive commits directly onto this PR's branch via maintainer edit (mirrored as `hao-ai-lab/FastVideo:pr1703-fixes`; fast-forward of the prior head — please
merge them into `feat/kaiqin/h3-vae-memory`):
- `aadb23f40` fixes the source of the +1.4% decode-time cost: the temporal slice of the CPU buffer is
  non-contiguous, so every per-chunk D2H copy staged through a **pageable CPU temporary** plus a CPU-side scatter
  (the pinned buffer was never actually used asynchronously). Copies now run per (batch, channel) plane —
  contiguous on both sides, direct memcpy — `non_blocking` when pinned, with one stream sync in
  `decode_to_pixels` under `try/finally`. Also hardens `_temporal_decode_plan` against non-positive plans and
  documents the `encode_pixels` input contract.
- `942f7db3d` adds a legacy-decode bit-exactness oracle test over the seam/pad geometries, `use_slicing`
  coverage for both new APIs, and a CUDA pinned-buffer exactness test.

All PR tests plus the new ones pass (14 passed, 1 CUDA-only skip on my CPU node). One ask before merge: re-run
the decode benchmark once on GPU after picking up `aadb23f40` to confirm the decode-time regression is gone.
With that, this LGTM — approving.
```

SolitaryThinker:
GPU re-validation of the PR at head `942f7db3d` (incl. `aadb23f40`/`942f7db3d`), as asked in the review. Environment: 1x GB200 tray (NVL72, driver 580.82.07), torch 2.12.0+cu130, model `/…/MiniMax-H3` (video VAE fp32, 10.42 GB params), production shape 768x1344x124 (latents `[1,24,37,48,84]`). All commands/scripts: the PR's own committed benchmark plus a decode-only harness; raw logs archived. Jobs: 2539 / 2551 / 2552 / 2553.

**Exactness and determinism (GPU, real weights)**
- `fastvideo/tests/vaes/test_minimax_h3_video_vae_streaming.py` + `fastvideo/tests/stages/test_minimax_h3_vae_streaming.py` + registry smoke: **15/15 pass on CUDA**, including `test_decode_to_pixels_pinned_buffer_matches_dense_on_cuda` (the async pinned path).
- Full-stage decode at the production shape, old path (`8208536cd`) vs PR head: **max_abs = 0.0, bitwise equal** (pin on and off). Same-seed decode is deterministic: one sha256 across 5 in-process iterations and across two separate processes, and that hash is **identical to the pre-PR output's** (`e98db820…`).

**Decode time — the +1.4% is gone**
Decode-only boundary (VAE resident on GPU; the boundary the PR body's 6.58s -> 6.67s (+1.4%) numbers correspond to), median of 5:

| config | median | vs base |
|---|---:|---:|
| base `8208536cd` | 6.815 s | — |
| PR pre-fix `74b409d7c`, pinned off/on | 6.616 / 6.581 s | −2.9% / −3.4% (pinning bought nothing, as diagnosed) |
| **PR + fix `942f7db3d`, pinned off** | **6.580 s** | **−3.4%** |
| **PR + fix `942f7db3d`, pinned on** | **6.367 s** | **−6.6%** (production default: `pin_cpu_memory=True`) |

Raw copy microbench of just the D2H patterns (1.54 GB video buffer, 7 chunks): pre-fix strided whole-slice copy 25.5 ms (56 GiB/s) vs fixed per-plane 11.1 ms (129 GiB/s) unpinned, 12.4 ms non_blocking pinned — the fix's pattern is ~2.3x faster at the copy level on GB200, and far more on PCIe-class hosts where pageable staging is slow.

**Peak memory (PR's committed benchmark, `--repetitions 5`, pin=False, `vae_cpu_offload=True`)**

| workload | base | PR head | change |
|---|---:|---:|---:|
| decode SP1 peak alloc | 14.774 GiB | 10.934 GiB | **−26.0%** |
| decode SP1 peak reserved | 16.508 GiB | 11.758 GiB | −28.8% |
| decode SP1 @ 243 frames peak alloc | 19.603 GiB | 10.959 GiB | **−44.1%** (streamed peak is length-flat) |
| decode SP4 sum of rank-local peaks | 59.094 GiB | 10.975 GiB | **−81.4%** |
| decode SP4 non-output ranks | 14.774 GiB/rank | **0.014 GiB / 0.000 s per rank** | eliminated |

Host peak RSS is unchanged (19.8–20.3 GiB across all three revisions on the committed benchmark).

**SP-4 end-to-end (full H3 pipeline, `sp_size=4`, MultiprocExecutor, PR head, base checkpoint)**
4-transformer-forward run (native 5-point grid) at 768x1344x124, 1 warmup + 2 timed requests, stage logging + 500 ms `nvidia-smi` sampling:
- **No hangs** — three full passes through the gated stages; default stage verification passed on every rank (i.e., the non-output-rank placeholders satisfied the verifiers). Videos and audio rendered.
- **Non-output ranks idle through decode, empirically**: during every video-decode window GPU 0 ran at ~70–74% util and climbed to 44.9 GB (VAE move + working set) while GPUs 1–3 sat at **0.0% util with flat memory** (33.7 GB, no VAE transfer). During denoise all four GPUs were at 100% — the gate affects exactly the decode stages.
- Rank-0 in-pipeline decode stage: 7.28–7.76 s, inside the 6.4–8.4 s band of our pre-PR reference runs for the same shape/executor, with e2e non-denoise fixed cost unchanged (~9.2 s vs 9.3 s reference) — i.e. no e2e decode-phase regression, while 3 of 4 GPUs stop burning ~7.5 s of decode compute and ~15 GB each per request.

**One benchmark-methodology footnote** (does not affect the above): the committed benchmark's specific configuration — `pin_cpu_memory=False` plus `vae_cpu_offload=True`, i.e. a 10.4 GB fp32 VAE H2D+D2H inside every timed repetition — is host-page-state sensitive on GB200 for *both* heads: the same pre-fix binary measured 6.76 s under one process history and 8.5–9.2 s under another. Under the benchmark's canonical sequence the fixed head measures slower (8.52 vs 6.76 s at 124f), but a phase split shows its decode+copies is actually *faster* (6.35 vs 6.45 s) and the entire delta sits in the *subsequent* `vae.to("cpu")` weight offload (1.94 vs 0.17 s; per-`cudaMemcpyAsync` 434 µs vs 121 µs under the profiler): the per-plane D2H DMA first-touches the fresh pageable output buffer's pages GPU-side, which degrades how the following 10.4 GB pageable weight offload faults. The effect is history-dependent, not a property of the code: across three independent jobs the *same* fixed-head pin=False config measured 7.17 / 8.52 / 9.57 s (and 14.6 s at 243f), the head-to-head delta swung from +0.28 s to +2.56 s, and under one history it fully inverted (fixed 7.3–7.6 s vs pre-fix 8.5–9.2 s steady reps). With a pinned output buffer — the production default, served by the caching pinned allocator — both heads sit in a tight band at this boundary (6.33 s pre-fix / 6.71 s fixed in one job, well inside that cross-job variance). If the unpinned+offload benchmark case is considered important, a trivial hardening is to take the per-plane path only when the buffer is pinned; happy to push that as a follow-up.

**Verdict**: the fix delivers — decode time is at or below baseline in every decode boundary (best case −6.6% with the production pinned buffer), the PR's memory wins hold and grow with video length (−26% SP1, −44% at 243 frames, −81% SP4 aggregate), outputs are bit-exact and deterministic, and the SP output-rank gate is hang-free with non-output ranks provably idle. LGTM to merge at `942f7db3d`.

SolitaryThinker (approved):

PR #1716, opened 2026-08-18
[bugfix]: fix modular ops activation checkpoint recomputation

shaoxiongduan:
## Purpose

Fix the modular trainer's `ops` activation-checkpoint policy. It previously
wrapped the transformer root and retained only a small set of PyTorch op
objects, so custom attention backends silently fell back to full attention
recomputation. Several modular model plugins also still imported the separate
legacy-training helper, which meant the corrected policy reached MiniMax H3
but not the other modular families.

## Changes

- Wrap each discovered transformer block in its own checkpoint region.
- Match supported fused attention operations by exact dispatcher name, avoiding
  import-time dependencies on optional backend registrations.
- Retain fused SDPA, FastVideo FlashAttention, and all autograd-enabled generic
  block-sparse attention forwards, including the current SM90, SM100a, and
  Triton VSA ops.
- Do not retain parameter collectives or inexpensive matrix multiplies.
- Remove the ineffective math-SDPA entry: PyTorch decomposes that path before
  the selective policy sees a retainable attention boundary.
- Route Wan and its descendants, LTX-2, Kandinsky5, and MiniMax H3 through the
  modular trainer's checkpoint helper; the legacy training stack is unchanged.
- Resolve Wan checkpoint configuration once so role-level overrides also enable
  checkpoint-safe causal KV-cache snapshots.
- Keep kernel registration drift coverage portable when `fastvideo_kernel` is
  unavailable on a supported platform.
- Document the `full` versus `ops` memory/runtime tradeoff and correct the
  canonical example-config path.

## Supported and fallback paths

`ops` retains supported dispatcher-visible fused attention outputs and can
reduce backward recomputation when memory headroom exists. It uses more
activation memory than `full`. Math SDPA, VMoBA, FA3, and `ATTN_QAT_TRAIN` do
not expose a retainable dispatcher boundary today, so their attention still
fully recomputes under `ops`. Memory-constrained jobs, including tight GB10
configurations, should retain `full`.

## Regression coverage

The CPU behavioral suite uses a real autograd-enabled custom op across four
transformer blocks. It verifies:

- forward counts: 4 without checkpointing, 8 under `full`, 4 when `ops`
  retains the custom op, and 8 for an unsupported op;
- exact output, loss, input-gradient, and parameter-gradient parity;
- per-block wrapper granularity and invalid-layout failure behavior;
- known and dynamically registered block-sparse op coverage;
- modular model-plugin wiring; and
- role-level versus global Wan causal checkpoint/cache resolution.

Fresh verification on rebased head `cce7fdf819a164b616cf946ac0e1889033f66f1e`:

```text
pytest fastvideo/tests/train/utils/test_activation_checkpoint.py -q
19 passed, 14 warnings in 0.26s

pytest fastvideo/tests/train/utils -q
98 passed, 14 warnings in 0.46s

pytest fastvideo/tests/train \
  --ignore=fastvideo/tests/train/models \
  --ignore=fastvideo/tests/train/methods -q
186 passed, 14 warnings in 0.49s

pytest fastvideo/tests/train/methods/test_minimax_h3_finetune.py::\
test_h3_model_uses_modular_activation_checkpointing -q
1 passed, 14 warnings in 0.25s

PYTHONPATH="$PWD" bash .buildkite/scripts/unit_test.sh
1012 passed, 7 skipped, 19 warnings in 23.32s

pre-commit run --files <all changed paths>
Passed
```

A real CUDA probe on the local NVIDIA GB10 confirmed that PyTorch 2.12
dispatches the tested shape through
`aten::_scaled_dot_product_flash_attention`. The `ops` policy executed four
underlying fused forward kernels across four blocks versus eight under `full`,
with exact output, loss, input-gradient, and parameter-gradient parity.

## Performance evidence scope

The earlier author measurements were single-sequence-parallel-rank GB200
experiments on an earlier/private H3 setup. They show the intended
speed-for-memory direction but are not presented here as a reproducible
current-tree SP=8 benchmark. This change's merge contract is functional
checkpoint behavior and numerical parity; production throughput should be
validated with each job's actual backend, sharding, and memory headroom.

SSIM is not required: this changes which activations are retained versus
recomputed during modular training, not inference output or model weights.

## Checklist

- [x] Rebased onto current `main` after PR #1715
- [x] Added behavioral, numerical, wrapper, failure, op-drift, model-wiring,
      and causal-cache tests
- [x] Updated modular training documentation and example guidance
- [x] Ran the exact Buildkite unit-lane script locally
- [x] Ran repository pre-commit hooks on all applicable changed paths

### DIFF
--- a/docs/training/train_infra.md
+++ b/docs/training/train_infra.md
@@ -51,7 +51,7 @@ torchrun --nproc_per_node=8 \
 ## Config Format
 
 Every run is defined by a single YAML file with five top-level sections.
-See `examples/train/example.yaml` for a fully-commented reference.
+See `examples/train/configs/example.yaml` for a fully-commented reference.
 
 ### `models` — Role-based model instances
 
--- a/docs/training/train_infra.md
+++ b/docs/training/train_infra.md
@@ -80,8 +80,16 @@ Common model parameters:
 | `trainable` | `true` | Whether the model's parameters require gradients |
 | `disable_custom_init_weights` | `false` | Skip custom weight initialization (use for teacher/critic) |
 | `flow_shift` | `3.0` | Timestep shifting factor |
-| `enable_gradient_checkpointing_type` | `null` | Gradient checkpointing (`"full"` or `null`) |
+| `enable_gradient_checkpointing_type` | `null` | Gradient checkpointing (`"full"`, `"ops"`, or `null`) |
 | `attention_backend` | `null` | Optional role-local backend for Wan models (for example `ATTN_QAT_TRAIN`); overrides the process default only while this role's transformer is built |
+
+`full` recomputes every transformer-block operation and is the
+memory-conservative choice. `ops` retains outputs from supported,
+dispatcher-visible fused attention operations. It can reduce recompute time at
+the cost of higher activation memory, so use it only when the training shape has
+verified memory headroom. Math SDPA, VMoBA, FA3, and `ATTN_QAT_TRAIN` do not
+expose a retainable dispatcher boundary and therefore still use full attention
+recomputation under `ops`.
 
 Which roles are needed depends on the training method:
 
--- a/examples/train/configs/example.yaml
+++ b/examples/train/configs/example.yaml
@@ -147,7 +147,11 @@ training:
     mode_scale: 1.0                   # default: 1.0
     precondition_outputs: false       # default: false
     moba_config: {}                   # default: {}
-    enable_gradient_checkpointing_type: full  # default: null ("full" or null)
+    # default: null ("full", "ops", or null). "full" minimizes activation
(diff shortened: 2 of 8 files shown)


alexzms (dismissed):
LGTM. Verified on GB200: ops now wraps per block and retains attention (6 flash_fwd launches across 6 blocks, vs 12 under full), with loss and input grad bit-identical to no checkpointing, and func.name() never raises on the ops a real step dispatches.

The red checks look stale — builds #4878/#4905 fall in a window where unit-tests were red repo-wide (#1726 and #1739 too), and ./fastvideo/tests/train/ passes on this branch locally (161 tests). Worth a CI re-run rather than a code change.

PR #1719, opened 2026-08-19
[kernel] sm_100a CUDA block-sparse VSA forward (Blackwell), 64- and 128-token blocks

lpc0220:
## What

A CUDA forward for VSA's fine block-sparse attention on Blackwell (sm_100a), added as a
third backend behind the existing VSA op alongside the Triton and CuTe-DSL paths. Forward
only: it returns `(out, lse)` with `lse` in exactly the form
`triton_block_sparse_attn_forward` writes -- `max(qk * qk_scale) + log2(l)`, `[B, H, S]`
fp32 -- so `block_sparse_attn_backward_triton` runs against it unchanged and training
keeps working.

One extension carries two instantiations of the kernel:
- **64-token blocks** -- matches `VSA_TILE_SIZE = (4, 4, 4)`, a drop-in for the existing
  sparsity with no tile-size or top-k granularity change.
- **128-token blocks** -- tile volume `(4, 4, 8)`; `_SUPPORTED_VSA_BLOCK_VOLUMES` gains 128.

The Python backend (`block_sparse_attn_sm100a`) infers the block size from
`seqlen // num_blocks` and dispatches; `is_supported()` is conservative, so anything
outside the supported regime (non-sm_100a, non-bf16, head_dim != 128, block sizes with no
build) falls back to Triton rather than risking a wrong answer.

## Why

Measured at the kernel level against the Triton fine kernel and FA4's block-sparse path
(flash_attn.cute, 256-token blocks) on bit-identical shared inputs and top-k lists
(GB200, bf16, H=8, D=128, 25% density; TFLOPS, higher is better):

| S | triton (blk64) | ours blk64 | vs triton | fa4_v256 | ours blk128 | vs fa4 |
|---:|---:|---:|---:|---:|---:|---:|
| 4096 | 169 | 570 | 3.4x | 221 | 823 | 3.7x |
| 8192 | 307 | 890 | 2.9x | 557 | 1053 | 1.9x |
| 16384 | 436 | 1073 | 2.5x | 985 | 1197 | 1.2x |
| 32768 | 487 | 1179 | 2.4x | 1344 | 1372 | 1.02x |
| 65536 | 517 | 1143 | 2.2x | 1411 | 1429 | 1.01x |
| 131072 | 523 | 1057 | 2.0x | 1376 | 1265 | 0.92x |

fa4 runs 256-token blocks (its minimum), so that pair is context rather than like-for-like.

End-to-end in-framework, this backend vs the Triton fine kernel at the deployed Wan shapes:

| shape | S | sm100a | triton | speedup |
|---|---:|---:|---:|---:|
| 480P | 39936 | 3.08 ms | 6.47 ms | 2.10x |
| 720P | 92160 | 16.26 ms | 33.36 ms | 2.05x |

## Testing

`tests/test_block_sparse_sm100a.py`: 23 passed on GB200. Every case runs at both block
sizes against an explicit PyTorch reference (Triton's forward is 64-granular, so it cannot
referee the 128 build): forward and lse parity, ragged `variable_block_sizes`, top-k values
that do not divide the kernel's group size, several sequence lengths, a non-vacuity guard
on the lse check, and rejection of unsupported configs.


SolitaryThinker:
Nice kernel — we measured it on GB200 for MiniMax-H3's packed VSA and hit one correctness bug plus a multi-arch build break. Since `maintainer_can_modify` is on, we pushed fixes directly to this branch (three additive commits, no history rewrite): **5bdf50f** (kernel fix), **00d7d59** (regression tests), **cdf9a4a** (multi-arch guard). Details + evidence below.

## Bug: `q2k_num` was read once per CTA pair, not per q-tile (fixed in 5bdf50f)

`decode_workitem` did

```c
it.num_kv_blocks = q2k_num[it.global_mtile0];
```

and applied that single count to BOTH q-tiles of the CTA's adjacent pair — in the `get_kv_block_id` clamp (`min(idx, it.num_kv_blocks - 1)`) and in `get_vbs_thresholds`' masking predicates (`b0 < it.num_kv_blocks`). Consequences:

1. **Non-uniform per-row counts silently corrupt one tile of any pair whose two rows differ.** The odd tile is computed with the even tile's count: its `q2k_idx` row is over-read past its own `q2k_num` (or truncated), and none of the extra blocks are masked. Nothing crashes; the output is just wrong.
2. **`q2k_num == 0` on an even row hangs the kernel.** `num_k_tiles` becomes 0, so the load warp issues `load_v(mtile, -1)` (negative `q2k_idx` offsets) while the softmax warps never run `softmax_step`, so `full_bar_alpha` never arrives and the correction warps' first mbarrier wait has no matching arrive.

This matters beyond synthetic metadata: any mask whose per-row top-k varies hits (1), and `map_to_index` produces per-row counts, so the Python-level uniformity today is an accident of the benchmarks, not a contract. Concrete case: MiniMax-H3's packed layout has two row classes (prefix text/audio query tiles attend densely; video query tiles attend prefix+top-k), so whether a CTA pair straddles the class boundary depends on the *parity of the prefix tile count* — i.e. prompt-length-dependent silent corruption.

**The fix keeps the warp-specialized pipeline intact**: `decode_workitem` loads both counts and keeps `it.num_kv_blocks = max(cnt0, cnt1, 1)` as the pair's shared trip count, so every warp role derives identical loop bounds and every producer/consumer mbarrier pairing stays symmetric for ANY counts (a fully-empty pair walks one fully-masked K-tile instead of degenerating to `num_k_tiles == 0`). The load warp clamps into each tile's OWN `q2k_idx` row (an empty tile loads block 0 and never dereferences its row); the softmax warps mask with each tile's OWN count, so padded/empty positions get threshold 0 → the whole block masks to `-inf` → `p = 0`, `l = 0`, `scale_own = 0` → exactly-zero output rows with a finite lse. Uniform counts execute the pre-fix code path unchanged.

Also in 5bdf50f: `is_supported()` dropped `int(variable_block_sizes.min())` — a GPU→CPU sync on every call of a dispatch predicate, which never guarded the `q2k_num` hazard anyway (the hang above passes it). It now checks static facts only, with the accepted-input contract documented in the docstring.

## Red/green (GB200, CUDA 13.0.88, torch 2.12.0+cu130)

H3-shaped case: 16 q-tiles, 3 dense prefix rows (count 16), video rows count 7 (prefix 3 + top-k 4), ragged `variable_block_sizes` in [32, 64] — prefix tile count ODD so pair (2,3) straddles the row classes. Masked-SDPA fp32 reference, bf16 tolerance 0.02 (this suite's own).

* **pre-fix**: diverges on exactly the straddling video tile 3 and nowhere else — per-tile max |out diff| `0.0010 0.0010 0.0011 `**`0.6077`**` 0.0016 … 0.0014`, lse max |diff| 1.3151.
* **pre-fix, zero-count rows** (rows 5, 6, 8, 9 of the same case): **hangs** (subprocess killed at a 240 s timeout), matching the mbarrier analysis.
* **fixed**: every tile ≤ 0.0020, lse |diff| 0.0000; zero-count rows produce exactly-zero output with finite lse, including the pair where only the even row is empty; same-seed reruns are bitwise identical.
* `tests/test_block_sparse_sm100a.py`: the original 23 pass on both builds (uniform counts are the same code path before/after). 00d7d59 adds these cases to the suite — the straddling-pair case, the zero-count case (run in a subprocess with a timeout, because the pre-fix failure mode is a hang and a regression should fail the suite, not wedge it), and a same-seed bitwise-determinism check — at both block sizes.

## TFLOPS with the fix in (your table's methodology)

GB200, bf16, B=1, H=8, D=128, 25% density, uniform top-k, bit-identical inputs/top-k lists, `triton.testing.do_bench`, FLOPs = 4·B·H·D·S·(topk·block). Single-arch `TORCH_CUDA_ARCH_LIST=10.0a` build:

| S | triton (blk64) | pre-fix blk64 | **fixed blk64** | fixed/pre-fix | pre-fix blk128 | **fixed blk128** |
|---:|---:|---:|---:|---:|---:|---:|
| 4096 | 178 | 528 | **528** | 1.000 | 656 | **657** |
| 8192 | 398 | 773 | **781** | 1.011 | 914 | **963** |
| 16384 | 489 | 1000 | **999** | 0.999 | 1127 | **1118** |
| 32768 | 513 | 1068 | **1085** | 1.016 | 1268 | **1284** |
| 65536 | 536 | 1066 | **1070** | 1.003 | 1288 | **1286** |
| 131072 | 538 | 1008 | **1023** | 1.015 | 1298 | **1279** |

The per-tile count loads + zero guard cost nothing measurable (within ±1.6%, more often slightly ahead). Absolute numbers land close to your table on a different node/toolkit (32k: 513 triton / 1085 fixed vs your 487 / 1179 — same ~2.1x-over-Triton story).

End-to-end at MiniMax-H3's deployed geometry (packed 38,010 tokens → 672 tiles at 64, 56 heads, D=128, sparsity 0.9, genuinely non-uniform counts): layer-level 19.2 ms (Triton-64) → **11.4 ms** (fixed sm100a), 1.68x, outputs matching Triton to 0.002 max |diff|; at an odd-prefix-tile prompt shape (28,370 tokens, 506 tiles) 12.2 ms → **7.6 ms**, 1.60x.

## Multi-arch builds (fixed in cdf9a4a)

As merged today, any `TORCH_CUDA_ARCH_LIST` containing more than `10.0a` breaks the wheel build: torch's cmake appends `-gencode` for every list entry on top of this TU's pinned `compute_100a` pass, and `tcgen05` / `.cta_group` don't exist outside sm_100a. We reproduced with `TORCH_CUDA_ARCH_LIST="10.0a;12.0a"` (CUDA 13.0.88, aarch64): the build fails at exactly the two sm100a TUs with `ptxas error: Instruction 'tcgen05.alloc' not supported on .target 'sm_120a'`, everything else compiles.

cdf9a4a guards the kernel body FA-style: kept on the host pass and the sm_100a device pass (`__CUDA_ARCH__ == 1000 && defined(__CUDA_ARCH_FEAT_SM100_ALL)`, so a plain `10.0` list entry is also stubbed rather than fed tcgen05); every other device pass compiles an empty stub, unreachable at runtime because `is_supported()`/the host launcher never dispatch off sm_100. Verified: the `10.0a;12.0a` wheel build fails at ptxas without the guard and compiles clean with it, and per-arch PTX shows the `compute_100a` image keeping the tcgen05 body while `compute_120a` gets the stub. The full branch test suite (23 original + 6 new) plus the red/green cases above were then re-run on GB200 against the multi-arch wheel, all green — i.e. the guard leaves the sm_100a images intact.

## Known remaining gap

The kernel is still not wired into `block_sparse_attn_from_indices` / the bool-map entry, so nothing dispatches to it by default (downstream we run it behind an opt-in env with a Triton fallback and the Triton backward — the lse is already in Triton's M format, which makes that pairing clean). Happy to send the dispatch wiring as a separate PR if you want it, or leave routing policy to you.

lpc0220:
> Nice kernel — we measured it on GB200 for MiniMax-H3's packed VSA and hit one correctness bug plus a multi-arch build break. Since `maintainer_can_modify` is on, we pushed fixes directly to this branch (three additive commits, no history rewrite): **[5bdf50f](https://github.com/hao-ai-lab/FastVideo/commit/5bdf50f6199c6c0893e4c2fafb9be7a686ef44f4)** (kernel fix), **[00d7d59](https://github.com/hao-ai-lab/FastVideo/commit/00d7d5907e406221f77471d20da55120fd8f8640)** (regression tests), **[cdf9a4a](https://github.com/hao-ai-lab/FastVideo/commit/cdf9a4af757f6c3c1493a697ea09cca2c7edd344)** (multi-arch guard). Details + evidence below.
> 
> ## Bug: `q2k_num` was read once per CTA pair, not per q-tile (fixed in [5bdf50f](https://github.com/hao-ai-lab/FastVideo/commit/5bdf50f6199c6c0893e4c2fafb9be7a686ef44f4))
> `decode_workitem` did
> 
> ```c
> it.num_kv_blocks = q2k_num[it.global_mtile0];
> ```
> 
> and applied that single count to BOTH q-tiles of the CTA's adjacent pair — in the `get_kv_block_id` clamp (`min(idx, it.num_kv_blocks - 1)`) and in `get_vbs_thresholds`' masking predicates (`b0 < it.num_kv_blocks`). Consequences:
> 
> 1. **Non-uniform per-row counts silently corrupt one tile of any pair whose two rows differ.** The odd tile is computed with the even tile's count: its `q2k_idx` row is over-read past its own `q2k_num` (or truncated), and none of the extra blocks are masked. Nothing crashes; the output is just wrong.
> 2. **`q2k_num == 0` on an even row hangs the kernel.** `num_k_tiles` becomes 0, so the load warp issues `load_v(mtile, -1)` (negative `q2k_idx` offsets) while the softmax warps never run `softmax_step`, so `full_bar_alpha` never arrives and the correction warps' first mbarrier wait has no matching arrive.
> 
> This matters beyond synthetic metadata: any mask whose per-row top-k varies hits (1), and `map_to_index` produces per-row counts, so the Python-level uniformity today is an accident of the benchmarks, not a contract. Concrete case: MiniMax-H3's packed layout has two row classes (prefix text/audio query tiles attend densely; video query tiles attend prefix+top-k), so whether a CTA pair straddles the class boundary depends on the _parity of the prefix tile count_ — i.e. prompt-length-dependent silent corruption.
> 
> **The fix keeps the warp-specialized pipeline intact**: `decode_workitem` loads both counts and keeps `it.num_kv_blocks = max(cnt0, cnt1, 1)` as the pair's shared trip count, so every warp role derives identical loop bounds and every producer/consumer mbarrier pairing stays symmetric for ANY counts (a fully-empty pair walks one fully-masked K-tile instead of degenerating to `num_k_tiles == 0`). The load warp clamps into each tile's OWN `q2k_idx` row (an empty tile loads block 0 and never dereferences its row); the softmax warps mask with each tile's OWN count, so padded/empty positions get threshold 0 → the whole block masks to `-inf` → `p = 0`, `l = 0`, `scale_own = 0` → exactly-zero output rows with a finite lse. Uniform counts execute the pre-fix code path unchanged.
> 
> Also in [5bdf50f](https://github.com/hao-ai-lab/FastVideo/commit/5bdf50f6199c6c0893e4c2fafb9be7a686ef44f4): `is_supported()` dropped `int(variable_block_sizes.min())` — a GPU→CPU sync on every call of a dispatch predicate, which never guarded the `q2k_num` hazard anyway (the hang above passes it). It now checks static facts only, with the accepted-input contract documented in the docstring.
> 
> ## Red/green (GB200, CUDA 13.0.88, torch 2.12.0+cu130)
> H3-shaped case: 16 q-tiles, 3 dense prefix rows (count 16), video rows count 7 (prefix 3 + top-k 4), ragged `variable_block_sizes` in [32, 64] — prefix tile count ODD so pair (2,3) straddles the row classes. Masked-SDPA fp32 reference, bf16 tolerance 0.02 (this suite's own).
> 
> * **pre-fix**: diverges on exactly the straddling video tile 3 and nowhere else — per-tile max |out diff| `0.0010 0.0010 0.0011 `**`0.6077`**` 0.0016 … 0.0014`, lse max |diff| 1.3151.
> * **pre-fix, zero-count rows** (rows 5, 6, 8, 9 of the same case): **hangs** (subprocess killed at a 240 s timeout), matching the mbarrier analysis.
> * **fixed**: every tile ≤ 0.0020, lse |diff| 0.0000; zero-count rows produce exactly-zero output with finite lse, including the pair where only the even row is empty; same-seed reruns are bitwise identical.
> * `tests/test_block_sparse_sm100a.py`: the original 23 pass on both builds (uniform counts are the same code path before/after). [00d7d59](https://github.com/hao-ai-lab/FastVideo/commit/00d7d5907e406221f77471d20da55120fd8f8640) adds these cases to the suite — the straddling-pair case, the zero-count case (run in a subprocess with a timeout, because the pre-fix failure mode is a hang and a regression should fail the suite, not wedge it), and a same-seed bitwise-determinism check — at both block sizes.
> 
> ## TFLOPS with the fix in (your table's methodology)
> GB200, bf16, B=1, H=8, D=128, 25% density, uniform top-k, bit-identical inputs/top-k lists, `triton.testing.do_bench`, FLOPs = 4·B·H·D·S·(topk·block). Single-arch `TORCH_CUDA_ARCH_LIST=10.0a` build:
> 
> S	triton (blk64)	pre-fix blk64	**fixed blk64**	fixed/pre-fix	pre-fix blk128	**fixed blk128**
> 4096	178	528	**528**	1.000	656	**657**
> 8192	398	773	**781**	1.011	914	**963**
> 16384	489	1000	**999**	0.999	1127	**1118**
> 32768	513	1068	**1085**	1.016	1268	**1284**
> 65536	536	1066	**1070**	1.003	1288	**1286**
> 131072	538	1008	**1023**	1.015	1298	**1279**
> The per-tile count loads + zero guard cost nothing measurable (within ±1.6%, more often slightly ahead). Absolute numbers land close to your table on a different node/toolkit (32k: 513 triton / 1085 fixed vs your 487 / 1179 — same ~2.1x-over-Triton story).
> 
> End-to-end at MiniMax-H3's deployed geometry (packed 38,010 tokens → 672 tiles at 64, 56 heads, D=128, sparsity 0.9, genuinely non-uniform counts): layer-level 19.2 ms (Triton-64) → **11.4 ms** (fixed sm100a), 1.68x, outputs matching Triton to 0.002 max |diff|; at an odd-prefix-tile prompt shape (28,370 tokens, 506 tiles) 12.2 ms → **7.6 ms**, 1.60x.
> 
> ## Multi-arch builds (fixed in [cdf9a4a](https://github.com/hao-ai-lab/FastVideo/commit/cdf9a4af757f6c3c1493a697ea09cca2c7edd344))
> As merged today, any `TORCH_CUDA_ARCH_LIST` containing more than `10.0a` breaks the wheel build: torch's cmake appends `-gencode` for every list entry on top of this TU's pinned `compute_100a` pass, and `tcgen05` / `.cta_group` don't exist outside sm_100a. We reproduced with `TORCH_CUDA_ARCH_LIST="10.0a;12.0a"` (CUDA 13.0.88, aarch64): the build fails at exactly the two sm100a TUs with `ptxas error: Instruction 'tcgen05.alloc' not supported on .target 'sm_120a'`, everything else compiles.
> 
> [cdf9a4a](https://github.com/hao-ai-lab/FastVideo/commit/cdf9a4af757f6c3c1493a697ea09cca2c7edd344) guards the kernel body FA-style: kept on the host pass and the sm_100a device pass (`__CUDA_ARCH__ == 1000 && defined(__CUDA_ARCH_FEAT_SM100_ALL)`, so a plain `10.0` list entry is also stubbed rather than fed tcgen05); every other device pass compiles an empty stub, unreachable at runtime because `is_supported()`/the host launcher never dispatch off sm_100. Verified: the `10.0a;12.0a` wheel build fails at ptxas without the guard and compiles clean with it, and per-arch PTX shows the `compute_100a` image keeping the tcgen05 body while `compute_120a` gets the stub. The full branch test suite (23 original + 6 new) plus the red/green cases above were then re-run on GB200 against the multi-arch wheel, all green — i.e. the guard leaves the sm_100a images intact.
> 
> ## Known remaining gap
> The kernel is still not wired into `block_sparse_attn_from_indices` / the bool-map entry, so nothing dispatches to it by default (downstream we run it behind an opt-in env with a Triton fallback and the Triton backward — the lse is already in Triton's M format, which makes that pairing clean). Happy to send the dispatch wiring as a separate PR if you want it, or leave routing policy to you.

Thank you for the careful review work. Great finding. I will integrate that bug fix in the code. 

> Happy to send the dispatch wiring as a separate PR if you want it, or leave routing policy to you. 
Let me check this with Claude.

PR #1732, opened 2026-08-21
[perf] Optimize MiniMax-H3 text encoder memory

H1yori233:
## Purpose

Reduce MiniMax-H3 text-encoder memory and checkpoint load cost while preserving its selected layer-50 conditioning output.

This incorporates and hardens the layer-truncation direction from #1711, redesigns the H3 forward contract following [vllm-omni#5691](https://github.com/vllm-project/vllm-omni/pull/5691), and adds checkpoint-serialized block FP8 support adapted from [sglang#34986](https://github.com/sgl-project/sglang/pull/34986). #1710 is orthogonal unified-memory/offload work and is intentionally not included.

## Changes

- build only the first 50 of 64 Qwen3-VL language layers, omit the unused final norm, and return the selected tensor without retaining a hidden-state tuple
- make the H3 conditioner own batching, multimodal insertion, DeepStack, and mRoPE construction; update the conditioning stage to consume `[sequence, hidden]`
- validate tap/build-depth configuration and preserve strict checkpoint loading while filtering only valid omitted layer keys
- add checkpoint-serialized E4M3 block FP8 loading with SGLang-style 1x128 activation quantization and FlashInfer groupwise GEMM on Blackwell
- keep embeddings, norms, and vision BF16; fail closed when checkpoint metadata, GPU capability, Triton, or FlashInfer are incompatible
- add unit coverage for truncation, the slim contract, strict loading, serialized FP8 metadata/weights, dependency gates, and FlashInfer layouts
- update the production-loader parity harness to compare official `hidden_states[50]` with the slim tensor for text, image, and video inputs

## Test Plan

```bash
python -m pytest \
  fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_truncation.py \
  fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py -q

pre-commit run --files \
  fastvideo/configs/models/encoders/minimax_h3_qwen3_vl.py \
  fastvideo/models/encoders/base.py \
  fastvideo/models/encoders/minimax_h3_checkpoint_fp8.py \
  fastvideo/models/encoders/minimax_h3_qwen3_vl.py \
  fastvideo/models/loader/component_loader.py \
  fastvideo/models/loader/text_encoder_quantization.py \
  fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_conditioning.py \
  fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py \
  fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_truncation.py \
  tests/local_tests/encoders/test_minimax_h3_qwen3_vl_parity.py \
  tests/local_tests/minimax_h3/README.md
```

## Test Results

- Unit tests: `22 passed`
- Pre-commit: yapf, ruff, codespell, mypy, and filename checks passed
- Slim forward (559 tokens, dense FA4):
  - peak allocated: `65,402.07 -> 65,139.69 MiB` (`-262.38 MiB`)
  - latency: `47.827 -> 48.344 ms` (`1.08% slower`; no latency speedup claimed)
- Serialized FP8 (559 tokens, 1x GB200):
  - checkpoint load: `51.644 -> 30.190 s` (`1.71x faster`)
  - peak allocated: `50,853.96 -> 28,851.57 MiB` (`-43.27%`)
  - steady encode: `44.76 -> 82.71 ms` (`1.85x slower`)
  - output cosine similarity: `0.999896`; relative RMSE: `1.442%`
- In official-size dense-FA4 E2E runs, text conditioning was `0.55%-2.38%` of request latency. FP8 is therefore presented as a memory/load optimization, not an E2E latency optimization.
- The updated real-checkpoint text/image/video exact parity test remains gated for a CUDA checkpoint run.

<details>

https://github.com/user-attachments/assets/afd500f4-6e2f-4d65-9d04-911d3e2dbb67

https://github.com/user-attachments/assets/198ae38e-f76b-49e5-88db-bcef94a5f220

https://github.com/user-attachments/assets/11a06423-21c3-4220-b6d4-ad9700e8b651

https://github.com/user-attachments/assets/2152449f-65b4-4347-b5a4-1b7ecd2e7993

https://github.com/user-attachments/assets/caf9beb1-36e3-4a28-8084-304eff7e66b0

https://github.com/user-attachments/assets/7a9e77aa-3842-4872-b87a-1a142aadc4c3

https://github.com/user-attachments/assets/a29ea410-9d9c-4252-ae02-d1920e53787e

https://github.com/user-attachments/assets/3bde8dc8-1e93-4cc4-8ac5-0ec46940821f

<summary>Unit-test output</summary>

```text
22 passed, 14 warnings in 6.34s
```

</details>

## Checklist

- [x] I updated documentation if needed (no documentation changes are required)

- [x] I updated the support matrix if adding a new model (N/A: no new model)

### DIFF
--- a/fastvideo/configs/models/encoders/minimax_h3_qwen3_vl.py
+++ b/fastvideo/configs/models/encoders/minimax_h3_qwen3_vl.py
@@ -62,14 +62,7 @@ class MiniMaxH3Qwen3VLArchConfig(TextEncoderArchConfig):
     hidden_size: int = 5120
     intermediate_size: int = 25600
     num_hidden_layers: int = 64
-    # H3 conditions on one intermediate hidden state and reads nothing above it,
-    # so the remaining layers are built, weight-loaded and then discarded: 14
-    # layers, 13.7 GB in bf16. Building exactly this many leaves that hidden
-    # state bit-identical, because the tuple records each layer's *input*, so
-    # entry N is the output of layer N-1. Set to None to keep the full stack.
-    # Must equal MINIMAX_H3_TEXT_ENCODER_LAYER in
-    # fastvideo/pipelines/basic/minimax_h3/packing.py; a test pins them together
-    # rather than importing across the models -> pipelines boundary.
+    output_hidden_state_index: int = 50
     num_hidden_layers_override: int | None = 50
     num_attention_heads: int = 64
     num_key_value_heads: int = 8
@@ -116,7 +109,7 @@ class MiniMaxH3Qwen3VLArchConfig(TextEncoderArchConfig):
     vision_initializer_range: float = 0.02
     vision_deepstack_visual_indexes: tuple[int, ...] = (8, 16, 24)
 
-    output_hidden_states: bool = True
+    output_hidden_states: bool = False
     stacked_params_mapping: list[tuple[str, str, str | int]] = field(default_factory=list)
     _fsdp_shard_conditions: list = field(default_factory=lambda: [
         _is_language_transformer_layer,
@@ -127,15 +120,16 @@ class MiniMaxH3Qwen3VLArchConfig(TextEncoderArchConfig):
     ])
 
     def __post_init__(self) -> None:
-        # Runs both at construction and after ``update_model_arch`` merges the
-        # checkpoint's config.json, so it also guards config-file overrides. A
-        # non-positive override would build no decoder layers at all, and a
-        # negative one would additionally make the surplus-key filter drop
-        # every ``language_model.layers.*`` checkpoint key, so the conditioner
-        # would "load" with no transformer stack and only fail at generation.
-        if self.num_hidden_layers_override is not None and self.num_hidden_layers_override < 1:
-            raise ValueError("MiniMax H3 Qwen3-VL num_hidden_layers_override must be a positive layer count "
-                             f"or None for the full stack; got {self.num_hidden_layers_override}.")
+        if self.output_hidden_state_index <= 0 or self.output_hidden_state_index > self.num_hidden_layers:
+            raise ValueError("MiniMax H3 Qwen3-VL output_hidden_state_index must be in "
+                             f"[1, {self.num_hidden_layers}], got {self.output_hidden_state_index}.")
+        if self.num_hidden_layers_override is not None:
+            if self.num_hidden_layers_override <= 0:
+                raise ValueError("MiniMax H3 Qwen3-VL num_hidden_layers_override must be positive or None.")
+            if self.num_hidden_layers_override < self.output_hidden_state_index:
+                raise ValueError("MiniMax H3 Qwen3-VL num_hidden_layers_override must build through "
+                                 f"hidden_states[{self.output_hidden_state_index}], got "
+                                 f"{self.num_hidden_layers_override}.")
 
         rope_scaling = dict(self.rope_scaling or {})
         self.mrope_interleaved = bool(rope_scaling.get("mrope_interleaved", self.mrope_interleaved))
--- a/fastvideo/models/encoders/base.py
+++ b/fastvideo/models/encoders/base.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from abc import ABC, abstractmethod
 from dataclasses import field
+from typing import Any, Generic, TypeVar
 
 import torch
 from torch import nn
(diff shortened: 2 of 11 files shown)


SolitaryThinker:
Reviewed and validated this on GB200 (torch 2.12.0+cu130, flashinfer 0.6.17 — your 0.6.8-validated path is signature/behavior compatible), and pushed two maintainer commits to this branch fixing the blocking issues found. The core of the PR held up very well under scrutiny: the layer-50 tap is preserved bit-exactly under the slim contract (test + index math both confirm early-return-after-DeepStack ≡ the old tuple entry), the fail-closed posture is consistently executed, the `Generic[TextEncoderOutputT]` base change is contained (import sweep over all 15 `TextEncoder` subclasses passes), and the trtllm/GB200 GEMM path is numerically excellent — I measured cosine 0.999999 / rel err ~0.17% vs a dequant reference across all five H3 projection shapes at m=559, and your Triton per-token-group quantizer matches a pure-torch reference bit-tight in both scale layouts.

## What was broken

**1. The default configuration crashed on the very first encode.** `FastVideoArgs.text_encoder_cpu_offload` defaults to True (and `basic_minimax_h3_t2v.py` ships with `text_encoder=True`), so the loader FSDP2-`fully_shard`s the conditioner — and `@torch.inference_mode()` on `encode_ids` then kills FSDP2's lazy unshard:

```
File torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py, in wait_for_unshard:
  with torch.autograd._unsafe_preserve_version_counter(t):
RuntimeError: Inference tensors do not track version counter.
```

The all-gather tensors are created inside the inference-mode region, and the version-counter hook can't read them. Your benchmarks ran offload-disabled, which is exactly the unexercised-default-path gap.

**2. The cutlass (sm12x/GB10) route crashed for `m % 4 != 0` — most real prompts.** FlashInfer documents that `m` "should be padded to a multiple of 4 before calling this function"; the route never padded. Reproduced by forcing `backend="cutlass"` on GB200: your own bench length m=559 fails with `cutlass gemm.can_implement(arguments) failed`, while m∈{4,128,512,560} all pass with cosine 0.9997 — so the scale layouts you build are correct and alignment is the only gap. (I also probed the layout question directly: your contiguous `(k/128, m)`/`(k/128, n/128)` MN layout is right; the `.t()`-view alternative degrades to 0.977.) The trtllm route tolerates any m and is unaffected.

## What I pushed

- `72afb9e5a` — `@torch.no_grad()` instead of `@torch.inference_mode()` on `encode_ids`. It frees the same activation memory (the −262 MiB comes from the early-return slim contract, not inference_mode's bookkeeping), is FSDP2-compatible, and as a bonus `prompt_embeds` are ordinary tensors again, so future on-the-fly-conditioning training can backprop without a clone at the stage boundary.
- `46b850d44` — pad the quantized activation rows and their `(k/128, m)` per-token scales with zeros to the next multiple of 4 on the cutlass route, and slice the padded rows off the GEMM output; trtllm stays unpadded. Adds the m=559 unit test (padding observed, MN scale layout pinned, padded rows zero, output sliced back) plus an aligned-m no-copy test — which also closes the missing cutlass twin of your trtllm layout test.

## Validation (GB200, 1 GPU)

- Unit suites on the fixed head: `test_minimax_h3_qwen3_vl_truncation.py` + `test_minimax_h3_qwen3_vl_checkpoint_fp8.py` — **24 passed** (your 22 + the 2 new cutlass tests).
- Forced-cutlass sweep with real flashinfer: **red on the unfixed head (m=559: `cutlass gemm.can_implement failed`) → green with the fix (m=559/560/512/4/128 all cosine ~0.9997 vs dequant reference)**.
- **Serialized FP8 under FSDP2 + CPU offload (the default path, finding 2)**: mirrored `TextEncoderLoader`'s exact sequence — CPU-built module, `_process_quantized_text_encoder_weights` on the runtime device, `shard_model(cpu_offload=True, reshard_after_forward=True, cuda mesh, pinned CPU)` — then forwards through the FSDP2-unsharded fp8 weight (a `DTensor` of `float8_e4m3fn`) + fp32 block scales with dynamic activation quant at m=559/560/4 under `torch.no_grad`: cosine ≥ 0.9996 vs dequant reference, repeat-consistent across reshard/unshard cycles. And the A/B: the first forward of an identical freshly-sharded module under `@torch.inference_mode()` reproduces exactly `RuntimeError: Inference tensors do not track version counter.` — confirming the no_grad commit is what unblocks the default offload path for FP8 too.
- The full pipeline with `text_encoder_cpu_offload=True` boots and generates end-to-end with the no_grad fix on an integration branch carrying this PR (it crashed with the error above before the fix).
- The branch (including these commits) merges textually clean into current main (`6d6a10be7`, post-#1731/#1735) — the earlier mergify conflict notice is stale.

## Remaining (non-blocking) suggestions

- `output_hidden_state_index == num_hidden_layers` is accepted by the config but returns the raw last-layer output where official `hidden_states[num_layers]` is post-final-norm — unreachable with H3 defaults (50 < 64); consider rejecting the boundary or applying `self.norm` there.
- `_configure_text_encoder_quantization` runs before the `supports_hf_from_pretrained` early-return, so any checkpoint whose `config.json` carries a `quantization_config` now hard-fails for every native `TextEncoder` — including ones HF `from_pretrained` could have loaded. Consider exempting the HF-passthrough path or flagging the cross-model behavior change in the description.
- For a `--override-text-encoder-safetensors` FILE override, the sibling `config.json` is trusted before the file's own safetensors metadata, which can misclassify a converted FP8 file inside a BF16 dir (and vice versa). Preferring embedded metadata for file overrides would remove the ambiguity.
- `_process_quantized_text_encoder_weights` round-trips every quantized linear through the GPU although H3's `process_weights_after_loading` is device-agnostic validation — on the offload path that's ~22 GB of avoidable PCIe traffic.
- A loader-level integration test driving `TextEncoderLoader.load_model` over a tiny serialized-FP8 fixture dir would cover the wiring (override interlock, strictness flip, post-load order) that the metadata tests don't. And since my offload validation is at the single-projection level, one full-model FP8-checkpoint + offload smoke on real weights is still worth a run when convenient.

With the pushed commits the two blocking items are closed; everything else above is follow-up material. Nice work on this one — the fail-closed discipline made it easy to validate.

SolitaryThinker (approved):

PR #1734, opened 2026-08-21
[perf] Optimize MiniMax H3 VAE decoding

Davids048:
## Purpose

Reduce MiniMax H3 video VAE decode latency for spatially tiled outputs and make the H3 GPU execution stages
attributable in Nsight Systems.

On the measured 124-frame GB200 workload, the complete change reduces the video VAE interval from 6.360 seconds to
2.775 seconds (2.29x faster) and reduces captured generation time from 20.61 seconds to 16.74 seconds.

## Changes

- Compile the H3 video and audio decoder `forward` methods through each VAE's `_compile_conditions` contract.
- Compile the fixed-shape video VAE tile projection and tile stitching operations with
  `torch.compile(mode="reduce-overhead")` so repeated tile work uses CUDA Graph replay.
- Route `MiniMaxH3VideoAttention` through FastVideo's selected dense attention backend. Selecting FlashAttention-4
  for H3 inference therefore also selects FlashAttention-4 for video VAE attention.
- Preserve main's rank-zero, pinned-CPU `decode_to_pixels` streaming path while adding a VAE-level NVTX
  range around streamed video decode.
- Add opt-in `FASTVIDEO_NVTX_PROFILE=1` ranges for H3 conditioning, DiT execution, every transformer block, video and
  audio VAE decode, temporal chunks, spatial tile splitting, tile decode, and tile stitching.
- Log the selected attention backend and CUDA device UUID per worker rank when NVTX profiling is enabled.
- Add contract tests for compile selection, attention dispatch, NVTX range behavior, tiled decode ranges, and
  pinned-CPU streaming decode.

## Test Plan

```bash
python -m pytest \
  fastvideo/tests/vaes/test_minimax_h3_video_vae_streaming.py \
  fastvideo/tests/stages/test_minimax_h3_vae_streaming.py \
  fastvideo/tests/vaes/test_minimax_h3_vae_compile.py \
  fastvideo/tests/contract/test_profiler_regions.py \
  fastvideo/tests/worker/test_gpu_worker.py -q
```

Profile one generation with `FASTVIDEO_NVTX_PROFILE=1`, `FASTVIDEO_ATTENTION_BACKEND=FLASH_ATTN`, and
`FASTVIDEO_FA4=1`, after three warmup generations.

## Test Results

The profiling workload uses one NVIDIA GB200 GPU, `MiniMaxAI/MiniMax-H3` revision
`9ac0dd7aabc2c651fcf0ace4c00b2bffd9c8c8a6`, 124 frames at 1344x768 and 24 FPS, five DiT evaluations, three warmup
generations, spatial VAE tiling, and no model offload.

| Configuration                                      | Captured generation | Video VAE interval |
| -------------------------------------------------- | ------------------: | -----------------: |
| Default-mode tiled decoder with PyTorch SDPA       |             20.61 s |         6.360306 s |
| Reduce-overhead tiled decoder with PyTorch SDPA    |             17.10 s |         3.168423 s |
| Reduce-overhead tiled decoder with FastVideo FA4   |             16.74 s |         2.774891 s |

The controlled attention-backend comparison reduces the video VAE interval by 12.42%, from 3.168423 seconds with
PyTorch SDPA to 2.774891 seconds with FastVideo FlashAttention-4. The complete optimized configuration reduces the
video VAE interval by 56.37% relative to the default-mode tiled baseline.

Nsight Systems records 399 CUDA Graph replay intervals in the optimized run: 196 decoder forwards, 196 tile
projections, and seven tile-stitching calls. The generated MP4 contains 124 frames at 1344x768, runs at 24 FPS, and
has a duration of 5.166667 seconds.

## Checklist

- [ ] `pre-commit run --all-files`
- [x] Focused tests added for the changed contracts
- [x] GPU memory behavior validated with spatial tiling retained
- [ ] SSIM regression test

### DIFF
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -19,7 +19,12 @@ from fastvideo.attention.backends.abstract import (
 from fastvideo.logger import init_logger
 
 logger = init_logger(__name__)
-logger.info("Using FlashAttention-%s backend", fa_version)
+# Every worker records the loaded FlashAttention implementation so a
+# distributed profiling log contains one backend receipt per rank.
+logger.info("Worker %s Using FlashAttention-%s backend",
+            os.environ.get("RANK", "0"),
+            fa_version,
+            local_main_process_only=False)
 
 # FP4 FA4 support: quantize Q/K to NVFP4 E2M1 for block-scaled MMA on Blackwell.
 # Requires: flash-attention-fp4, flashinfer, cutlass-dsl. Enable via nvfp4_fa4=True kwarg.
--- a/fastvideo/attention/backends/flash_attn.py
+++ b/fastvideo/attention/backends/flash_attn.py
@@ -228,7 +228,15 @@ class FlashAttentionImpl(AttentionImpl):
     ) -> None:
         self.causal = causal
         self.softmax_scale = softmax_scale
-        self.nvfp4_fa4 = extra_impl_args.get("nvfp4_fa4", False) or os.environ.get("FASTVIDEO_NVFP4_FA4", "0") == "1"
+        # An explicit ``nvfp4_fa4`` impl arg wins over the process-wide
+        # FASTVIDEO_NVFP4_FA4 env opt-in, so precision-sensitive layers (e.g.
+        # the FP32-pinned H3 VAE attention) can force-disable FP4 Q/K
+        # quantization while the DiT keeps it. When the arg is absent the env
+        # keeps its previous semantics.
+        nvfp4_fa4 = extra_impl_args.get("nvfp4_fa4")
+        if nvfp4_fa4 is None:
+            nvfp4_fa4 = os.environ.get("FASTVIDEO_NVFP4_FA4", "0") == "1"
+        self.nvfp4_fa4 = bool(nvfp4_fa4)
         if self.nvfp4_fa4:
             cap = torch.cuda.get_device_capability()
             assert cap in [(10, 0), (10, 3)], (f"NVFP4 FA4 requires Blackwell (sm100a/sm103a), got sm{cap[0]}{cap[1]}")
--- a/fastvideo/envs.py
+++ b/fastvideo/envs.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
     NVCC_THREADS: str | None = None
     CMAKE_BUILD_TYPE: str | None = None
     VERBOSE: bool = False
+    FASTVIDEO_NVTX_PROFILE: bool = False
     FASTVIDEO_TORCH_PROFILER_DIR: str | None = None
     FASTVIDEO_TORCH_PROFILER_RECORD_SHAPES: bool = False
     FASTVIDEO_TORCH_PROFILER_WITH_PROFILE_MEMORY: bool = False
@@ -220,6 +221,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # Use dedicated multiprocess context for workers.
     "FASTVIDEO_WORKER_MULTIPROC_METHOD":
     lambda: os.getenv("FASTVIDEO_WORKER_MULTIPROC_METHOD", "spawn"),
+
+    # Emit lightweight NVTX ranges for external profilers such as Nsight Systems.
+    "FASTVIDEO_NVTX_PROFILE":
+    lambda: os.getenv("FASTVIDEO_NVTX_PROFILE", "0") != "0",
 
     # Enables torch profiler if set. Path to the directory where torch profiler
     # traces are saved. Note that it must be an absolute path.
--- a/fastvideo/models/dits/minimax_h3.py
+++ b/fastvideo/models/dits/minimax_h3.py
@@ -25,6 +25,7 @@ from fastvideo.layers.quantization import QuantizationConfig
 from fastvideo.layers.visual_embedding import Timesteps
 from fastvideo.models.dits.base import BaseDiT
 from fastvideo.platforms import AttentionBackendEnum
+from fastvideo.profiler import nvtx_range
 from fastvideo.utils import get_compute_dtype
 
 MINIMAX_H3_MODALITY_NUM = 3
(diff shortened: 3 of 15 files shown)


SolitaryThinker:
Reviewed this on GB200 (torch 2.12.0+cu130) against base `c4ad4227c`, and pushed three maintainer commits to this branch that fix the blockers found — the profiling/observability work, the selective decoder `_compile_conditions` hook, and the attention dispatch all checked out and are untouched.

## What was broken

**1. Tiled `decode()`/`encode()`/`encode_pixels()` crashed on CUDA on the first call (blocker).** With `mode="reduce-overhead"`, each `_stitch_tiles` output is a CUDA-graph static buffer, but the non-streaming consumers (`_decode` via `torch.cat(list(self._decode_chunks(z)))`, `_encode`, `_encode_pixels`) hold chunk/clip results across the next `_stitch_tiles` replay, which overwrites the pooled storage. Reproduced on this PR head (`dca423fd3`) with a random-weight VAE, 2x2 spatial tiles, 2 temporal chunks (i.e. any real video; encode fails at >17 frames):

```
RuntimeError: Error: accessing tensor output of CUDAGraphs that has been overwritten by a
subsequent run. Stack trace: ... line 722, in _stitch_tiles: return torch.cat(result_rows, dim=-2)
```

Your profiling runs don't hit it because the streaming stage (`decode_to_pixels`) copies each chunk to the pinned CPU buffer before the next chunk decodes, and the new tests either mock `_stitch_tiles` or run untiled/CPU. Note this also silently pulled the *encode* path (preprocessing) into a decode-titled PR.

**2. The two unconditional `@torch.compile(mode="reduce-overhead")` decorators bypassed the opt-in compile contract this same PR establishes for the decoder** — and they regressed the defaults. Measured on GB200 (124f 1344x768, FastH3 5-step, n=3 after warmup, A/B differing only in this PR):

| config | video decode stage (SP-4) | video decode stage (1x) |
|---|---|---|
| pre-PR eager | 8.18 s | 8.81 s |
| this PR at defaults (unconditional tile compile) | **11.02 s (+2.8 s regression)** | — |
| this PR + `enable_torch_compile_vae` opt-in | **6.05 s** | **4.86 s** |

So the hardcoded compile made default decode *slower*, while the opt-in path (which this PR makes effective for H3 — nice) is the best measured config. The defaults also permanently reserved cudagraph pool memory (toy tiled geometry: 132.5 MiB peak allocated vs 38.8 on main, ~3.4x; reserved 192 vs 60 MiB; scales with resolution/frames — on exactly the memory-constrained GPUs spatial tiling exists for), required a working inductor+triton toolchain even when compile was never requested, and `dynamic=False` recompiles per resolution/tile grid.

**3. `FASTVIDEO_NVFP4_FA4=1` silently FP4-quantized VAE attention.** `FlashAttentionImpl` read the env with `extra_impl_args.get("nvfp4_fa4", False) or env`, so the DiT opt-in also quantized Q/K inside this FP32-pinned VAE with no opt-out.

## What I pushed

- `b83232a94` — clone the stitched canvas at the tile-driver returns so cudagraph-owned storage never escapes (`_decode_clip`/`_encode_clip`), plus the missing CUDA regression test: real tiled `decode()` (2 chunks, 2x2 tiles) + `encode()`/`encode_pixels()` (2 clips) with `_stitch_tiles` unmocked, asserting bitwise repeat-consistency and value parity vs the eager helpers. Red on `dca423fd3` (the error above), green with the fix — both verified on GB200.
- `ebb5ab2b3` — gate both tile-helper compiles behind the existing opt-in: they now compile in `prepare_for_compile()`, which `ComposedPipelineBase._maybe_compile_pipeline_module` already invokes for the `vae` module right before `_compile_conditions` when `enable_torch_compile_vae` is set. Defaults stay fully eager (and the F1 clone only runs when compiled, so default allocation behavior is untouched); the opt-in keeps exactly the reduce-overhead behavior you shipped, including its measured win. Tests reworked accordingly (defaults-stay-eager, opt-in installs cudagraph helpers + is idempotent, driver ownership contract in both modes).
- `4c17beba2` — an explicit `nvfp4_fa4` impl arg now wins over the `FASTVIDEO_NVFP4_FA4` env in both directions (env semantics unchanged when the arg is absent; no existing caller passes it), and the VAE constructs its impl with `nvfp4_fa4=False`. New unit tests pin the precedence.

## Validation (GB200, 1 GPU, torch 2.12.0+cu130)

- Regression test: **red on `dca423fd3` → green at `b83232a94`** (exact cudagraph-overwrite error, first tiled decode).
- Final head: the PR's own test plan + reworked/new tests — **39 passed** (`test_minimax_h3_vae_compile.py`, `test_minimax_h3_video_vae_streaming.py`, `stages/test_minimax_h3_vae_streaming.py`, `contract/test_profiler_regions.py`, `worker/test_gpu_worker.py`); nvfp4 precedence tests **4 passed** under FA4 (`FASTVIDEO_FA4=1`), file skips cleanly where no flash-attention package exists.
- **Defaults now bitwise-match pre-PR main (`c4ad4227c`)**: tiled decode, streaming decode, untiled fp16 + fp32 decode, and tiled encode all `max_abs=0.0e+00`, with identical peak memory (38.8 MiB allocated / 60.0 MiB reserved on both sides) — i.e. the crash, the +2.8 s regression, and the pool-floor regression are gone at defaults.
- Opt-in probe: `prepare_for_compile()` + tiled decode/encode repeat-consistent, decode within 2e-3 of eager under fp16 autocast (inductor fusion tolerance), encode bitwise-equal.
- Earlier full-suite sweep (adjacent `tests/stages` + `tests/worker` + `tests/contract` + api/ops/train/entrypoints/workflow/dataset) had identical failure sets on PR head and main — environment-only, none PR-caused.

## Remaining notes (non-blocking, no action pushed)

- Resolving the backend against `torch.bfloat16` while the module is `_keep_in_fp32_modules` means direct `decode()`/`encode()` callers without the stage's fp16 autocast drop from fp32 SDPA to bf16 FA when flash_attn is installed (the impl's documented cast). The shipped pipeline autocasts, so fine there — worth a code comment for direct-API callers. FWIW with the selector resolving to TORCH_SDPA I measured untiled decode bitwise-identical to main, so the dispatch itself is numerics-preserving.
- `flash_attn.py`'s per-rank receipt now logs at import time in every process, and `RANK` may not be set yet in some launchers; consider moving it to worker init.
- `nvtx_range` re-reads the env per call even when disabled — a cached module-level flag would be free in the per-block loop.
- #1731 (now merged) touches the same `minimax_h3_denoising.py` hunk this PR rewrites. I checked: this branch merges textually clean into current main (`6d6a10be7`), and the merged file interleaves correctly (the `vsa_dense_first_n` block sits above `profiler_region("inference_denoising")`, the per-step sparsity logic inside the loop, `cudagraph_mark_step_begin` retained) — no rebase needed, but worth one look after CI runs the merged result.

With the pushed commits I think this is good to go — the NVTX/profiling surface, the attention dispatch, and the (now opt-in end to end) compile path all validate clean.

PR #1735, opened 2026-08-21
[perf] Add opt-in MiniMax-H3 Sol-Engine fusions

H1yori233:
## Purpose

Add opt-in, inference-only Triton fusions for MiniMax-H3 while leaving the default eager path unchanged.

The operator designs and routing approach are ported and adapted from NVlabs Sana Sol-Engine:

- [`models/minimax_h3/GB200/fusions.py`](https://github.com/NVlabs/Sana/blob/sol-engine/models/minimax_h3/GB200/fusions.py)
- [`models/minimax_h3/GB200/fusion_install.py`](https://github.com/NVlabs/Sana/blob/sol-engine/models/minimax_h3/GB200/fusion_install.py)

This port integrates those kernels with FastVideo's native MiniMax-H3 model, inference fallback policy, tensor contracts, and tests.

## Changes

- Add three independently selectable fusion groups:
  - `modulate`: RMSNorm plus row-indexed scale/shift, and residual/gate plus the following RMSNorm/modulation
  - `qknorm_rope`: per-head Q/K RMSNorm plus MiniMax-H3 partial RoPE
  - `swiglu`: packed value-first SwiGLU
- Add `FASTVIDEO_MINIMAX_H3_FUSIONS`:
  - empty, `0`, or `none`: keep eager execution
  - `1` or `all`: enable all three groups
  - comma-separated names: enable a subset
- Keep the kernels CUDA-only and inference-only, with eager fallback for autograd, CPU, and Dynamo/`torch.compile` capture.
- Reject unknown fusion names and unsupported tensor contracts explicitly.
- Add routing, contract, BF16 operator-parity, and FA4 block-integration tests.

The default behavior remains unchanged unless the environment variable is explicitly enabled.

### Operator benchmark

Measured on GB200 with production-shaped BF16 inputs using the fully fused Sol-Engine-style implementation:

| Fusion | Eager | Fused | Speedup |
|---|---:|---:|---:|
| RMSNorm/modulation | 1.6514 ms | 0.2147 ms | 7.69x |
| QK RMSNorm + partial RoPE | 1.8476 ms | 0.3885 ms | 4.76x |
| SwiGLU | 0.6717 ms | 0.1574 ms | 4.27x |

### End-to-end benchmark

Official MiniMax-H3 T2VA workload: 4x GB200, SP=4, TP=1, dense FlashAttention-4, no VSA, local checkpoint, 1344x768, 124 frames, 50 steps, seed 0, and no `torch.compile`. Model loading is excluded. Each arm used one warmup request followed by one measured request.

| Configuration | Generation | Request E2E including save | Change |
|---|---:|---:|---:|
| Eager | 66.3046 s | 66.7215 s | - |
| All Sol-style fusions | 56.9235 s | 57.3841 s | generation -14.1484%; E2E -13.9946% |

This is 1.1648x generation speedup and 1.1627x request-E2E speedup.

## Known numerical behavior

The Sol-Engine-style kernels retain reductions and intermediate arithmetic in FP32 registers until their final BF16 stores. FastVideo's eager PyTorch path materializes BF16 tensors between operators. The opt-in fused path therefore matches within the tested operator tolerances, but it is intentionally not bit-identical.

In one same-seed A/B, decoded-frame mean SSIM was 0.7403. Manual inspection found the output acceptable, but this does not pass a strict same-seed SSIM regression. This PR does not claim eager-output SSIM parity, and the eager path remains the default.

An exactness-oriented alternative was investigated but is not included. It uses PyTorch RMSNorm to preserve the CUDA reduction order and restores BF16 materialization at the residual/update, modulation, RoPE-product, and SwiGLU activation boundaries. That path produced bit-exact decoded RGB output (`SSIM=1.0`, `MAE=0`), but reduced the generation improvement from 14.15% to 7.55%:

| Configuration | Generation | Request E2E | Improvement vs eager |
|---|---:|---:|---:|
| Fully fused Sol-style path | 56.9235 s | 57.3841 s | generation 14.1484%; E2E 13.9946% |
| Exactness alternative, not included | 61.2965 s | 61.7164 s | generation 7.5531%; E2E 7.5015% |

## Test Plan

```bash
python -m pytest \
  fastvideo/tests/transformers/test_minimax_h3_fusion_routing.py \
  fastvideo/tests/transformers/test_minimax_h3_modulation_fusion.py \
  fastvideo/tests/transformers/test_minimax_h3_qknorm_rope_fusion.py \
  fastvideo/tests/transformers/test_minimax_h3_swiglu_fusion.py -q

pre-commit run --all-files
git diff --check
```

## Test Results

- Local focused suite: `27 passed, 7 skipped` (CUDA cases skipped locally).
- GB200 FA4 focused suite: `34 passed, 8 warnings in 6.12s`; managed run `20260821t071546z-a63ccce7-c71f5798` succeeded with exit code 0.
- The GB200 run explicitly selected `FLASH_ATTN`, set `FASTVIDEO_FA4=1`, enabled all fusions, and unset `FASTVIDEO_H3_VSA_PROBE`; it did not load a checkpoint.
- `pre-commit run --all-files`: passed.
- `git diff --check`: passed.
- Four-GPU dense-FA4/no-VSA eager and fused end-to-end runs completed successfully for the performance table.
- Strict same-seed SSIM parity is not claimed for the fully fused path; see Known numerical behavior.

<details>

https://github.com/user-attachments/assets/21c20577-ed16-494c-8e2a-d11ceb1bcdd7

https://github.com/user-attachments/assets/0bf501cd-5036-4611-9916-1ac901103f7d

https://github.com/user-attachments/assets/92715a77-1d17-425c-b553-440dc03b49c9

https://github.com/user-attachments/assets/d635e37d-45c1-4e90-9834-e298423c8301

https://github.com/user-attachments/assets/7063589d-755a-4ba2-80fa-d406b9ac72f2

https://github.com/user-attachments/assets/26ada18f-68dd-4218-bf83-6f10f19dfaf7

https://github.com/user-attachments/assets/dc25e761-5b04-4c2c-8fb2-a6c7233789c1

https://github.com/user-attachments/assets/9ced748c-62ff-4fae-96e0-66c0c39823e8

</details>

## Checklist

- [x] I updated documentation if needed (no documentation change is required)

- [ ] I verified SSIM regression tests pass (known opt-in numerical trade-off documented above)
- [ ] I updated the support matrix if adding a new model (not applicable; no new model is added)

### DIFF
--- a/fastvideo/envs.py
+++ b/fastvideo/envs.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
     FASTVIDEO_TRACE_FUNCTION: int = 0
     FASTVIDEO_ATTENTION_BACKEND: str | None = None
     FASTVIDEO_FA4: bool = False
+    FASTVIDEO_MINIMAX_H3_FUSIONS: str = ""
     FASTVIDEO_WORKER_MULTIPROC_METHOD: str = "spawn"
     FASTVIDEO_TARGET_DEVICE: str = "cuda"
     MAX_JOBS: str | None = None
@@ -216,6 +217,13 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # (FA4's backward asserts sm90+ and its pack_gqa fails to JIT there).
     "FASTVIDEO_FA4":
     lambda: os.getenv("FASTVIDEO_FA4", "0") != "0",
+
+    # Opt-in MiniMax-H3 inference-only Triton fusions adapted from the
+    # NVlabs/Sana Sol-Engine implementation. Accepts `all`, `1`, or a
+    # comma-separated subset of `modulate,qknorm_rope,swiglu`. An empty value
+    # (the default), `0`, or `none` keeps the eager implementation.
+    "FASTVIDEO_MINIMAX_H3_FUSIONS":
+    lambda: os.getenv("FASTVIDEO_MINIMAX_H3_FUSIONS", ""),
 
     # Use dedicated multiprocess context for workers.
     "FASTVIDEO_WORKER_MULTIPROC_METHOD":
--- a/fastvideo/models/dits/minimax_h3.py
+++ b/fastvideo/models/dits/minimax_h3.py
@@ -10,6 +10,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
+from fastvideo import envs
 from fastvideo.attention import DistributedAttention
 from fastvideo.attention.layer import DistributedAttention_VSA
 from fastvideo.attention.selector import get_attn_backend
@@ -24,11 +25,39 @@ from fastvideo.layers.mlp import MLP
 from fastvideo.layers.quantization import QuantizationConfig
 from fastvideo.layers.visual_embedding import Timesteps
 from fastvideo.models.dits.base import BaseDiT
+from fastvideo.models.dits.minimax_h3_fusions import (
+    fused_qknorm_rope,
+    fused_residual_gate_rmsnorm_modulate,
+    fused_rmsnorm_modulate,
+    minimax_h3_swiglu,
+)
 from fastvideo.platforms import AttentionBackendEnum
 from fastvideo.utils import get_compute_dtype
 
 MINIMAX_H3_MODALITY_NUM = 3
 _CFG = MiniMaxH3Config()
+_MINIMAX_H3_FUSION_NAMES = frozenset({"modulate", "qknorm_rope", "swiglu"})
+
+
+def _enabled_minimax_h3_fusions(value: str | None = None) -> frozenset[str]:
+    """Parse the independently switchable inference fusion set."""
+    raw = envs.FASTVIDEO_MINIMAX_H3_FUSIONS if value is None else value
+    normalized = raw.strip().lower()
+    if normalized in {"", "0", "none"}:
+        return frozenset()
+    if normalized in {"1", "all"}:
+        return _MINIMAX_H3_FUSION_NAMES
+    enabled = frozenset(item.strip() for item in normalized.split(",") if item.strip())
+    unknown = enabled - _MINIMAX_H3_FUSION_NAMES
+    if unknown:
+        supported = ",".join(sorted(_MINIMAX_H3_FUSION_NAMES))
+        raise ValueError(f"Unknown MiniMax H3 fusion(s) {sorted(unknown)}; expected a subset of {supported}.")
+    return enabled
+
+
+def _can_run_minimax_h3_fusion(tensor: torch.Tensor) -> bool:
+    """Triton kernels are inference-only and stay outside Dynamo capture."""
+    return tensor.is_cuda and not torch.is_grad_enabled() and not torch.compiler.is_compiling()
 
 
 class MiniMaxH3RotaryPosEmbed(nn.Module):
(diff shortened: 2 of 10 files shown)


SolitaryThinker:
Thanks — this is a clean, well-tested port: opt-in by env var, strict kernel wrappers with the model owning fallback policy, and the default path really is untouched (I verified a seeded `MiniMaxH3TransformerBlock` forward is bitwise identical to `main` with the env unset, on GB200/bf16/SDPA). Kernel parity, routing, and contract tests all pass on GB200, and I measured 2.9x/3.8x/3.9x kernel-level wins at production dims, consistent with your table.

I found one real bug and pushed fixes for it plus a few small hardening items directly to this branch (maintainer edit, two additive commits — nothing of yours was rewritten):

1. **ac98869 — int32 overflow in `_qknorm_partial_rope_kernel` (fixed).** `row` was never cast to int64, so `row * head_dim` wrapped at 2^31 total elements. Your other two kernels in this PR already do `tl.program_id(0).to(tl.int64)`; this one now matches. Repro on GB200, before the cast: `fused_qknorm_rope(torch.randn(1, 8_500_000, 2, 128, bf16, cuda), ...)` (2.176e9 elements) → `CUDA error: an illegal memory access was encountered`, while `seq=8_300_000` (just under 2^31) matched eager. After the cast the same over-2^31 shape matches eager (head and tail rows, 2e-2 tol). For H3 (56 heads x 128 head_dim) the boundary was `batch*seq >= 299,593` tokens per rank — a 15s 1344x768 clip at SP=1 is already at 74% of that line, so this was reachable and crashed rather than degraded. The commit adds a GPU regression test at the over-2^31 shape (skips below 20 GiB free).

2. **cbab605 — routing/observability hardening.**
   - A positive-engagement test: the FA4 block parity test also passes if `_can_run_minimax_h3_fusion` silently degrades to always-eager (both sides then match exactly). The new test monkeypatches the four fused symbols with counters and asserts 1/1/2/1 calls in one CUDA `inference_mode` block forward, and that a grad-enabled forward leaves the counters unchanged.
   - One-time `logger.info` of the resolved fusion set at model init, a warning when the set is requested but Triton is unavailable, and a `prepare_for_compile` warning (the pipeline calls this hook right before compiling) that `torch.compile` capture makes the fusions inert inside compiled block forwards — previously the flag could look active with no runtime signal either way.
   - `_can_run_minimax_h3_fusion` now also checks Triton availability, so an enabled fusion on a CUDA build without a working Triton falls back to eager instead of hitting the strict wrappers' `RuntimeError` mid-forward.
   - Documented the `[0, table_rows)` index contract on the modulation wrappers (eager `index_select` raises on out-of-range; the kernel can't without a device sync — in-model callers are safe by construction).

Validation on 1x GB200 (bf16): the new overflow regression test fails with the illegal memory access on the pre-fix kernel and passes on this branch; the full fusion suite is green (35 passed, 1 FA4-environment skip — up from 33/1 with the two new tests); the default path with the env unset is still bitwise identical to `main`.

One question for the maintainers on the numerics: the disclosed same-seed SSIM 0.7403 is a real output change, but it's opt-in and default-off, and the PR is upfront about it — I'd just ask that the env var + caveat get a user-facing mention (cookbook or H3 example), and that the bit-exact variant you benchmarked stays on the table as a possible `exact` mode later. Never enable this env during SSIM/parity work.

SolitaryThinker (approved):

PR #1740, opened 2026-08-22
[perf] Add fused NVLink all-to-all for Ulysses

shaoxiongduan:
## Summary

- Add an opt-in fused NVLink all-to-all for conventional Ulysses sequence parallelism. The CUDA kernel combines the local layout transpose with peer writes instead of materializing NCCL's contiguous send layout first.
- Route through the fused path only for world sizes 2, 4, 6, or 8 when every rank is in the NCCL load-store-accessible team. SP1 remains allocation-free; packed SP and non-LSA topologies keep the existing NCCL implementation.
- Make every construction, setup, growth, per-call, and teardown decision group-atomic. Unsupported layouts, dtypes, capture, oversized operands, rank-local failures, and lifecycle mismatches fall back unanimously.

## Safety and compatibility

- Reserve one NCCL LSA barrier slot per launched CTA and index it with `blockIdx.x`.
- Initialize NCCL 2.29 device-API structures with their required initializers and clean up in device-communicator, window, allocation order.
- Bound the persistent registered window to 1 GiB per rank.
- Keep the raw pybind path eager-only. `torch.compile` regions decline before reading tensor state and use the existing compiler-visible NCCL collective.
- Build the kernel only on CUDA when NCCL 2.29 device headers and a runtime library are available. ROCm and older/minimal builders produce a wheel without this optional surface.
- Version the kernel package as 0.3.3. Merging the version change triggers the existing wheel-publish workflow; the root 0.3.2 pin should move to 0.3.3 in the normal follow-up after the wheels are available.

## Validation

- `pre-commit` on every changed file: passed.
- CUDA 13.0 / NCCL 2.29.7 / GB200 `sm_100a` wheel build: passed; `fastvideo-kernel==0.3.3`, fused ops importable.
- Optional build gate: `OFF` config excludes the source; `ON` config detects the required NCCL API and builds it.
- CPU layout and compiler-decline contracts: 9 passed.
- GPU parity/stress: SP2 and SP4 passed exact forward, inverse, both backward directions, three dtypes, growth, and 100 rounds at 36 CTAs. SP8 is in-tree but was skipped on the four-GPU validation node.
- Rank-local fault matrix: 20 passed across SP2/SP4, covering capability, configuration, allocation, post-registration, post-device-communicator, teardown, capacity, capture, layout, and lifecycle divergence.

The earlier per-collective speedup measurements predate the all-rank safety protocol and are intentionally not restated here. No published `h3_perf.md` result includes this PR.


SolitaryThinker (changes requested):
I do not think this head is safe to merge yet. The speedup is promising, but the current distributed failure modes can produce silent corruption or a collective hang.

Blocking issues:

1. Every CUDA CTA aliases the same NCCL LSA barrier. `ulysses_all_to_all.cu` sets `lsaBarrierCount = 1` but launches up to 36 CTAs, while `ulysses_all_to_all.cuh` passes barrier index `0` from every CTA. NCCL's device-API contract requires a distinct barrier slot per cooperative group/CTA (normally indexed by `blockIdx.x`). Sharing one generation counter across independently scheduled CTAs is unsupported. Please allocate `nCTAs` barriers and index them per CTA, then add a long multi-CTA stress gate.

2. Fallback is not group-atomic. Only the initial `_can_attempt()` decision is voted. Helper construction, allocation/registration, dtype/layout/device/capture guards, and capacity growth can diverge rank-locally after that vote. One rank can enter ordinary NCCL while peers enter the fused device barrier, which mismatches collective order and hangs. Every build/growth/per-call decision needs an all-rank agreement protocol, including failure injection after allocation and registration. The documented oversized fallback is not implemented; the current path grows without a cap.

3. The target environment uses NCCL 2.29.7, where `ncclDevCommRequirements` must be initialized with `NCCL_DEV_COMM_REQUIREMENTS_INITIALIZER`. The current code zeroes the structure, leaving required size/magic/version fields unset. Failure after window registration also lacks RAII cleanup.

4. The call is not compiler-visible. Base-H3's measured route uses 52 `fullgraph=True` regions from #1741, but this path executes a tensor `.item()`, mutable Python lifecycle state, and raw pybind calls without a `torch.library.custom_op`/fake implementation. It therefore needs a pre-sized, compiler-safe boundary before it can be included in that grid.

5. Packaging/build optionality is incomplete. The current public and benchmark environments still load `fastvideo-kernel==0.3.2` without `comm_ops`, so the feature cannot engage. NCCL device headers/sources are added unconditionally, including the documented ROCm build path. Please bump/publish the kernel dependency and gate the build by backend/NCCL capability.

6. Exact head contains no tests: the parity and fault-injection files were removed in `c454cfb3`. A distributed optimization with persistent registered memory needs in-tree SP2/SP4/SP8 parity, all-rank engagement assertions, growth/fault/teardown tests, and topology fallback coverage.

Grid applicability after those fixes:

- 1x should remain an allocation-free no-op.
- SP4 within one tray can benefit when all 4 ranks arm.
- PR #1739 packed SP bypasses this implementation, so gains are not additive.
- Preview VSA must keep packed SP disabled; conventional Ulysses is the only applicable path.
- SP8 over two trays is expected to fall back unanimously when the LSA team is only 4 ranks; engagement must be 0/8 or 8/8, never mixed.

No published `h3_perf.md` timing includes this PR. Please keep it out of the serving grid until the barrier, group-agreement, NCCL-initialization, compiler, packaging, and test gates above are satisfied.

SolitaryThinker:
Rebased and repaired the PR after validating each requested-change item.

- Old head: `61ab307aaf469969d2b72d36b39a4d835a19867f`
- New base: `604e0205a45c7a3ea4accae3a6ff4b54a60793c7`
- New head: `c27841439a0f5396a97cf05a5cc31e4a254f7b6d`

The six blockers were valid and are addressed as follows:

1. The device communicator now reserves 36 LSA barriers and each CTA uses `blockIdx.x`; the test keeps all 36 CTAs crossing both barriers for 100 rounds.
2. Helper capability, allocation, window registration, device-communicator creation, complete call signatures, growth, capture, capacity, lifecycle, and teardown now have all-rank decisions over the CPU control group. Registered memory is capped at 1 GiB per rank.
3. NCCL 2.29 requirements/properties use their required initializers. Setup is split into locally allocatable and collective phases, and cleanup follows device communicator -> registered window -> allocation.
4. The fused pybind route is explicitly eager-only. During `torch.compile`, it declines before tensor-state reads and leaves the existing compiler-visible NCCL route intact.
5. The source is optional on CUDA/NCCL capability and excluded from ROCm/minimal builds. `fastvideo-kernel` is versioned as 0.3.3, which triggers the existing publish workflow after merge; the root pin remains at the published 0.3.2 until the normal follow-up once 0.3.3 wheels exist.
6. In-tree tests now cover W=2/4/6/8 layouts, SP2/SP4/SP8 runtime routing, exact forward/inverse/both backwards across bf16/fp16/fp32, unanimous engagement, growth, multi-CTA stress, and 20 rank-local failure cases including teardown.

Validation on one 4x GB200 node, CUDA 13.0, NCCL 2.29.7:

- optional CMake OFF/ON configure: passed
- `fastvideo-kernel==0.3.3` sm100a wheel build/import: passed
- CPU layout + compile-decline contracts: 9 passed
- fused SP2/SP4 parity and stress: 2 passed; SP8 skipped on the four-GPU node
- rank-local failure matrix: 20 passed in 424.25 s
- pre-commit on all changed files after the final rebase: passed

Applicability is unchanged: SP1 is an allocation-free no-op, packed SP from #1739 bypasses this path, fullgraph regions from #1741 use NCCL, and a two-tray SP8 group with only four LSA peers falls back unanimously. The previous speed table was removed because it predates the new all-rank safety protocol; no `h3_perf.md` result includes this PR.

SolitaryThinker (approved):
Re-reviewed at `c27841439a0f5396a97cf05a5cc31e4a254f7b6d`. The barrier allocation, group-atomic fallback/lifecycle, NCCL 2.29 initialization, compiler fallback, optional packaging, and restored distributed-test blockers are resolved. The CUDA 13 / NCCL 2.29.7 GB200 build and focused SP2/SP4 parity, 36-CTA stress, and 20-case rank-local fault gates pass.

The 0.3.3 wheel publication and root-pin bump remain the established post-merge release sequence; the feature safely stays unavailable against the current 0.3.2 wheel. The refreshed PR body no longer claims pre-safety-protocol timing.

PR #1745, opened 2026-08-22
[perf] Keep odd MiniMax-H3 VSA tiles on sm100a

SolitaryThinker:
## Purpose

The MiniMax-H3 sm100a block-sparse kernel assigns adjacent query-tile pairs to each CTA. A tile-64 request with an odd logical tile count therefore fails the kernel support gate and falls back to Triton. Long Preview shapes such as 345 frames commonly have that odd geometry, leaving the faster CUDA path unused.

This change appends one internal, zero-valid partner tile for eligible no-grad sm100a calls. The trained score search, block mask, gate-compress branch, metadata, gradients, fallbacks, and returned packed sequence remain on the original logical tiles.

## Changes

- Add one transport-only partner to odd tile-64, no-grad sm100a requests.
- Give the partner an all-false query/key mask and zero variable block size.
- Trim the partner before every Triton fallback and before returning sm100a output.
- Preserve the existing logical geometry for dense/sparse score selection and gate-compress.
- Clear stale padding when a builder-owned buffer is reused across same-sized but different tile geometries.
- Add CPU routing/shape/buffer tests plus a real GB200 sm100a-vs-Triton oracle.

The feature remains behind the existing `FASTVIDEO_VSA_SM100A=1` opt-in. Tile-256, even-tile, grad-enabled, and default-off paths retain their previous behavior.

## Test Plan

```bash
pre-commit run --all-files
python -m pytest -q fastvideo/tests/attention/test_vsa_h3_sm100a_route.py
```

## Test Results

- Current composed head `286203d9e` (`main`/#1741 plus this change): **15 passed** on GB200, including the real odd-tile sm100a route compared with the Triton-64 numerical oracle (job 2997).
- Full pre-commit suite: passed.
- The real-kernel gate requires equal logical output shape and `torch.testing.assert_close(..., atol=0.04, rtol=0.02)` against Triton.
- Preview FastH3 F4, 768x1344x345, four DiT forwards, one excluded warmup plus three timed repeats:
  - 1x: **54.809 s -> 47.212 s** end to end; **39.998 s -> 29.699 s** denoising.
  - SP-4: **18.236 s -> 15.468 s** end to end; **11.428 s -> 9.107 s** denoising.

F4 changes floating-point operation order through its separate fusion opt-in and remains a report-only performance profile. This PR does not broaden that claim; it specifically removes the odd-tile fallback while retaining the established sm100a numerical envelope.

## Checklist

- [x] I ran pre-commit on the exact composed head
- [x] I added default-off, grad, even/odd, fallback, stale-buffer, and hardware coverage
- [x] I verified the actual sm100a extension on GB200
- [x] I kept training code and configuration out of the change

### DIFF
--- a/fastvideo/attention/backends/video_sparse_attn_h3.py
+++ b/fastvideo/attention/backends/video_sparse_attn_h3.py
@@ -34,9 +34,13 @@ for the tile-64 FORWARD only: ``FASTVIDEO_VSA_SM100A=1`` sends no-grad
 forwards through the sm_100a CUDA block-sparse kernel
 (``fastvideo_kernel.block_sparse_attn_sm100a``, upstream PR #1719 plus
 our per-q-tile ``q2k_num`` fix) when the extension is built, the device
-is sm_100, and the geometry qualifies; grad-tracking forwards and every
-backward stay on Triton unchanged. If the env is set but a precondition
-fails, the route logs one warning and falls back.
+is sm_100, and the geometry qualifies. The CUDA kernel assigns adjacent
+pairs of query tiles to CTAs, so an odd logical tile count receives one
+internal, zero-valid partner tile for the no-grad call only. Score search,
+the trained mask, gate-compress, and the returned packed sequence remain on
+the original logical tiles. Grad-tracking forwards and every backward stay
+on Triton unchanged. If the env is set but a precondition fails, the route
+logs one warning and falls back.
 """
 
 import functools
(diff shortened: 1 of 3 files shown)


SolitaryThinker:
**S — Scope:** Validated exact head `82a5b0db692eac7df441965ce5559d56655f4b39` on NVIDIA GB200 (sm_100, CUDA 13.0) with the sm100a extension loaded.

**T — Test:** `pytest fastvideo/tests/attention/test_vsa_h3_sm100a_route.py::test_real_sm100a_odd_route_matches_triton_oracle -vv -s` — passed.

**E — Evidence:** Slurm job `3017` completed `0:0` and emitted exactly once: `MiniMax-H3 VSA tile-64 forward: using the sm100a CUDA block-sparse kernel`.

PR #1428, opened 2026-06-03
[bugfix] tests: include ltx2_3_base in expected LTX2 preset set (#1427)

Mister-Raggs:
## Problem

`fastvideo/tests/api/test_presets.py::TestLtx2Presets::test_ltx2_presets_registered` fails on a clean `main`:

```
>       assert names == {"ltx2_base", "ltx2_distilled", "ltx2_two_stage"}
E       AssertionError
E         Extra items in the left set:
E         'ltx2_3_base'
```

#1397 (LTX-2.3 transformer support) registered a fourth preset `ltx2_3_base` (`fastvideo/pipelines/basic/ltx2/presets.py`) and made it the LTX2 **default** (`fastvideo/registry.py`), but the test still hard-codes the previous three-preset set. Because this lives in the shared unit-test job (`run_unit_test` → `pytest ./fastvideo/tests/api/ ...`), it red-lights CI for every open PR, not just the one that surfaced it.

## Fix

Add `ltx2_3_base` to the expected set so it matches the presets actually registered for the `ltx2` family:

```python
assert names == {"ltx2_base", "ltx2_3_base", "ltx2_distilled", "ltx2_two_stage"}
```

Verified these are exactly the four `PresetSpec`s with `model_family="ltx2"` registered in `fastvideo/pipelines/basic/ltx2/presets.py`.

## Test evidence

Test-only change to a single assertion set; the `run_unit_test` CI job (which covers `fastvideo/tests/api/`) is the validation surface and will exercise it on this PR.

Fixes #1427

PR #1750, opened 2026-08-24
[bugfix] tests: include ltx2_3_base in local LTX-2 preset set

alanhuangyoo:
`test_ltx2_typed_surface_preflight` asserts an exact set of registered ltx2 presets. #1397 added a fourth (`ltx2_3_base`), so the assertion has been failing since then:

```
assert names == {"ltx2_base", "ltx2_distilled", "ltx2_two_stage"}
AssertionError: Extra items in the left set: 'ltx2_3_base'
```

#1428 already fixed the identical assertion in `fastvideo/tests/api/test_presets.py`. This applies the same one-line change to the copy in `tests/local_tests/`, which was missed because local_tests don't run in CI.

Before, on main:

```
tests/local_tests/ltx2/  ->  8 passed, 1 failed, 8 skipped
```

After:

```
tests/local_tests/ltx2/                ->  9 passed, 8 skipped
fastvideo/tests/api/test_presets.py    ->  64 passed
```

Env: torch 2.12.0+cu130, H20, single node. The 8 skips are the weight-dependent parity tests; unrelated to this change.

### DIFF
--- a/tests/local_tests/ltx2/test_ltx2_pipeline_smoke.py
+++ b/tests/local_tests/ltx2/test_ltx2_pipeline_smoke.py
@@ -316,9 +316,9 @@ def test_ltx2_typed_surface_preflight() -> None:
         LTX2AudioDecodingStage, LTX2DenoisingStage, LTX2LatentPreparationStage, LTX2TextEncodingStage,
     )
 
-    # All three LTX-2 presets registered.
+    # All four LTX-2 presets registered.
     names = {p.name for p in get_presets_for_family("ltx2")}
-    assert names == {"ltx2_base", "ltx2_distilled", "ltx2_two_stage"}
+    assert names == {"ltx2_base", "ltx2_3_base", "ltx2_distilled", "ltx2_two_stage"}
 
     # Two-stage preset has the denoise + refine topology and pulls its
     # refine allowed_overrides from the typed dataclass.


alanhuangyoo:
Still valid on today's `main` (`a943220c`), and the case is stronger than the PR body says: **this exact fix was already accepted for the other copy of the same assertion three months ago.**

`fastvideo/tests/api/test_presets.py:341` on `main`:

```python
assert names == {"ltx2_base", "ltx2_3_base", "ltx2_distilled", "ltx2_two_stage"}
```

`tests/local_tests/ltx2/test_ltx2_pipeline_smoke.py:321` on `main`:

```python
# All three LTX-2 presets registered.
assert names == {"ltx2_base", "ltx2_distilled", "ltx2_two_stage"}
```

Same call (`get_presets_for_family("ltx2")`), same expected set, one of them a preset short. The comment still says "All three" while `presets.py:71` registers four — and `ltx2_3_base` is not some edge case, `registry.py:304` makes it the family's `default_preset`.

The API copy was fixed by **#1428** on 2026-06-04, titled *"[bugfix] tests: include ltx2_3_base in expected LTX2 preset set"* — nearly the same title as this PR. The local copy was missed, and because `tests/local_tests/` only runs on a GPU host it does not show up in PR CI, so nothing has caught it since.

So this is the second half of an already-merged fix rather than a new judgement call.

Status, unchanged: `pre-commit`, `fastcheck-passed` and all eight `buildkite/pr-fastcheck/*` lanes are green. `.github/mergify.yml` also wants `check-success=full-suite-passed` and one approval; `full-suite` runs off the `ready` label, and both `/merge` and the label need write permission, so this needs a maintainer either way.

Worth noting those green ticks are from a base 40 commits behind `main`. Happy to rebase to refresh them — I have left it alone so far because a force-push resets review state and the diff is one line.

PR #1760, opened 2026-08-26
[ci]: fix merge-gate Buildkite state filters

SolitaryThinker:
## Problem

The merge-gate launcher fails before it can schedule Buildkite. Recent runs such as `32933734982` and `32934071377` stop in **Cancel previous Buildkite builds** with:

```text
jq: Cannot index string with string "env"
```

The list-builds request sends `state=running,scheduled`. Buildkite's REST API requires repeated Rails-array parameters for multiple states: `state[]=running&state[]=scheduled`. The rejected response is an error object, but the current `curl -sS` still pipes it into a `jq` filter that assumes an array of builds.

Official contract: https://buildkite.com/docs/apis/rest-api/builds

## Solution

- Send one `state[]` query parameter per requested state.
- Exclude unneeded jobs and expanded pipeline data from this polling request.
- Enable strict shell error handling and make HTTP errors fail the step.
- Validate that the API returned an array of build objects before filtering it.
- Make cancellation request failures visible instead of continuing silently.
- Add a CPU-only workflow contract test so the invalid state encoding and fail-open behavior cannot return.

If the corrected request reports an authorization error, the repository's Buildkite token must be updated to include `read_builds`; cancellation and build creation require `write_builds`.

## Validation

```text
pytest fastvideo/tests/contract/ -q
  88 passed

pre-commit run --files \
  .github/workflows/ci-trigger-full-suite.yml \
  fastvideo/tests/contract/test_ci_test_collection.py
  passed, including GitHub Actions workflow lint

git diff --check
  passed
```

### DIFF
--- a/.github/workflows/ci-trigger-full-suite.yml
+++ b/.github/workflows/ci-trigger-full-suite.yml
@@ -44,16 +44,31 @@ jobs:
           PR_BRANCH: ${{ github.event.pull_request.head.ref }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
+          set -euo pipefail
           # Match both branch and PR number: forks can reuse the same branch name.
-          builds=$(curl -sS --get -H "Authorization: Bearer $BUILDKITE_API_TOKEN" \
+          response=$(curl -sS --fail-with-body --get -H "Authorization: Bearer $BUILDKITE_API_TOKEN" \
             --data-urlencode "branch=$PR_BRANCH" \
-            --data-urlencode "state=running,scheduled" \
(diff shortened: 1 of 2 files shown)


SolitaryThinker:
Superseded by #1762 at reviewed head `2dc322b60b24507834b447019d71ff75a58adb10`. That PR combines the correct repeated Buildkite state filters and strict response validation from this branch with the necessary best-effort cancellation policy, checked cancellation attempts, non-leaking error handling, and CPU contract coverage. Closing this duplicate rather than merging two competing workflow fixes. The unrelated GameCraft T2V SSIM observation from full build #5043 remains tracked by draft diagnostic #1763.

PR #1761, opened 2026-08-26
[perf]: MiniMax H3 on GB10 - load pipeline components on demand instead of all at once

KyleNeverGivesUp:
## Problem

A pipeline materializes every component before the first stage runs, so peak memory is the sum of all components even though no two of them are needed at the same moment. For MiniMax-H3 r16 that sum is 96.3 GiB: text encoder 48.0, DiT 37.8, video VAE 9.7, audio VAE 0.6. For FastH3 it is 124.0 GiB.

The existing CPU offload flags cannot help, for two reasons. They act after loading has already finished, so they do nothing about the load-time peak. And on a unified-memory device such as GB10 there is no second pool to offload into, so moving weights to the host frees nothing.

A GB10 has 121 GiB. FastH3's components do not fit on it at all.

## What this changes

Adds `lazy_module_load`, off by default. Heavy components become a `LazyModule` proxy that loads on first use. The pipeline derives, from what its stages actually hold, which stage is the last user of each component, and installs a release hook there. Peak becomes the largest overlapping set rather than the sum.

## Measured

One GB10, 1 GPU, single prompt, 124 frames. Every number below is
`torch.cuda.memory_allocated()` sampled where a component is loaded or
released, so it measures component residency rather than the run's true peak,
which also includes activations.

### MiniMax-H3 r16, 192x320, 4 steps

| moment | cuda allocated |
|---|---|
| after load, before stage 0 | ~0 |
| video VAE materialized | 9.7 GiB |
| text encoder materialized | 57.7 GiB, peak |
| text encoder released | 9.7 GiB, freed 48.0 |
| DiT materialized | 47.5 GiB |
| audio VAE materialized | 48.1 GiB |
| DiT and video VAE released | 0.6 GiB |
| audio VAE released | 0.03 GiB |

Peak 57.7 GiB against 96.3 GiB resident without the flag. The same command without the flag was killed by the machine's out-of-memory daemon during the DiT load an hour earlier, at `fsdp_load.py:584`.

The decisive moment is two adjacent log lines, 4 ms apart:

```
Released deferred module text_encoder, cuda allocated 57.72 -> 9.74 GiB, freed 47.98 GiB
Loading deferred module transformer
```

The DiT starts loading from 9.7 GiB rather than from 58.3 GiB.

### FastH3 Preview v0.2, VSA-H3 attention on the Triton kernel

| resolution | packed tokens | resident components | result |
|---|---|---|---|
| 192x320 | 29,760 | 75.81 GiB | completed, 525 s, video and audio written |
| 384x672 | 124,992 | 75.81 GiB | completed, 745 s, video and audio written |
| 512x896 | 222,208 | 75.82 GiB | completed, 930 s, video and audio written |
| 640x1120 | 347,200 | 75.83 GiB | killed during the first DiT forward |
| 768x1344 | 499,968 | 75.83 GiB | killed during the first DiT forward |

Component residency does not move with resolution, because it is set by the
weights. What moves is activation memory, which this change does not touch.
The last two rows load every component and then die in
`minimax_h3_denoising.py:190`, so the ceiling on one GB10 sits between 222,208
and 347,200 packed tokens. That ceiling also reflects whatever else the shared
machine was running, since the out-of-memory daemon watches the host rather
than this process, so treat it as a lower bound.

DiT 35.05B parameters, 65.5 GiB. The four components sum to 124.0 GiB against the device's 121 GiB, so without the flag this model cannot load at all and there is no eager baseline to compare against.

### Still applies to FastH3 Preview v1

FastVideo has since released [FastH3 Preview v1](https://huggingface.co/FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-DataFree). The tables above were measured on v0.2 and are left as recorded, but nothing this change depends on moved: the two checkpoints ship a byte-identical `transformer/config.json`, the same 35.05B parameters, and the same 688 student tensors, so the component sum that overflows a 121 GiB device is unchanged.

Measured with this flag on v1, using [KyleNeverGivesUp/FastH3-4-step-Preview-v1-r16](https://huggingface.co/KyleNeverGivesUp/FastH3-4-step-Preview-v1-r16), which is v1 with the rank-reduced AdaLN converter from #1699 applied:

| frames | resolution | packed tokens | FP8 | resident weights | peak allocated | time |
|---|---|---|---|---|---|---|
| 124 | 768x1344 | 499,968 | no | 51.70 GiB | 69.2 GiB | 902 s |
| 345 | 768x1344 | 1,403,136 | yes | 33.06 GiB | 83.7 GiB | 3096 s |

Both peaks match the same conversion applied to v0.2 to the byte, which is what you would expect if peak is set by the architecture and the packed token count rather than by which training run produced the weights.

The 345-frame row is the longest MiniMax-H3 generates, 14.38 s of synchronized video and audio in one pass on a single GB10. It needs FP8 to fit, and FP8 only reaches the feed-forward stack after #1780.

## Cost

A released component is read from disk again on the next generation. Within one generation nothing reloads, since the release point is the last stage that holds the component, not the last stage that uses it. Across generations the whole set reloads, which for this model is several minutes. That is why the flag is off by default and documented as something to enable only when the model does not otherwise fit.

## Design notes

Four things here would fail silently if done the obvious way.

**The release hook lives on `PipelineStage.__call__`, not in the pipeline's stage loop.** Several pipelines override `forward`, and a loop-based hook would leave those pipelines deferring but never freeing, with no error.

**The proxy forwards `__class__`.** Callers branch on `isinstance(module, FSDPModule)` and similar. A proxy answering `False` would take the wrong branch and produce a wrong result rather than a crash. The cost is that an isinstance check materializes, which is the correct trade.

**Self-returning methods hand back the proxy, not the component.** Stages write `self.vae = self.vae.to(device)` in a dozen places. Returning the component there would replace the proxy with a strong reference the pipeline cannot release, and the run would look completely normal while freeing nothing.

**`add_stage` rebuilds the schedule if it runs after the schedule was built.** Every pipeline in the tree builds its stages inside `create_pipeline_stages`, but nothing enforced it. A stage appended later could hold a component an earlier stage had already been told to free.

Building the schedule also has to avoid `isinstance` on proxy values, since that would materialize every deferred component before the run starts. There is a test for exactly that.

Training keeps every component resident and warns if the flag is set. If no stage holds a deferred component the pipeline warns rather than silently doing nothing.

## Scope

Deferral is opt-in per pipeline. `_lazy_module_names` is empty in the base class and MiniMax-H3 lists the four components measured above. Every other pipeline is unchanged, and setting the flag there logs a warning and does nothing.

The reason is that releasing a component and loading it again is only correct when nothing outside the loader has changed it, and two habits in this tree break that without raising:

- Mutating a component after load. `LongCatPipeline.initialize_pipeline` turns on block-sparse attention and writes parameters into every transformer block. It runs once, so a re-materialized component silently comes back with the feature off.
- Reading a component's attributes while stages are built. The shared `DenoisingStage.__init__` derives the attention backend from `transformer.hidden_size`, which materializes the DiT during `post_init` and defeats the deferral it was meant to gain.

A pipeline opts in once someone has checked it against both. `LTX2Pipeline`, `MagiHumanPipeline`, and `LingbotVideoPipeline` additionally override `load_modules` without delegating to the base implementation, so they would need that too.

The flag lowers weight residency only. It does not change activation memory, so a resolution whose activations do not fit still does not fit. On one GB10, FastH3 runs at 512x896 and is killed at 640x1120, where q, k, v, and the VSA gate alone account for 18.5 GiB against roughly 30 GiB left after the weights. The repository's reference configuration is 768x1344 on four GPUs with `sp_size=4`, which splits those activations across ranks.

## Relationship to the existing MPS special cases

`denoising.py:653`, `decoding.py:343`, and `sr_denoising.py:279` already do a hard-coded version of this behind `if torch.backends.mps.is_available()`, with matching reload paths keyed on `fastvideo_args.model_loaded`. Apple Silicon is unified memory too, so that is the same problem found on a different device.

Those three differ from this change in two ways. They still load everything eagerly first, so they lower the steady-state footprint but not the load-time peak, which is where H3 on GB10 dies. And they never release the text encoder, which for H3 is the largest component at 48 GiB.

This change could subsume all three later. It does not touch them here.

## Tests

31 cases in `fastvideo/tests/stages/test_lazy_module_load.py`, CPU only, covering the proxy contract, the release schedule, the identity rule for self-returning methods, the late-stage rebuild, and the enablement rules. Placed under `stages/` because `fastvideo/tests/loader/` is not collected by any CI lane on this branch's base.

`lazy_module_load` is registered in `docs/design/inference_schema_parity_inventory.yaml` and mapped in `api/compat.py`, so `test_fastvideo_args_fields_are_classified` stays green.

### DIFF
--- a/docs/inference/offloading.md
+++ b/docs/inference/offloading.md
@@ -152,6 +152,15 @@ the model already fits.
 This option applies to inference only. Training keeps every component resident
 and logs a warning if the flag is set.
 
+Deferral is opt-in per pipeline. Releasing a component and loading it again is
+only safe when nothing outside the loader has changed it, and two common habits
+break that without raising: mutating a component after load, as LongCat does
+when it enables block-sparse attention, and reading a component's attributes
+while stages are built, as the shared denoising stage does to pick an attention
+backend. A pipeline therefore lists the components it has checked in
+`_lazy_module_names`, which is empty in the base class. MiniMax-H3 opts in. On
+a pipeline that has not, the flag logs a warning and changes nothing.
+
 ## General Recommendations
 
 ### Single GPU Inference
--- a/docs/inference/offloading.md
+++ b/docs/inference/offloading.md
@@ -139,7 +139,9 @@ pipeline that is roughly `max(text encoder, DiT + VAE)` rather than
 
 A freed component is read from disk again on the next generation, so a
 multi-prompt run pays one reload per component per request. For a large text
-encoder that is tens of seconds.
+encoder that is tens of seconds. If pipeline-level `torch.compile` is enabled,
+the compile setup is reapplied after each reload; PyTorch can reuse its graph
+and kernel caches when the component structure and input shapes are unchanged.
 
 #### Usage Recommendation
 
--- a/docs/inference/offloading.md
+++ b/docs/inference/offloading.md
@@ -12,6 +12,7 @@ text_encoder_cpu_offload: bool = True
 image_encoder_cpu_offload: bool = True
 vae_cpu_offload: bool = True
 pin_cpu_memory: bool = True
+lazy_module_load: bool = False
 ```
 
 On unified-memory accelerators such as NVIDIA GB10 and Apple silicon, FastVideo
@@ -121,6 +122,36 @@ These options introduce performance overhead due to PCIe data transfer.
 
 We recommend enabling these options when OOM happens.
 
+### `lazy_module_load`
+
+Every option above moves weights between host and device. This one changes
+whether they are in memory at all.
+
+By default a pipeline loads every component before the first stage runs, so
+peak memory is the sum of all of them even though no two are needed at the same
+moment. With `lazy_module_load` enabled, each heavy component loads on first use
+and is freed once the last stage that needs it has returned, so peak memory
+becomes the largest overlapping set instead of the sum. For a text-to-video
+pipeline that is roughly `max(text encoder, DiT + VAE)` rather than
+`text encoder + DiT + VAE`.
+
+#### Performance Impact
+
+A freed component is read from disk again on the next generation, so a
+multi-prompt run pays one reload per component per request. For a large text
+encoder that is tens of seconds.
+
+#### Usage Recommendation
+
+Enable this when a model does not fit at load time, which the CPU offload
+options above cannot help with because they act after loading. It is
+particularly relevant on unified-memory devices, where host and device draw on
+the same pool and moving weights to the host frees nothing. Leave it off when
+the model already fits.
+
+This option applies to inference only. Training keeps every component resident
+and logs a warning if the flag is set.
+
 ## General Recommendations
 
 ### Single GPU Inference
@@ -130,6 +161,12 @@ We recommend enabling `dit_layerwise_offload`. If OOM happens, also enable `imag
 ### Multi-GPU Inference
 
 We recommend enabling `use_fsdp_inference` and disabling both `dit_layerwise_offload` and `dit_cpu_offload`. If OOM happens, consider enabling `text_encoder_cpu_offload`, `image_encoder_cpu_offload`, and `vae_cpu_offload`. If OOM still happens, consider enabling `dit_cpu_offload`.
+
+### When the Model Does Not Fit at Load Time
+
+The offload options only help once loading has finished. If the run dies while
+components are still being placed, or if the machine has unified memory so
+there is no separate host pool to offload into, enable `lazy_module_load`.
 
 ## Examples
 
--- a/examples/inference/basic/basic_fasth3.py
+++ b/examples/inference/basic/basic_fasth3.py
@@ -48,6 +48,12 @@ def build_parser(description: str | None = None) -> argparse.ArgumentParser:
     # License review completes. A local snapshot can be passed here instead.
     parser.add_argument("--prompt", required=True)
     parser.add_argument("--output", default="outputs/fasth3")
+    parser.add_argument("--lazy-module-load",
+                        action="store_true",
+                        help="load each heavy component on first use and free it after the last stage that "
+                        "needs it, so peak memory is the largest overlapping set instead of the sum of every "
+                        "component. Enable when the model does not fit at load time; costs a reload per "
+                        "generation, so leave it off when it does fit")
     parser.add_argument("--profile",
                         choices=("all", "strict"),
                         default="all",
@@ -260,6 +266,7 @@ def build_generator_config(args: argparse.Namespace) -> GeneratorConfig:
                 text_encoder=True,
                 vae=True,
                 pin_cpu_memory=args.pin_cpu_memory,
+                lazy_module_load=args.lazy_module_load,
             ),
             compile=CompileConfig(
                 enabled=args.torch_compile,
--- a/fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py
+++ b/fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py
@@ -92,6 +92,10 @@ class MiniMaxH3BasePipeline(LoRAPipeline, ComposedPipelineBase):
         "scheduler",
         "audio_scheduler",
     ]
+    # Deferral is safe here: no stage reads a component's attributes while it
+    # is being constructed, and `initialize_pipeline` only inspects the
+    # schedulers, which are never deferred.
+    _lazy_module_names = ("text_encoder", "transformer", "vae", "audio_vae")
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._ref2va = False
--- a/fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py
+++ b/fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py
@@ -119,6 +119,15 @@ class MiniMaxH3BasePipeline(LoRAPipeline, ComposedPipelineBase):
     def _defer_denoise_modules(self, fastvideo_args: FastVideoArgs) -> bool:
         if not fastvideo_args.inference_mode or bool(getattr(fastvideo_args, "training_mode", False)):
             return False
+        # Both mechanisms defer the same four modules and both decide when to
+        # free them. Running them together interleaves two owners over one set
+        # of components: this override would strip the denoise modules from the
+        # load list while the base wrapped the encoder in a proxy, and both
+        # would then release it. `lazy_module_load` is asked for explicitly and
+        # is the more general path, so it wins when it is on.
+        if bool(getattr(fastvideo_args, "lazy_module_load", False)):
+            logger.info("MiniMax-H3 sequential module load off: lazy_module_load owns deferral")
+            return False
         requested = fastvideo_args.h3_sequential_load
         if requested is True:
             return True
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -30,6 +30,7 @@ surfaces:
       image_encoder_cpu_offload: generator.engine.offload.image_encoder
       vae_cpu_offload: generator.engine.offload.vae
       pin_cpu_memory: generator.engine.offload.pin_cpu_memory
+      lazy_module_load: generator.engine.offload.lazy_module_load
       enable_torch_compile: generator.engine.compile.enabled
       enable_torch_compile_text_encoder: generator.engine.compile.text_encoder_enabled
       enable_torch_compile_vae: generator.engine.compile.vae_enabled
--- a/examples/inference/basic/basic_minimax_h3_t2v.py
+++ b/examples/inference/basic/basic_minimax_h3_t2v.py
@@ -48,6 +48,12 @@ def parse_args() -> argparse.Namespace:
                         "training-port semantics: no kwargs; fullgraph + emulate_precision_casts injected). "
                         "First generation pays the inductor JIT (~1-2 min); use --repeats >= 2 and time "
                         "the last repeat. FASTVIDEO_INFERENCE_TORCH_COMPILE=1 is equivalent")
+    parser.add_argument("--lazy-module-load",
+                        action="store_true",
+                        help="load each heavy component on first use and free it after the last stage that "
+                        "needs it, so peak memory is the largest overlapping set instead of the sum of every "
+                        "component. Enable when the model does not fit at load time; costs a reload per "
+                        "generation, so leave it off when it does fit")
     parser.add_argument("--repeats",
                         type=int,
                         default=1,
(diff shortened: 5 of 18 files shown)


docs/getting_started/installation/spark_performance.md
```diff
@@ -161,6 +161,12 @@ is power-cycled. To avoid it:
   on: "CPU" offload uses the same unified RAM. Multi-GPU FSDP sharding remains
   available because it partitions weights without parking them in a separate
   host pool.
+- **MiniMax H3 / FastH3** still needs sequential loading on one GB10. The Qwen3-VL
+  conditioner is tens of gigabytes of BF16. If the DiT and VAEs load while that
+  encoder is still resident, the process is a typical `earlyoom` kill (Python is
+  preferred). The CUDA pipeline now encodes first, releases the encoder, then
+  loads DiT and VAEs onto the accelerator (`to_cpu` follows `cpu_offload`, which
+  is off here). See [Offloading](../../inference/offloading.md).
```
aryan5v: This reads as if CUDA H3 now sequential-loads by default. In this PR it does not: `lazy_module_load` defaults to `False`, and `basic_fasth3.py` only turns it on with `--lazy-module-load`.

On one GB10, FastH3 still loads encoder + DiT + VAEs together unless that flag is set, which is the `earlyoom` case this bullet describes.

Please state the flag, and either auto-enable it on unified memory or show `--lazy-module-load` in the Spark command. The GPU-direct `to_cpu=cpu_offload` sentence can stay.

docs/inference/offloading.md
```diff
@@ -21,6 +22,16 @@ pool there, so offload adds transfers and duplicate residency instead of freeing
...
 
+MiniMax H3 CUDA inference uses a second lever that does not copy weights to a
+host pool. The pipeline loads the Qwen3-VL text encoder, runs conditioning, then
+releases that encoder before it loads the DiT and video/audio VAEs. The MLX FastH3
+runtime uses the same phase order. When host offload is off, DiT safetensors are
+read onto the accelerator instead of CPU-then-copy. Input-preparation geometry
+(spatial ratio, latent channels, audio sample rate) comes from the VAE arch
+configs until those weights load. A later `generate()` on the same worker
+currently re-enters conditioning after the encoder has been released; start a
+new generator for a new prompt until prompt-cache reload exists.
```
aryan5v: This paragraph does not match the code in this PR.

- Sequential encode-then-release is not default CUDA H3 behavior; it is `--lazy-module-load`.
- Input-prep geometry is not taken from VAE arch configs here. The first input-prep access of `spatial_compression_ratio` / `latent_channels` materializes the full video VAE.
- A later `generate()` on a lazy worker re-runs the loader after `release()`; it does not require a new process. The broken second-prompt behavior and the MLX phase-order sentence belong to a different design.

Please describe `lazy_module_load` only (the section below already does that) and drop this always-on sequential-start writeup.

examples/inference/basic/basic_fasth3.py
```diff
@@ -48,6 +48,12 @@ def build_parser(description: str | None = None) -> argparse.ArgumentParser:
     # License review completes. A local snapshot can be passed here instead.
     parser.add_argument("--prompt", required=True)
     parser.add_argument("--output", default="outputs/fasth3")
+    parser.add_argument("--lazy-module-load",
+                        action="store_true",
+                        help="load each heavy component on first use and free it after the last stage that "
+                        "needs it, so peak memory is the largest overlapping set instead of the sum of every "
+                        "component. Enable when the model does not fit at load time; costs a reload per "
+                        "generation, so leave it off when it does fit")
```
aryan5v, line 56: Default `all` on one GB10 cannot keep FastH3's four components resident (PR body: 124 GiB vs 121 GiB). Leaving this off means the Spark example still dies at load unless the user already knows the flag.

Either auto-enable on unified memory (same place `disable_offload_on_unified_memory` already runs) or default this on when `num_gpus == 1` and document the reload cost. Off-by-default is reasonable on 4×GB200.

fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py
```diff
@@ -52,6 +52,10 @@ class MiniMaxH3BasePipeline(LoRAPipeline, ComposedPipelineBase):
         "scheduler",
         "audio_scheduler",
     ]
+    # Deferral is safe here: no stage reads a component's attributes while it
+    # is being constructed, and `initialize_pipeline` only inspects the
+    # schedulers, which are never deferred.
+    _lazy_module_names = ("text_encoder", "transformer", "vae", "audio_vae")
```
aryan5v, line 98: Opt-in list is right, but last-holder scheduling still overlaps more than H3 needs:

- Input prep holds `vae` and reads two ints in forward, so the video VAE loads before the text encoder (your r16 table: 9.7 then 57.7 GiB peak).
- Video decode holds `transformer` only for `patch_size`, so the DiT stays through VAE decode (FastH3 75.8 GiB resident).

If those scalars come from arch config / the batch instead of the live modules, encode peak drops the VAE and decode can free the DiT. Construction-time access is already tested; first-forward attribute reads are not.

aryan5v (changes requested):
Reviewed the lazy-load path, H3 opt-in, GPU-direct DiT iterator, and the Spark/offload docs against GB10 FastH3 behavior.

The `LazyModule` contract is in good shape: `is_lazy_module` avoids `__class__` materialization, self-returning `.to()` keeps the proxy, compile is a materialize transform, LoRA config reads are deferred, nested-stage walk and failure-path release are covered, and the CPU tests hit the production `load_modules` / H3 stage constructors. The `to_cpu=cpu_offload` change in `fsdp_load.py` is the right GB10 load path.

Please fix the docs before merge. They currently describe always-on sequential H3 start (encode, drop the encoder, then load DiT/VAE, arch-config geometry, one prompt per worker). This PR is an opt-in `lazy_module_load` flag that defaults to off, reloads from disk on the next `generate()`, and materializes the video VAE on the first input-prep attribute read. Spark operators following those docs will still OOM.

Two residency issues on the motivating model, neither blocked by tests:

1. `MiniMaxH3InputPreparationStage` reads `vae.spatial_compression_ratio` and `vae.latent_channels` in forward, so the 9.7 GiB video VAE loads before Qwen3-VL. That is the 57.7 GiB encode peak in the PR table. Those values are already on `MiniMaxH3VideoVAEArchConfig` / `MiniMaxH3AudioVAEArchConfig` and do not need the weights.
2. `MiniMaxH3VideoDecodingStage` holds `transformer` only for `patch_size`, so the 65 GiB DiT stays resident through VAE decode (the 75.8 GiB FastH3 row). Stash `patch_size` on the batch during latent prep / denoise and drop the DiT before decode.

Nit: VSA `prepare_for_compile` graph sharing is a separate change. Fine if it is required for compile-across-reload, otherwise it is easier to review on its own.

aryan5v:
Addressed the CHANGES_REQUESTED review on a branch you can merge into this PR: https://github.com/KyleNeverGivesUp/FastVideo/pull/2 (`aryan5v:aryan/1761-lazy-geometry` → `lazy-module-load`). I cannot push your fork (`permissions.push: false`).

What changed:
- Input prep holds the checkpoint-updated VAE **arch config**, not the live VAE proxy, so the first `spatial_compression_ratio` / `latent_channels` read no longer materializes 9.7 GiB before Qwen.
- Decode and latent packing read `patch_size` from `pipeline_config.dit_config` (overlayed from `transformer/config.json` in `initialize_pipeline`, no weights). Decode no longer holds the DiT, so last-holder scheduling drops it before VAE decode.
- Docs now describe `--lazy-module-load` (opt-in / auto), not always-on sequential start. A later `generate()` reloads from disk.
- Auto-enable: `lazy_module_load=None` turns on after the worker binds a unified-memory device. `basic_fasth3.py` / `basic_minimax_h3_t2v.py` also default it on when `--num-gpus 1`. `--no-lazy-module-load` keeps everything resident (reasonable on 4×GB200).

Spark2 GB10 check, alpine 768×1344×124, VSA-DataFree, seed 2026, `--vsa-kernel triton --no-fa4 --no-warmup --repeats 1 --lazy-module-load`, **full VAE** (not TAEH3):

1. `geometry from config: patch_size=(1, 2, 2) spatial_compression_ratio=16 latent_channels=24` — input prep 0.05 ms, no VAE weights
2. Load Qwen 47.98 GiB → release
3. Load DiT 66.12 GiB → denoise ~138 s → **release DiT (freed 65.51 GiB)**
4. **Then** load VAE 10.34 GiB and decode
5. Release VAE 9.71 GiB

E2E 399.3 s. Denoise matches the sequential-start full-VAE run (~138 vs 139 s). The extra wall time vs 336.1 s sequential is cold Qwen load inside `generate()` and first `torch.compile` of the VAE at decode, not a slower DiT.

CUDA TAEH3 is **not** in this follow-up; that is a separate PR against `main`.

aryan5v:
#1803 supersedes this PR for the remaining FastH3-on-Spark work.

## Already on main

The GB10 “do not load encoder + DiT + VAE all at once” path landed in [#1793](https://github.com/hao-ai-lab/FastVideo/pull/1793) (`h3_sequential_load`, auto on unified memory, GPU-direct DiT load). That is the overlapping slice of this PR. GitHub marks this branch `CONFLICTING` against main because of that merge, not because the Spark-pair work failed.

## Preserved from this PR

What survived is the **behavior**, not this branch’s `LazyModule` machinery:

- Encode with Qwen3-VL, release the encoder, then load DiT / VAEs so a single 128 GB GB10 can start FastH3.
- GPU-direct DiT safetensors when host offload is off (also in #1793).
- Dual-Spark still needs that split: sequence parallel **replicates** the DiT (~66 GiB/node), so each box still cannot hold encoder + DiT + VAE together. On main that is `--h3-sequential-load` / auto, not `--lazy-module-load`.

Still **only on this branch** (not on main, not in the Spark-pair commit): generic `LazyModule` + release-after-last-stage, in-process reload for a second `generate()`, geometry from checkpoint JSON so decode does not hold a live DiT, compile/LoRA hooks around lazy reload, and `test_lazy_module_load.py`. Those need a fresh rebase onto main if they are still wanted; they should not merge through this conflicted stack.

## What #1803 adds on top

[#1803](https://github.com/hao-ai-lab/FastVideo/pull/1803) is the follow-on, stacked on **main + #1793**, not on this lazy-load implementation:

- Ray executor: driver-local `set_log_queue` / `clear_log_queue` (Queue is not picklable across nodes) and copy `NCCL_*` onto workers.
- FastH3 `--execution-backend {mp,ray}` (auto-ray when `RAY_ADDRESS` is set).
- Dual-Spark QSFP / Ray bring-up, cookbook recipe, YAML, env script.
- Measured pair runs: SP=2 124f **292 s**, 345f **587 s** vs ~374–393 s on one Spark.

Please review / land Spark-pair on #1803. This PR should not merge as-is on top of #1793.

PR #1764, opened 2026-08-26
[ci][diagnostic]: capture GameCraft FA4 candidate on GB200

SolitaryThinker:
Diagnostic only — do not merge.

This disposable branch generates one current-FA4 HunyuanGameCraft T2V candidate on GB200 and makes that small MP4 recoverable from the public Buildkite job log. It is separate from PR #1763 so build #5058 and its exact head remain untouched.

Safety gates:

- based on merged `upstream/main` at `c9c558575804037ee95fd41014385a69bdbf7a85`;
- runs only `test_gamecraft_t2v_similarity`, never I2V;
- pins `FASTVIDEO_FA4=1`, default-quality parameters, and bootstrap mode off;
- requires checkout SHA to equal `BUILDKITE_COMMIT`, exactly one visible GB200, and resolved FlashAttention 4;
- leaves the existing `0.93` SSIM threshold unchanged and requires a fresh matching SSIM JSON;
- performs no Hugging Face upload and needs no media-upload credential;
- forces lane exit 2, so it cannot be mistaken for a passing quality gate and does not trigger the exit-1 SSIM retry policy.

Pytest's noisy output is redirected to `/tmp`; only a bounded tail is printed. The generated MP4 is capped at 1 MiB and emitted last as numbered 3-KiB binary/base64 chunks between strict markers with byte count and SHA-256. The companion decoder accepts raw marker text or Buildkite's public `{ "output": ... }` log JSON, refuses output overwrite, and verifies the complete chunk sequence, decoded size, and SHA-256.

Local validation:

- `bash -n .buildkite/scripts/lanes/ssim.sh`
- Python bytecode compilation of `.buildkite/scripts/gamecraft_candidate_log.py`
- `pytest fastvideo/tests/contract/test_ci_test_collection.py -q` (13 passed)
- targeted `pre-commit` (filename check passed; configured hooks skip these Buildkite paths)
- `git diff --check`
- exact round trip of the backed-up 447,050-byte canonical GB200 MP4 from raw and simulated timestamp-prefixed public Buildkite JSON with `/` converted to `&#47;` and timestamp tags entity-escaped; both recovered SHA-256 `29e8025d5a1495093ec7ccb150fc3d2eed1e13f1a67368a2aeacd6d52ae6e2e9`; a deleted chunk was rejected
- entity-escaped simulated public JSON size: 654,626 bytes, 148 envelope lines, maximum rendered line length 4,594 bytes

After the candidate is decoded and visually reviewed, close this PR. Any canonical reference upload is a separate, explicitly user-authorized operation.

### DIFF
--- a/.buildkite/scripts/gamecraft_candidate_log.py
+++ b/.buildkite/scripts/gamecraft_candidate_log.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Emit and recover a small GameCraft MP4 through a Buildkite job log."""
+
+from __future__ import annotations
+
+import argparse
+import base64
+import binascii
+import hashlib
+import html
+import json
+import re
+import sys
+from pathlib import Path
+
+MARKER = "FV_GAMECRAFT_FA4_MP4"
+CHUNK_BYTES = 3 * 1024
+MAX_MEDIA_BYTES = 1024 * 1024
+
+
+def _sha256(data: bytes) -> str:
+    return hashlib.sha256(data).hexdigest()
+
+
+def emit(media_path: Path) -> None:
+    if media_path.suffix.lower() != ".mp4":
+        raise ValueError(f"Candidate must be an MP4: {media_path}")
+
+    data = media_path.read_bytes()
+    size = len(data)
+    if not 0 < size <= MAX_MEDIA_BYTES:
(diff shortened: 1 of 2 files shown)


SolitaryThinker:
Rebased onto current main so the reference candidate uses the current runtime path; no content changes (range-diff clean).

PR #1768, opened 2026-08-27
[feat] Add Cosmos Predict2.5 DFD continuation to DreamVerse

Mister-Raggs:
## Purpose

Add continuous Cosmos Predict2.5 generation to DreamVerse using two complementary four-step students from #1767:

- Distilled Text2World creates an unconditioned bootstrap segment.
- Data-Forcing Distillation (DFD) Video2World conditions each later segment on the prior terminal frame.

This replaces the earlier independent one-shot Cosmos profile. The PR is intentionally stacked on #1767; until that PR merges, GitHub will include its core model commits in this diff.

## Architecture

- Add a dedicated `Cosmos25DFDGenerationBackend` behind the model-owned `GenerationBackend` interface introduced by merged PR #1800.
- Keep Cosmos lifecycle, request construction, conditioning, silent audio, and overlap metadata isolated from the existing LTX2 and FastH3 backends.
- Load two explicit FastVideo package roles during initialization and warm both request shapes before readiness.
- Retain a copied RGB terminal frame after each segment and pass it to the DFD package as exactly one conditioning frame.
- Route an initial user image directly to DFD. It is not treated as streamed overlap, so its first frame is retained.
- Route reset/unconditioned sessions back through the T2W bootstrap package.
- Give the slower Cosmos profile a 30-minute session lease while preserving the existing five-minute default for other backends. Deployments can override it with `FASTVIDEO_SESSION_TIMEOUT_SECONDS`.

## Media contract

- T2W bootstrap: `704x1280x77`, 24 FPS, four steps.
- DFD continuation: `704x1280x81`, 24 FPS, four steps.
- DFD repeats its conditioning image as decoded frame 0 by design. DreamVerse returns one-frame video and audio trim metadata for retained-frame continuations, so the browser receives 80 new DFD frames with no duplicated boundary.
- Cosmos produces no audio. The backend synthesizes duration-matched silent 24 kHz audio; one continuation-frame trim removes exactly 1,000 samples.
- The existing fMP4 streamer remains model-agnostic and unchanged because all DreamVerse backends use 24 FPS.

## Configuration

```bash
export DREAMVERSE_MODEL_ID=cosmos25-dfd
export DREAMVERSE_MODEL_PATH=/path/to/Cosmos-Predict2.5-2B-Distilled-TrigFlow-FastVideo
export DREAMVERSE_COSMOS25_DFD_MODEL_PATH=/path/to/Cosmos-Predict2.5-2B-DFD-FastVideo
export FASTVIDEO_SESSION_TIMEOUT_SECONDS=1800
export ENABLE_TORCH_COMPILE=0
dreamverse-server --host 0.0.0.0 --port 8009
```

Public converted package IDs are still pending, so both local conversion outputs are currently required.

## Tests

Local, no model download:

- Cosmos backend, registry, generation-worker, and session-lease tests: 25 passed.
- Additional GPU-pool, prompt, stress, and session tests: 54 passed.
- YAPF, Ruff, codespell, PyMarkdown, filename checks, and mypy: passed from a valid Python-package worktree path.
- Import-free compilation and `git diff --check`: passed.

The remaining local suite was not collected in the lightweight Mac venv because it lacks `fastapi`; four existing FastH3 tests also require its missing `torchvision`. The branch-specific focused suite passes.

Model evidence inherited from #1767:

- T2W and DFD real-weight DiT parity pass.
- Complete four-step DFD latent rollout parity passes against upstream.
- Native DFD conversion, strict load, decode, and save pass at `704x1280x81`.
- Native boundary MAE is `2.845237` versus a final-frame control of `52.130680`.
- Reference T2W -> DFD and DFD -> DFD boundaries are visually seamless; the prompt-directed right camera pivot succeeds. Boundary MAE is `2.694` and `2.459`, versus unrelated control `39.142`.

## DreamVerse integration smoke

Validated on an NVIDIA DGX Spark / GB10 with both converted packages resident in one DreamVerse GPU worker:

- Both T2W and DFD generators loaded together and both startup warmup request shapes completed before readiness.
- A browser-driven three-stage curated rollout completed through T2W bootstrap -> DFD continuation -> DFD continuation.
- Segment prompts changed from straight driving, to beginning a right turn, to completing the turn and accelerating.
- fMP4 streaming, per-boundary one-frame/audio trimming, browser playback, IndexedDB archival, multi-segment remux, and MP4 download completed successfully.
- Observed generation latency was approximately 152 seconds for the T2W segment and 172 seconds for a DFD segment, confirming this is continuation-quality integration rather than real-time generation on GB10.
- The first rollout exposed the legacy 300-second session lease expiring during active DFD work. Commit `56c3b1df5` gives the Cosmos profile a 1,800-second default and adds an environment override; the full rollout then completed.

## Media

The native DFD output is documented on #1767. The full three-stage DreamVerse browser-stream capture is [attached in the GB10 validation comment](https://github.com/hao-ai-lab/FastVideo/pull/1768#issuecomment-5631971842) ([direct MP4](https://github.com/user-attachments/assets/afb90826-19b7-409c-b1d8-193dc8d59edd)).

## Scope and remaining work

- This is continuation-quality integration, not real-time generation: full segments took roughly 152-172 seconds on GB10 after model load.
- Runtime LoRA changes are rejected for this backend.
- Both generators are intentionally resident so either bootstrap or DFD can run without a mid-session reload.
- Replace the provisional package IDs after the converted T2W and DFD packages are published.

## Checklist

- [x] Rebase onto current main and merged H3 backend architecture.
- [x] Replace independent clips with T2W bootstrap plus DFD continuation.
- [x] Cover package selection, image input, reset, retained-frame continuation, one-frame trim, silence, warmup, shutdown, and backend dispatch.
- [x] Preserve LTX2/FastH3 backends and the shared streaming contract.
- [x] Complete the two-package, three-stage DreamVerse deployment smoke.
- [x] Attach the browser-stream continuation capture.
- [ ] Replace provisional package IDs when public converted packages are published.


Mister-Raggs:
### DGX Spark / GB10 browser validation

Three-stage DreamVerse rollout:

1. Cosmos distilled T2W bootstrap
2. DFD continuation beginning the right turn
3. DFD continuation completing the turn

https://github.com/user-attachments/assets/afb90826-19b7-409c-b1d8-193dc8d59edd


Validated continuous browser playback, boundary trimming, fMP4 streaming, archival, remux, and MP4 download.

SolitaryThinker:
Pushed `9da031854`: keep the Cosmos2.5 conditioning encode deterministic and wire the new scheduler tests into CI.

**Fixed**
- `fastvideo/models/vaes/cosmos25wanvae.py:44` — `_TensorLatentDist.sample()` returned `mean + stddev * noise` and `Cosmos25WanVAE.encode` always supplied a posterior std. That silently changed the existing Cosmos2.5 I2W/V2W route (`Cosmos25V2WLatentPreparationStage` -> `retrieve_latents` -> `.sample()`) from the deterministic encode mean to a seeded posterior sample, and it made the new DFD route condition on a sampled latent. FastGen builds this condition deterministically (`CosmosPredict2.build_conditioning` calls `vae.encode(cond_frames, mode="argmax")`, and `mode="argmax"` maps to `dist.mode()`), and this PR's own decision log keeps the existing Cosmos25 path unchanged. Reverted the file to `main` so both routes use the encode mean.
- `.buildkite/scripts/unit_test.sh:13` — `fastvideo/tests/schedulers/` was in no CI lane, so `test_cosmos25_dfd_scheduler.py` and `test_cosmos25_distilled_scheduler.py` never executed. Added the directory to the unit lane.
- `fastvideo/pipelines/basic/cosmos/presets.py:58` — the DFD preset advertised `num_inference_steps`/`guidance_scale` overrides that `Cosmos25DFDInputValidationStage` always rejects. The DFD preset now uses a stage spec with no allowed overrides, so the rejection happens at preset validation instead of at runtime.
- `fastvideo/pipelines/stages/latent_preparation.py:803` — dropped the no-op `sample if batch_size == 1 else sample` ternary and the redundant `if batch_size == 1: break` in the DFD conditioning encode.

**Verified**
- `python3 -m py_compile` on all changed Python files, `bash -n` on `unit_test.sh`, and a 120-column check on the changed files.
- CI is the gate: `pre-commit` and `buildkite/pr-fastcheck` are re-running against the new head (`9da031854`). No local test suite was run.

**Left open**
- No SSIM/quality lane for the new Cosmos2.5 distilled/DFD routes; that needs GPU validation plus reference artefacts, and `tests/local_tests/cosmos25/PORT_STATUS.md` documents the visual gates instead.
- `apps/dreamverse/dreamverse/config.py:91` still defaults to the provisional `FastVideo/Cosmos-Predict2.5-2B-*` package IDs, which the PR body and `docs/inference/support_matrix.md` already track as pending.
- Duplicated denoising-stage/latent-prep skeletons and the duplicated conversion-script helpers were left as-is (refactors, not review fixes).
- The VAE change originates in the stacked #1767 content, so mirroring this revert there keeps both branches consistent.

PR #1776, opened 2026-08-28
[perf] Add MiniMax H3 MLX VSA and SIMD attention

aryan5v:
Depends on #1770. This PR is stacked on its MLX H3 runtime branch and narrows to the changes below after #1770 merges.

## Summary

- Add VSA-capable H3 MLX conversion, checkpoint metadata, runtime routing, CLI controls, and dense-checkpoint guards.
- Keep the chunked gather plus SDPA backend as `auto`; add an explicit `--vsa-impl simd` Metal backend for tile size 64 and head dimension 128, with automatic reference fallback.
- Use fused MLX RMSNorm in H3 and remove redundant contiguous copies from dense SDPA.
- Remove the slower scalar Metal prototype. Custom quantized GEMM, packed QKV, compile-once blocks, and quality-changing VAE defaults are intentionally excluded.

## Performance

Development measurements on an M4 Max with 36 GB unified memory, MLX 0.31.2, INT6, 832x480x124, four steps, tile 64, and 0.9 sparsity:

| Measurement | Reference VSA | SIMD VSA |
| --- | ---: | ---: |
| DiT forward | 499.3 s | 436.4 s (-12.6%) |
| Denoise peak | 21.85 GiB | 21.15 GiB |
| Sparse attention microbenchmark | 586 ms | 444 ms (-24.2%) |

SIMD is intentionally opt-in. Dynamic sparse routing produces a different valid sample from the reference path (17.7 dB video PSNR in the development comparison), so `auto` remains on reference VSA.

## Validation

- `pre-commit run --files <stacked diff>`
- `38 passed` for the focused H3 MLX VSA, parity, and fast-mode tests on MLX 0.31.2
- `38 passed` for the same tests on MLX 0.32.2
- Not run: 720p end-to-end generation or a final combined 480p generation after aggregation

### DIFF
--- a/examples/inference/basic/mlx_fasth3.py
+++ b/examples/inference/basic/mlx_fasth3.py
@@ -34,7 +34,7 @@ import argparse
 import json
 from pathlib import Path
 
-from fastvideo.mlx_runtime.minimax_h3_vsa import parse_dense_layers
+from fastvideo.mlx_runtime.minimax_h3_vsa import PUBLIC_VSA_IMPLS, parse_dense_layers
 
 
 def parse_args() -> argparse.Namespace:
@@ -93,8 +93,8 @@ def parse_args() -> argparse.Namespace:
                         help="prefix-key policy: always keep (exempt) or FLOP-matched top-k (compete)")
     parser.add_argument("--vsa-dense-first-n-steps", type=int, default=0, help="run the first N denoise steps dense")
     parser.add_argument("--vsa-dense-layers", default="", help="comma-separated layer indices forced dense")
-    parser.add_argument("--vsa-impl", choices=("auto", "reference", "metal"), default="auto",
-                        help="sparse attention implementation; auto uses chunked gather+SDPA (metal is opt-in)")
+    parser.add_argument("--vsa-impl", choices=PUBLIC_VSA_IMPLS, default="auto",
+                        help="sparse attention implementation; auto uses chunked gather+SDPA (simd is opt-in)")
     return parser.parse_args()
 
 
--- a/examples/inference/basic/mlx_fasth3.py
+++ b/examples/inference/basic/mlx_fasth3.py
@@ -80,7 +80,10 @@ def parse_args() -> argparse.Namespace:
         "--vsa",
         action=argparse.BooleanOptionalAction,
         default=False,
-        help="enable MiniMax H3 VSA; requires a VSA-capable MLX checkpoint from --include-vsa",
+        help=(
+            "enable MiniMax H3 VSA; requires a VSA-capable MLX checkpoint "
+            "from --include-vsa"
+        ),
     )
     parser.add_argument(
         "--vsa-sparsity",
@@ -88,13 +91,42 @@ def parse_args() -> argparse.Namespace:
         default=0.9,
         help="VSA sparsity in [0, 1); 0.9 is the trained FastH3 policy",
     )
-    parser.add_argument("--vsa-tile-size", type=int, default=64, choices=(64, 256), help="VSA tile size in tokens")
-    parser.add_argument("--vsa-prefix-mode", choices=("exempt", "compete"), default="exempt",
-                        help="prefix-key policy: always keep (exempt) or FLOP-matched top-k (compete)")
-    parser.add_argument("--vsa-dense-first-n-steps", type=int, default=0, help="run the first N denoise steps dense")
-    parser.add_argument("--vsa-dense-layers", default="", help="comma-separated layer indices forced dense")
-    parser.add_argument("--vsa-impl", choices=PUBLIC_VSA_IMPLS, default="auto",
-                        help="sparse attention implementation; auto uses chunked gather+SDPA (simd is opt-in)")
+    parser.add_argument(
+        "--vsa-tile-size",
+        type=int,
+        default=64,
+        choices=(64, 256),
+        help="VSA tile size in tokens",
+    )
+    parser.add_argument(
+        "--vsa-prefix-mode",
+        choices=("exempt", "compete"),
+        default="exempt",
+        help=(
+            "prefix-key policy: always keep (exempt) or FLOP-matched top-k "
+            "(compete)"
+        ),
+    )
+    parser.add_argument(
+        "--vsa-dense-first-n-steps",
+        type=int,
+        default=0,
+        help="run the first N denoise steps dense",
+    )
+    parser.add_argument(
+        "--vsa-dense-layers",
+        default="",
+        help="comma-separated layer indices forced dense",
+    )
+    parser.add_argument(
+        "--vsa-impl",
+        choices=PUBLIC_VSA_IMPLS,
+        default="auto",
+        help=(
+            "sparse attention implementation; auto uses chunked gather+SDPA "
+            "(simd is opt-in)"
+        ),
+    )
     return parser.parse_args()
 
 
--- a/examples/inference/basic/mlx_fasth3.py
+++ b/examples/inference/basic/mlx_fasth3.py
@@ -21,6 +21,11 @@ center-cropped after decode.
 
 This entrypoint currently supports text-to-video-with-audio only. It does not
 yet wire FL2VA, Ref2VA, spatial fast mode, or two-pass refinement.
+
+VSA is off by default so existing dense MLX checkpoints keep the same
+behavior. Convert with ``--include-vsa`` and pass ``--vsa`` to enable the
+sparse path. Attention activations stay BF16; INT6/INT8/INT4 apply only to
+linear weights, including the optional gate projection.
 """
 
 from __future__ import annotations
@@ -28,6 +33,8 @@ from __future__ import annotations
 import argparse
 import json
 from pathlib import Path
+
+from fastvideo.mlx_runtime.minimax_h3_vsa import parse_dense_layers
 
 
 def parse_args() -> argparse.Namespace:
@@ -69,6 +76,25 @@ def parse_args() -> argparse.Namespace:
         default=True,
         help="decode with the reference 256px overlapping VAE tiles (disable only for diagnostics)",
     )
+    parser.add_argument(
+        "--vsa",
+        action=argparse.BooleanOptionalAction,
+        default=False,
+        help="enable MiniMax H3 VSA; requires a VSA-capable MLX checkpoint from --include-vsa",
+    )
+    parser.add_argument(
+        "--vsa-sparsity",
+        type=float,
+        default=0.9,
+        help="VSA sparsity in [0, 1); 0.9 is the trained FastH3 policy",
+    )
+    parser.add_argument("--vsa-tile-size", type=int, default=64, choices=(64, 256), help="VSA tile size in tokens")
+    parser.add_argument("--vsa-prefix-mode", choices=("exempt", "compete"), default="exempt",
+                        help="prefix-key policy: always keep (exempt) or FLOP-matched top-k (compete)")
+    parser.add_argument("--vsa-dense-first-n-steps", type=int, default=0, help="run the first N denoise steps dense")
+    parser.add_argument("--vsa-dense-layers", default="", help="comma-separated layer indices forced dense")
+    parser.add_argument("--vsa-impl", choices=("auto", "reference", "metal"), default="auto",
+                        help="sparse attention implementation; auto uses chunked gather+SDPA (metal is opt-in)")
     return parser.parse_args()
 
 
@@ -95,11 +121,19 @@ def main() -> None:
         fast_factor=args.fast_factor,
         fast_sharpen=args.fast_sharpen,
         rife_weights_dir=args.rife_weights_dir,
+        vsa=args.vsa,
+        vsa_sparsity=args.vsa_sparsity,
+        vsa_tile_size=args.vsa_tile_size,
+        vsa_prefix_mode=args.vsa_prefix_mode,
+        vsa_dense_first_n_steps=args.vsa_dense_first_n_steps,
+        vsa_dense_layers=parse_dense_layers(args.vsa_dense_layers),
+        vsa_impl=args.vsa_impl,
     )
     print(json.dumps({
         "video_path": result.video_path,
         "timings_s": {k: round(v, 2) for k, v in result.timings.items()},
         "peak_memory_gib": {k: round(v, 2) for k, v in result.peak_memory_gib.items()},
+        "vsa": result.vsa,
         "audio_samples": int(result.waveform.shape[-1]),
     }, indent=2))
 
--- a/examples/inference/basic/mlx_fasth3.py
+++ b/examples/inference/basic/mlx_fasth3.py
@@ -22,9 +22,10 @@ center-cropped after decode.
 This entrypoint currently supports text-to-video-with-audio only. It does not
 yet wire FL2VA, Ref2VA, spatial fast mode, or two-pass refinement.
 
-VSA is off by default so existing dense MLX checkpoints keep the same
-behavior. Convert with ``--include-vsa`` and pass ``--vsa`` to enable the
-sparse path. Attention activations stay BF16; INT6/INT8/INT4 apply only to
+VSA is off by default; existing dense MLX checkpoints remain supported.
+H3 uses fused MLX RMSNorm, which can change BF16 rounding compared with the
+older explicit normalization path. Convert with ``--include-vsa`` and pass
+``--vsa`` to enable the sparse path. Attention activations stay BF16; INT6/INT8/INT4 apply only to
 linear weights, including the optional gate projection.
 """
 
@@ -34,7 +35,12 @@ import argparse
 import json
 from pathlib import Path
 
-from fastvideo.mlx_runtime.minimax_h3_vsa import PUBLIC_VSA_IMPLS, parse_dense_layers
+
+def _dense_layers(value: str) -> tuple[int, ...]:
+    layers = tuple(int(part.strip()) for part in value.split(",") if part.strip())
+    if any(layer < 0 for layer in layers):
+        raise argparse.ArgumentTypeError("dense layer indices must be non-negative")
+    return layers
 
 
 def parse_args() -> argparse.Namespace:
@@ -115,12 +121,13 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "--vsa-dense-layers",
-        default="",
+        type=_dense_layers,
+        default=(),
         help="comma-separated layer indices forced dense",
     )
     parser.add_argument(
         "--vsa-impl",
-        choices=PUBLIC_VSA_IMPLS,
+        choices=("auto", "reference", "simd"),
         default="auto",
         help=(
             "sparse attention implementation; auto uses chunked gather+SDPA "
@@ -158,7 +165,7 @@ def main() -> None:
         vsa_tile_size=args.vsa_tile_size,
         vsa_prefix_mode=args.vsa_prefix_mode,
         vsa_dense_first_n_steps=args.vsa_dense_first_n_steps,
-        vsa_dense_layers=parse_dense_layers(args.vsa_dense_layers),
+        vsa_dense_layers=args.vsa_dense_layers,
         vsa_impl=args.vsa_impl,
     )
     print(json.dumps({
--- a/fastvideo/mlx_runtime/minimax_h3.py
+++ b/fastvideo/mlx_runtime/minimax_h3.py
@@ -26,7 +26,13 @@ Apple Silicon memory controls:
   the repeated AdaLN projection work from each denoising step.
 - **Affine INT8, INT6, or INT4 quantization** of attention/FFN matrices
   (group size 64). Modulation, embeddings, norms, and input/output projections
-  remain in higher precision.
+  remain in higher precision. Quantization is weight-only: attention Q/K/V
+  stay BF16 (or the selected activation dtype).
+- **Optional VSA.** Dense conversion still drops
+  ``transformer_blocks.*.attn.to_gate_compress.weight``. ``--include-vsa``
+  keeps those 50 projections, quantizes them with the same affine grid, and
+  records ``vsa.capable`` in the manifest. Runtime VSA is opt-in and never
+  enabled for dense-only checkpoints.
 
 Checkpoint layout: the released H3 checkpoint uses the diffusers reference
 module names 1:1 (``transformer_blocks.{i}.attn.to_out.0.weight``,
@@ -61,6 +67,18 @@ from fastvideo.mlx_runtime.fastwan import (
     silu,
     timestep_embedding,
     weight_dtype,
+)
+from fastvideo.mlx_runtime.minimax_h3_vsa import (
+    MiniMaxH3VSAConfig,
+    MiniMaxH3VSAGeometry,
+    MiniMaxH3VSAStats,
+    VSA_GATE_KEY_SUFFIX,
+    build_h3_tile_geometry,
+    dense_only_vsa_error,
+    dit_seq_shape_from_layout,
+    geometry_is_supported,
+    h3_vsa_attention,
+    prefix_segments_from_layout,
 )
 
 logger = init_logger(__name__)
@@ -107,18 +125,22 @@ _QUANTIZABLE_SUFFIXES = (
     "ff.net.0.proj.weight",
     "ff.net.2.weight",
 )
+_VSA_QUANTIZABLE_SUFFIXES = _QUANTIZABLE_SUFFIXES + (VSA_GATE_KEY_SUFFIX, )
 
-# The released FastH3 student carries VSA routing projections. The MLX path is
-# dense, so retaining these weights wastes about 3.6 GiB without affecting a
-# single output value.
+# The released FastH3 student carries VSA routing projections. Dense MLX
+# conversion drops them (~3.6 GiB BF16) because they do not affect fused SDPA.
+# Pass include_vsa=True to keep and quantize them.
 _IGNORED_DENSE_KEY_PARTS = ("attn.to_gate_compress", )
 
 
-def _is_quantizable(key: str) -> bool:
-    return key.endswith(_QUANTIZABLE_SUFFIXES)
+def _is_quantizable(key: str, *, include_vsa: bool = False) -> bool:
+    suffixes = _VSA_QUANTIZABLE_SUFFIXES if include_vsa else _QUANTIZABLE_SUFFIXES
+    return key.endswith(suffixes)
 
 
-def _is_ignored_dense_key(key: str) -> bool:
+def _is_ignored_dense_key(key: str, *, include_vsa: bool = False) -> bool:
+    if include_vsa:
+        return False
     return any(part in key for part in _IGNORED_DENSE_KEY_PARTS)
 
 
@@ -473,7 +495,38 @@ def apply_h3_rotary(x, cos, sin):
     return mx.concatenate([out.astype(x.dtype), x_pass], axis=-1)
 
 
-def _attention(weights: dict[str, Any], x, cos, sin, *, num_heads: int, head_dim: int, eps: float, use_rope: bool):
+def _gate_weight_is_active(weight: Any) -> bool:
+    import mlx.core as mx
+
+    if weight is None:
+        return False
+    if isinstance(weight, QuantizedMatrix):
+        # Affine dequant is scale * code + bias. Packed codes with a zero scale
+        # do not contribute; a zero-scale / nonzero-bias group is a constant.
+        active = mx.any(weight.scales != 0)
+        if weight.biases is not None:
+            active = active | mx.any(weight.biases != 0)
+        return bool(active.item())
+    return bool(mx.any(weight != 0).item())
+
+
+def _attention(
+    weights: dict[str, Any],
+    x,
+    cos,
+    sin,
+    *,
+    num_heads: int,
+    head_dim: int,
+    eps: float,
+    use_rope: bool,
+    vsa_geometry: MiniMaxH3VSAGeometry | None = None,
+    vsa_sparsity: float = 0.0,
+    vsa_exempt: bool = True,
+    vsa_impl: str = "auto",
+    vsa_stats: MiniMaxH3VSAStats | None = None,
+    use_gate_compress: bool = False,
+):
     """H3 attention: bias-free qkv, per-head qk RMSNorm, partial RoPE."""
     import mlx.core as mx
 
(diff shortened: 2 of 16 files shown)


fastvideo/mlx_runtime/minimax_h3_vsa.py
```diff
@@ -0,0 +1,704 @@
...
+    allow = block_mask[:, token_tile][:, :, token_tile] & token_valid[None, None, :]
+    return allow
+
+
+def _reference_token_sdpa(q_tiled, k_tiled, v_tiled, block_mask: np.ndarray, geometry: MiniMaxH3VSAGeometry,
+                          scale: float):
+    import mlx.core as mx
+
+    allow = _token_mask_from_block_mask(block_mask, geometry)
+    bias = mx.where(mx.array(allow), mx.array(0.0, dtype=mx.float32), mx.array(-1e9, dtype=mx.float32))
```
SolitaryThinker: **[blocker] fp32 additive mask crashes bf16 SDPA on small canvases**

`_reference_token_sdpa` builds `bias` as float32 (`mx.where(..., mx.array(0.0, dtype=mx.float32), mx.array(-1e9, dtype=mx.float32))`) and passes it as an additive mask while q/k/v are the DiT's bf16 activations (the converter hardcodes bf16 and `transformer_blocks.*` are never kept fp32). MLX 0.32.2 rejects any non-bool mask whose dtype doesn't promote to the q/k/v type — `mlx/fast.cpp` throws `ValueError: [scaled_dot_product_attention] Mask type must promote to output type bfloat16.` eagerly at op construction, so there's no fallback.

Every geometry that satisfies `_reference_full_mask_fits` (num_tiles ≤ 24) hits this with a converted checkpoint, e.g. `mlx_fasth3.py --vsa --height 128 --width 128` at defaults → 18 tiles → crash in step 0. 480p/720p take the bool-masked gather path, which is why it wasn't seen. All current tests drive this path with fp32 tensors (`mx.random.normal` default).

Fix: pass the bool `allow` mask directly (as `_reference_gather_sdpa` does — also avoids materializing up to a 1 GiB fp32 bias at the 256M-element cap) or `bias.astype(q_tiled.dtype)`, and add a bf16 test on a ≤24-tile geometry.

fastvideo/mlx_runtime/minimax_h3_vsa.py
```diff
@@ -0,0 +1,704 @@
...
+            exempt,
+        )
+        chosen = resolve_impl(impl, dim, geometry.tile_elems)
+        prefix_out = _dense_sdpa(query[:geometry.prefix_length], key, value, scale)
+        n_prefix_pad = geometry.num_prefix_tiles * geometry.tile_elems
+        if chosen == "simd":
+            from fastvideo.mlx_runtime.minimax_h3_vsa_simd import simd_block_sparse
+
+            try:
+                video_tiled = simd_block_sparse(q_tiled, k_tiled, v_tiled, block_idx, block_num, geometry, scale)
```
SolitaryThinker, line 706: **[blocker] The SIMD→reference fallback cannot catch Metal kernel failures**

This `try/except` only catches eager Python-side errors. `simd_block_sparse` returns an unevaluated `CustomKernel` array; MLX compiles and runs the Metal source inside `CustomKernel::eval_gpu` (verified in MLX 0.32.2: `metal_kernel()` construction and `__call__` never touch Metal, `get_library` compiles at eval, and `transforms.cpp` rethrows from `mx.eval`). So a compile or runtime kernel failure surfaces at `mx.eval(packed)` in `forward_with_cache` — outside this `except` — and aborts generation, contradicting `docs/attention/vsa/index.md`'s "falls back to reference on … compilation failure". On current M-series hardware the kernel source looks fine, so this half is latent.

Reproducible today on any no-Metal MLX build (Linux `mlx` wheel): `simd_kernel_available()` returns True (construction never touches Metal), the kernel call raises `[metal_kernel] No Metal back-end.` eagerly, is swallowed here, and `test_simd_kernel_parity_with_reference_identical_tiles` passes vacuously comparing reference vs reference instead of skipping — while every sparse block of every step re-attempts the kernel and logs a fresh warning (nothing ever sets `_SIMD_KERNEL_ERROR` on failure; the `resolve_impl` warnings are also per-call).

Fix: probe once (build + run a tiny tile-64/head-128 shape + `mx.eval`), memoize the failure reason in `_SIMD_KERNEL_ERROR`, and use `warning_once`.

fastvideo/mlx_runtime/minimax_h3.py
```diff
@@ -712,6 +809,56 @@ def precompute_adaln(self, timesteps: np.ndarray, *, drop_weights: bool = True)
                 block["adaln_proj.linear.bias"] = None
         return self._adaln_cache
 
+    @property
+    def vsa_capable(self) -> bool:
+        return self._vsa_capable
+
+    def configure_vsa(self, config: MiniMaxH3VSAConfig | None) -> None:
```
SolitaryThinker, line 817: **[blocker] `configure_vsa` assigns before validating and never invalidates `_vsa_geometry`**

Two issues:
1. `self.vsa_config` is assigned before the `vsa_capable` check, so a dense-only DiT that raised `DenseOnlyVSACheckpointError` is left with `enabled=True`. A caller that catches the error and retries with `vsa_config=None` (the pipeline skips `configure_vsa` then) hits `dit.vsa_config.enabled` → `prepare_vsa_geometry` → sparsity 0.9 with `use_gate_compress=False`: pure top-k routing without trained gates on a dense-only checkpoint, silently.
2. `_vsa_geometry` is never cleared, and `forward_with_cache`'s lazy prepare only runs `if self._vsa_geometry is None`, so `configure_vsa(tile_size=64)` → forward → `configure_vsa(tile_size=256)` → forward silently reuses the tile-64 geometry; likewise a 480x832 → 832x480 swap shares `total_seq_length` and passes `_attention`'s guard with the wrong tiling.

`generate()` is only safe because it loads a fresh DiT per call. Fix: validate before assigning, set `self._vsa_geometry = None` in `configure_vsa`, and key the lazy prepare on the layout.

fastvideo/mlx_runtime/minimax_h3_vsa_simd.py
```diff
@@ -0,0 +1,311 @@
...
+        if (active && g0 < q_valid) {
+            v0 = float(q[q_base_tile + g0 * D + kk * 8 + c0]);
+        }
+        if (active && g1 < q_valid) {
+            v1 = float(q[q_base_tile + g1 * D + kk * 8 + c0]);
+        }
+        sg_qtile[lane] = v0;
+        sg_qtile[lane + 32] = v1;
+        simdgroup_barrier(mem_flags::mem_threadgroup);
+        simdgroup_load(qfrag[kk], sg_qtile, 8);
```
SolitaryThinker, line 92: **[blocker] Missing WAR `simdgroup_barrier` in the Q-prologue loop**

Iteration kk+1's per-lane stores `sg_qtile[lane] = v0; sg_qtile[lane + 32] = v1;` have no barrier after iteration kk's cooperative `simdgroup_load(qfrag[kk], sg_qtile, 8)` — a cross-lane write-after-read race under the Metal memory model (the lane→address maps differ: lane L stores {L, L+32} while the 8x8 fragment layout reads e.g. lane 0 → {0,1}, lane 16 → {32,33}). Every other `sg_qtile` reuse point in this kernel (lines 151, 187, 211) is guarded.

Correct on today's lockstep Apple GPUs, but undefined per the spec — an implementation that lets the stores overtake the load corrupts every Q·K score for that SIMD group's 8 rows with no error.

Fix: add `simdgroup_barrier(mem_flags::mem_threadgroup);` at the top of the loop body (16 barriers at kernel start; negligible cost).

scripts/checkpoint_conversion/convert_minimax_h3_mlx.py
```diff
@@ -76,10 +93,17 @@ def main() -> None:
...
-            print(f"[skip] {fmt} already converted at {out_dir}", flush=True)
-            continue
-        print(f"[convert] {fmt} (dtype={dtype}, spec={spec})", flush=True)
+        already = (out_dir / H3_MANIFEST_FILENAME).exists() and (out_dir / H3_WEIGHTS_FILENAME).exists()
+        if already:
+            capable = mlx_h3_checkpoint_vsa_capable(out_dir)
+            if bool(capable) == bool(args.include_vsa):
+                print(f"[skip] {fmt} already converted at {out_dir} (vsa.capable={capable})", flush=True)
+                continue
+            raise SystemExit(
```
SolitaryThinker: **`SystemExit` mid-loop breaks multi-format re-runs that used to be idempotent**

The old skip-if-exists loop converted missing formats and skipped existing ones. Now the first dense/VSA mismatch aborts before the remaining formats convert — and nothing stops mixed dense/VSA formats from accumulating in the same `--out` in the first place (the check only runs for formats that already exist). E.g.: run A `--formats int8` (dense), run B `--formats int6 --include-vsa` (succeeds, no check), run C `--formats 'int8 int6 int4'` → int8 skipped, int6 mismatches → `SystemExit`, int4 never converted. No re-run of C can complete in either mode.

Fix: skip mismatched existing formats with a warning (or validate all requested formats up front) instead of exiting mid-loop.

fastvideo/mlx_runtime/minimax_h3.py
```diff
@@ -712,6 +809,56 @@ def precompute_adaln(self, timesteps: np.ndarray, *, drop_weights: bool = True)
...
+        self._gate_active[block_index] = active
+        if weight is not None and not active:
+            logger.info_once("MiniMax H3 VSA compression gate is all-zero; skipping the pooled branch.")
+        return active
+
+    def _vsa_block_kwargs(self, block_index: int, step_index: int) -> dict[str, Any]:
+        geometry = self._vsa_geometry if self.vsa_config.enabled else None
+        sparsity = self.vsa_config.layer_sparsity(block_index, step_index) if geometry is not None else 0.0
+        stats = None
+        if block_index == 0:
```
SolitaryThinker: **`GenerationResult.vsa` stats describe only block 0 of the last step, with the wrong formula in compete mode**

The stats object is allocated only for `block_index == 0` on the final step, and `h3_vsa_attention` fills it from that one call. So: with `--vsa-dense-layers 0` (or `--vsa-dense-first-n-steps` = num steps) the report shows `impl='dense'`, `configured_sparsity=0.0` and default tile/prefix values even though 49 layers ran sparse; a SIMD failure on any block ≥ 1 is invisible; and in compete mode each query keeps between `k_vid` and `k_vid + num_prefix_tiles` video tiles (`k_sel = min(k_vid + P, n_tiles)`) yet `achieved_sparsity` reports the exempt-mode `1 - k_vid/num_video_tiles`. The example prints this as authoritative.

Fix: fill `configured_sparsity`/`tile_size`/`prefix_mode` from `vsa_config`, and aggregate `impl`/fallback across blocks and steps.

fastvideo/mlx_runtime/minimax_h3.py
```diff
@@ -773,6 +920,7 @@ def forward(
                 head_dim=self.head_dim,
                 norm_eps=self.norm_eps,
                 qk_norm_eps=self.qk_norm_eps,
+                **self._vsa_block_kwargs(block_index, 0),
```
SolitaryThinker, line 943: **`forward()` silently ignores an enabled VSA config**

This path threads `**self._vsa_block_kwargs(block_index, 0)` but never calls `prepare_vsa_geometry` and hard-codes step 0. With `configure_vsa(MiniMaxH3VSAConfig(enabled=True, ...))`, `_vsa_block_kwargs` sees `_vsa_geometry is None` → sparsity 0.0, `use_gate_compress=False` → plain SDPA in all blocks, no warning. `forward()` and `forward_with_cache()` now compute different math for the same weights, and a VSA parity test written on top of `tiny_h3.mlx_output` would compare dense against dense and pass vacuously.

Fix: raise when `vsa_config.enabled` and `_vsa_geometry is None` (or accept layout/step_index), or document `forward()` as dense-only.

fastvideo/mlx_runtime/minimax_h3_vsa.py
```diff
@@ -0,0 +1,704 @@
...
+    attn = mx.softmax(scores, axis=-1)
+    out_c = attn @ v_pooled  # [H, n_tiles, D]
+    out_c = out_c.transpose(1, 0, 2)  # [n_tiles, H, D]
+    heads, dim = gate_tiled.shape[1], gate_tiled.shape[2]
+    gate_view = gate_tiled.reshape(geometry.num_tiles, geometry.tile_elems, heads, dim)
+    return (out_c[:, None, :, :] * gate_view).reshape(geometry.padded_length, heads, dim)
+
+
+def resolve_impl(requested: VSAImpl, head_dim: int, tile_elems: int) -> str:
+    if requested in ("reference", "auto") and not _AUTO_PREFERS_SIMD:
```
SolitaryThinker: **`--vsa-impl reference` will route to SIMD the moment `_AUTO_PREFERS_SIMD` flips**

An explicit `requested='reference'` is honoured only because `_AUTO_PREFERS_SIMD` is currently False: `if requested in ("reference", "auto") and not _AUTO_PREFERS_SIMD` sends `'reference'` down the SIMD-eligibility path once the flag is flipped (as the comment at line 48 plans), silently running the kernel that "does not yet match reference video" for the documented correctness baseline. `test_simd_unsupported_tile_falls_back_to_reference` only covers `'auto'` and unsupported shapes.

Fix: `if requested == "reference" or (requested == "auto" and not _AUTO_PREFERS_SIMD): return "reference"`, plus a test for `requested='reference'` on tile 64 / head dim 128.

fastvideo/mlx_runtime/minimax_h3_pipeline.py
```diff
@@ -576,28 +608,46 @@ def mux(self,
...
+            vsa_dense_layers: tuple[int, ...] = (),
+            vsa_impl: str = "auto",
     ) -> GenerationResult:
         timings: dict[str, float] = {}
         peaks: dict[str, float] = {}
 
         if fast_sharpen < 0:
             raise ValueError(f"fast_sharpen must be non-negative, got {fast_sharpen}.")
         _validate_checkpoint_step_ladder(self.dit_checkpoint, num_steps)
+        vsa_config = MiniMaxH3VSAConfig(
```
SolitaryThinker, line 646: **Dense runs abort on out-of-range `--vsa-*` knobs they never use**

`MiniMaxH3VSAConfig` (and its `__post_init__` validation) is constructed unconditionally before the `enabled` check, and the example CLI always forwards every `--vsa-*` value. So `mlx_fasth3.py ... --vsa-sparsity 1.0` *without* `--vsa` dies with `ValueError: VSA sparsity must be in [0, 1)` before the dense generation starts.

Fix: build the config only when `vsa` is set, or validate lazily in `configure_vsa`.

fastvideo/mlx_runtime/minimax_h3.py
```diff
@@ -473,27 +494,86 @@ def apply_h3_rotary(x, cos, sin):
...
-def _attention(weights: dict[str, Any], x, cos, sin, *, num_heads: int, head_dim: int, eps: float, use_rope: bool):
+def _gate_weight_is_active(weight: Any) -> bool:
+    import mlx.core as mx
+
+    if weight is None:
+        return False
+    if isinstance(weight, QuantizedMatrix):
+        # Affine dequant is scale * code + bias. Packed codes with a zero scale
+        # do not contribute; a zero-scale / nonzero-bias group is a constant.
+        active = mx.any(weight.scales != 0)
```
SolitaryThinker: **Quantized gate is always "active": MLX affine quantization clamps scales to ≥ 1e-7**

`mx.any(weight.scales != 0)` can never be False for a `QuantizedMatrix`: MLX's affine quantizer computes `scale = max((w_max - w_min) / n_bins, eps)` with `eps = 1e-7` (verified in MLX 0.32.2, `quantized.h:2665` / CPU `quantized.cpp:1184`), nonzero in bf16/fp16/fp32. So every `--include-vsa` int8/int6/int4 gate is classified active, the all-zero-gate skip and its `info_once` log are unreachable, and (via `use_vsa = ... or gate is not None`) dense-first steps and dense layers run the full tile/pool/score/softmax/gate branch that contributes exactly zero. The torch model tests the raw weight, so the two runtimes diverge on this check. Output is unaffected, and shipped FastH3-Preview gates are trained — this only wastes compute on hypothetical zero-gate exports.

Fix: for affine quantization test `mx.any(weight.weight != 0) | mx.any(weight.biases != 0)` (an all-zero group yields codes 0 and bias 0), or record activeness from the raw weight at conversion time in the manifest.

fastvideo/mlx_runtime/minimax_h3.py
```diff
@@ -473,27 +494,86 @@ def apply_h3_rotary(x, cos, sin):
...
+            active = active | mx.any(weight.biases != 0)
+        return bool(active.item())
+    return bool(mx.any(weight != 0).item())
+
+
+def _h3_rms_norm(x, weight, eps: float):
+    """Use MLX's fused normalization kernel for H3's last-axis RMSNorm."""
+    import mlx.core as mx
+
+    return mx.fast.rms_norm(x, weight, eps)
```
SolitaryThinker, line 514: **`_h3_rms_norm` silently bypasses the documented `FASTVIDEO_MLX_FAST_NORM` opt-in**

All nine H3 norm sites now call `mx.fast.rms_norm` unconditionally, replacing `fastwan.rms_norm` whose docstring promises the numerically-explicit path stays the baseline unless `FASTVIDEO_MLX_FAST_NORM=1`. The env toggle is now a no-op for H3 (while still toggling FastWan), and dense H3 output on Metal is no longer bit-reproducible against the pre-PR runtime despite the PR's "existing dense MLX checkpoints keep the same behavior" framing. The fp32 parity test (atol 2e-4, CPU fallback in CI) can't see a bf16-rounding-level drift.

Fix: route through `fastwan.rms_norm` (honouring the toggle), or make the fused path the documented H3 default and update the fastwan docstring/example text accordingly.

fastvideo/tests/mlx/test_mlx_minimax_h3_vsa.py
```diff
@@ -0,0 +1,449 @@
...
+    geo = build_h3_tile_geometry(spec["prefix_segments"], _dit_shape(spec), 256)
+    q, k, v = _tiny_qkv(geo.total_seq_length, seed=5)
+    mx.random.seed(6)
+    gate = mx.random.normal(q.shape) * 0.1
+    with_gate = h3_vsa_attention(q, k, v, geo, sparsity=0.5, gate_compress=gate, impl="reference")
+    without = h3_vsa_attention(q, k, v, geo, sparsity=0.5, gate_compress=None, impl="reference")
+    assert not mx.allclose(with_gate, without, atol=1e-5, rtol=1e-5).item()
+
+
+def test_int6_converter_round_trip_fifty_gate_matrices() -> None:
```
SolitaryThinker, line 231: **The `--include-vsa` convert→save→load path has no test, and this file isn't wired into CI**

This test quantizes 50 random matrices and checks two key predicates — it never exercises `mlx_h3_dit_from_diffusers_safetensors(include_vsa=True)`, `save_mlx_h3_checkpoint` with a gate key, `load_mlx_h3_checkpoint` on a VSA export, or `mlx_h3_checkpoint_vsa_capable` on a capable manifest. The gate routing through `assign`, the `missing_gates` KeyError, the manifest `vsa.capable` flag, and the `blocks.{i}.attn.to_gate_compress.weight` round trip all have zero coverage (only the no-gate/False branches are asserted). And `.github/workflows/ci-macos-mlx.yml` enumerates test files explicitly and doesn't list this file, so none of the new tests run in CI (`test_mlx_minimax_h3_parity.py` is missing from that lane too).

Suggested: a one-block safetensors fixture with `transformer_blocks.0.attn.to_gate_compress.weight`, run through convert → save → load, asserting `vsa_capable` and a `QuantizedMatrix` gate, plus the negative KeyError case; add the file to `ci-macos-mlx.yml`.

examples/inference/basic/mlx_fasth3.py
```diff
@@ -29,6 +34,8 @@
 import json
 from pathlib import Path
 
+from fastvideo.mlx_runtime.minimax_h3_vsa import PUBLIC_VSA_IMPLS, parse_dense_layers
```
SolitaryThinker: **Module-level `fastvideo` import pulls in torch before argparse runs**

This import executes `fastvideo/__init__.py` (→ torch) and all of `fastvideo/mlx_runtime/__init__.py` before `parse_args()`, undoing the pre-PR pattern where the only fastvideo import sat inside `main()`. `--help` now pays the full torch + MLX import (and fails with ImportError instead of printing usage on a machine without torch/diffusers).

Fix: import inside `main()` after `parse_args()`, or inline the three-element choices tuple and a comma split.

fastvideo/mlx_runtime/minimax_h3_vsa.py
```diff
@@ -0,0 +1,704 @@
...
+            raise ValueError(f"VSA sparsity must be in [0, 1), got {self.sparsity}.")
+        if self.tile_size not in VSA_H3_TILE_SHAPES:
+            raise ValueError(f"VSA tile_size must be one of {sorted(VSA_H3_TILE_SHAPES)}, got {self.tile_size}.")
+        if self.prefix_mode not in VSA_PREFIX_MODES:
+            raise ValueError(f"VSA prefix_mode must be one of {VSA_PREFIX_MODES}, got {self.prefix_mode!r}.")
+        if self.impl not in VSA_IMPLS:
+            raise ValueError(f"VSA impl must be one of {VSA_IMPLS}, got {self.impl!r}.")
+        if self.dense_first_n_steps < 0:
+            raise ValueError(f"vsa_dense_first_n_steps must be >= 0, got {self.dense_first_n_steps}.")
+        object.__setattr__(self, "dense_layers", tuple(int(layer) for layer in self.dense_layers))
```
SolitaryThinker, line 89: **`dense_layers="10"` is silently split into digits `(1, 0)`**

`tuple(int(layer) for layer in self.dense_layers)` iterates a string character by character, so `MiniMaxH3VSAConfig(dense_layers='10')` (or `generate(..., vsa_dense_layers='10')`) makes layers 0 and 1 dense and leaves layer 10 sparse with no error. Negative or out-of-range indices are also accepted and silently never match (the non-negative check lives only in the CLI's `parse_dense_layers`).

Fix: reject `str` and validate `0 <= layer < num_blocks` here (and consider reusing the existing `parse_int_list` in the CLI).

SolitaryThinker (commented):
Automated review (Claude Code, max effort — 10 finder angles, adversarially verified per finding; posted on the author's behalf by @SolitaryThinker). 15 findings as inline comments, most severe first per file.

**Overall**: the port's core math is faithful — tile geometry, routing, scoring, the gate-compress branch, and packed-sequence order were traced against the PyTorch VSA-H3 backend and match; the gate key round-trips through converter/saver/loader; and removing the `mx.contiguous` guards from dense SDPA was checked and is safe (the SSIM-validated FastMetal Wan path feeds identically strided views to the same fused kernel), though the now-contradictory guards remain in `minimax_h3_vsa.py:388`, the video/audio VAEs — worth making consistent.

**Four findings look like merge-blockers**:
1. the small-canvas reference path (≤24 tiles, e.g. any canvas ≤128px per side) passes an fp32 additive mask to bf16 SDPA and crashes every converted checkpoint;
2. the documented SIMD→reference fallback structurally cannot catch Metal compile/runtime failures (MLX compiles lazily at `mx.eval`, outside the `try`);
3. `configure_vsa` assigns before validating and never invalidates cached geometry;
4. a missing WAR `simdgroup_barrier` in the Metal kernel's Q prologue (latent on today's lockstep hardware).

**Coverage**: `test_mlx_minimax_h3_vsa.py` isn't listed in `ci-macos-mlx.yml` (nor is `test_mlx_minimax_h3_parity.py`), and no test exercises the `--include-vsa` convert→save→load path, so CI currently catches none of the above.

MLX-internal claims (SDPA mask dtype rejection, lazy kernel compile, affine scale clamp, bf16 scatter atomics) were verified against the pinned MLX 0.32.2 source, not from memory. Four additional candidates were refuted during verification and are not reported; several smaller cleanups were cut by the finding cap.

fastvideo/mlx_runtime/minimax_h3_vsa.py
```diff
@@ -0,0 +1,704 @@
...
+
+def _tile_hidden(x, geometry: MiniMaxH3VSAGeometry):
+    """Scatter packed rows ``[S, H, D]`` into the zero-padded tile buffer."""
+    import mlx.core as mx
+
+    seq, heads, dim = x.shape
+    if seq != geometry.total_seq_length:
+        raise ValueError(f"VSA-H3 metadata was built for sequence length {geometry.total_seq_length}, got {seq}.")
+    buf = mx.zeros((geometry.padded_length, heads, dim), dtype=x.dtype)
+    buf = buf.at[mx.array(geometry.untile_combined_index)].add(x)
```
SolitaryThinker: **[perf] `_tile_hidden` uses a contended bf16 scatter-add where a plain gather is bit-identical**

`mx.zeros(...).at[idx].add(x)` on bf16 goes through MLX's packed-32-bit CAS loop (bf16 isn't in `is_metal_atomic`; adjacent lanes contend on the same word), plus a full zero-fill and a fresh host→device int64 index upload — and this runs up to 5× per block per step (q, k, v, prefix re-tile, gate). Since `untile_combined_index` is validated injective, a single gather with a precomputed padded index is exactly equivalent: per 4-step generation this is roughly 106G (480p) / 246G (720p) contended CAS updates and ~260/560 GB of zero-fill traffic. The prefix call at line 677 is the worst instance — it concatenates `prefix_out` with a full-sequence zeros block and scatter-tiles all ~15-34k rows to keep ~512 slots. Note `buf[idx] = x` wouldn't help: MLX's None-scatter on 16-bit types uses the same CAS loop; only a gather avoids atomics.

Fix: precompute per geometry a `tile_gather_index` of length `padded_length` (inverse permutation; pad slots → an appended zero row), cache as `mx.array`, and tile with `mx.concatenate([x, zeros((1,H,D))])[tile_gather_index]`; build `prefix_tiled` from `untile_combined_index[:prefix_length]` only. Lower-impact hot-path costs in the same family: the per-chunk `mx.eval` in the gather loop (~1.7k/7k blocking evals per generation) and `_pool_tiles`' three full fp32 copies per block.

fastvideo/mlx_runtime/minimax_h3_vsa.py
aryan5v: Replaced the FP32 additive bias with the boolean allow mask in ed0dd009f. Added BF16 small-canvas tests for both prefix modes and comparison with the gathered reference path; they pass on Metal.

fastvideo/mlx_runtime/minimax_h3_vsa.py
aryan5v: Fixed in ed0dd009f: availability now executes and evaluates a tiny probe for each supported dtype, caches failures, and warns once. The runtime evaluation is inside the fallback boundary too. Tests cover no-Metal detection, actual lazy Metal compilation failure, cached runtime failure, and assert that SIMD parity really executed SIMD rather than reference fallback.

fastvideo/mlx_runtime/minimax_h3.py
aryan5v: Fixed in ed0dd009f. Configuration validates before mutation, clears geometry on success, and cached forwards check a geometry key containing the prefix segments, video layout, and tile size. Regression tests include failed configuration preserving state, tile-size changes, and actual cached forwards with same-length swapped spatial axes.

fastvideo/mlx_runtime/minimax_h3_vsa_simd.py
aryan5v: Added the threadgroup-memory simdgroup barrier at the top of the Q-prologue loop in ed0dd009f. Actual SIMD execution/parity tests pass on Metal.

scripts/checkpoint_conversion/convert_minimax_h3_mlx.py
aryan5v: Fixed in ed0dd009f: an existing mismatched format is preserved and skipped with a warning; conversion continues for missing formats. Added a mixed dense/VSA rerun regression that verifies no overwrite and conversion of the remaining format.

fastvideo/mlx_runtime/minimax_h3.py
aryan5v: Fixed in ed0dd009f. Configuration fields come from the requested VSA config; per-call stats aggregate across all blocks and steps, including mixed implementations and later fallbacks. Compete mode counts the selected video columns rather than assuming exempt-mode selection. Tests cover dense overrides, multiple steps, later failures, and independently computed compete-mode counts.

fastvideo/mlx_runtime/minimax_h3.py
aryan5v: Made forward() explicitly dense-only in ed0dd009f. It raises when VSA is enabled and directs callers to forward_with_cache(), which has layout and step information. Added a regression asserting the error before computation.

fastvideo/mlx_runtime/minimax_h3_vsa.py
aryan5v: Fixed in ed0dd009f. Explicit reference selection returns reference unconditionally; a regression flips the auto preference and asserts that reference does not even probe SIMD.

fastvideo/mlx_runtime/minimax_h3_pipeline.py
aryan5v: Fixed in ed0dd009f. Disabled VSA uses a disabled default config, without validating unused API options. A pipeline regression verifies that dense generation reaches media preflight despite invalid unused VSA parameters, while enabled VSA still rejects them.

fastvideo/mlx_runtime/minimax_h3.py
aryan5v: Fixed in ed0dd009f by checking packed codes and affine biases, not nonzero scales. INT4/INT6/INT8 regressions distinguish quantized all-zero gates from nonzero constant gates.

fastvideo/mlx_runtime/minimax_h3.py
aryan5v: Used the documented-default option in ed0dd009f: H3 retains fused RMSNorm, and the H3 helper, VSA guide, CLI example, and FastWan helper documentation now distinguish H3 from the Wan-only FASTVIDEO_MLX_FAST_NORM toggle. Removed the same-behavior claim and explicitly documented possible BF16 rounding differences.

fastvideo/tests/mlx/test_mlx_minimax_h3_vsa.py
aryan5v: Added two-block safetensors convert/save/load regressions for INT4/INT6/INT8 in ed0dd009f, checking exact packed gate tensors, manifest capability, actual loaded QuantizedMatrix gates, and missing-gate rejection. Commit 90c5cf8a1 adds H3 parity, VSA, these regressions, and fast-mode tests to both CI lanes.

examples/inference/basic/mlx_fasth3.py
aryan5v: Fixed in ed0dd009f. CLI parsing now uses only stdlib code and literal implementation choices; runtime imports happen afterward. Subprocess tests run --help and invalid dense-layer arguments with python -S, without installed dependencies.

fastvideo/mlx_runtime/minimax_h3_vsa.py
aryan5v: Fixed in ed0dd009f: config rejects strings, booleans, non-integers, and negative indices; configure_vsa checks the block-count upper bound before mutation. The CLI parses its comma-separated string through argparse before constructing the pipeline. Added API and CLI regressions.

fastvideo/mlx_runtime/minimax_h3_vsa.py
aryan5v: Implemented cached inverse gather indices with an appended zero row in ed0dd009f, including a prefix-only gather that avoids full-sequence prefix retiling. Ragged/padded FP32, FP16, and BF16 tests are bit-identical to the old scatter reference. No production latency claim yet. Kept per-chunk evaluation and FP32 pooling unchanged to avoid changing memory bounds or numerical behavior in this fix.

PR #1780, opened 2026-08-28
[bugfix]: apply FP8 to MiniMax H3's feed-forward

KyleNeverGivesUp:
## Problem

`FP8Config.get_quant_method` chooses what to quantize by substring-matching a layer's state-dict name against `_FP8_SUFFIXES`. That tuple carried `ffn.fc_in` and `ffn.fc_out`, which fits WanVideo and LTX-2. MiniMax H3 mounts its feed-forward on `.ff`, so the layer is `transformer_blocks.0.ff.fc_in` and `"ffn.fc_in" in "transformer_blocks.0.ff.fc_in"` is False.

All 104 H3 feed-forward linears were skipped, 50 main blocks and 2 refiner blocks at two linears each, and kept their bf16 weights.

Nothing surfaced it. H3's attention projections are named `to_q`, `to_k`, `to_v` and `to_out`, which the tuple already matched, so quantization ran without error, freed real memory, and produced correct video while covering 8.01B of the checkpoint's 22.09B parameters. Partial coverage was indistinguishable from full coverage.

Every model added so far has had its feed-forward registered, including Kandinsky5's in #1601. H3 is the only one where it is missing, so this is an omission against a consistent pattern rather than a deliberate exclusion.

## What this changes

**The two H3 names are matched by true suffix, not substring.** They live in a separate `_FP8_EXACT_SUFFIXES` tuple checked with `endswith`. `DiTConfig.prefix` is settable from the CLI, so under a substring rule a root prefix containing `.ff.fc_in` would tag every descendant, including `adaln_proj.linear`, whose forward reads `self.linear.weight.dtype` after the converter has removed that weight. The original tuple keeps substring matching because it depends on it: `to_q` is also how Kandinsky5's `to_query` is reached.

WanVideo and GLM-Image are unaffected even though their state dicts also carry `.ff.fc_in`. Both build that module without passing `quant_config` or a `prefix`, so `get_quant_method` is never consulted for it. `param_names_mapping` only renames checkpoint keys during weight loading and takes no part in quantization selection.

**Partial coverage is now reported.** `convert_model_to_fp8` counts what it quantized, what was handed an `FP8Config` but matched no entry, and what was never a candidate at all, then logs all three with the distinct trailing names of the unmatched layers. On H3 that reads:

```
FP8: quantized 312 linear layers holding 20.04B parameters; left 108 requested
linear layers holding 2.05B parameters in their loaded dtype because their names
match no _FP8_SUFFIXES entry: adaln_proj.linear, attn.to_gate_compress,
audio_proj_in, audio_proj_out, context_embedder, norm_out.linear, proj_in,
proj_out, time_embedder.fc_in, time_embedder.fc_out; 1 further linear layers
were never handed an FP8Config
```

The names are exactly what a new `_FP8_SUFFIXES` entry would have to spell, so the next model with different naming is one log line away from a fix rather than a code read.

**Zero coverage warns.** `_maybe_quantize_model` only calls the converter once it finds a layer carrying an `FP8QuantizeMethod`. If FP8 is requested and not one layer matches, the converter never runs and the log above never appears, so the worst case was the quietest. It now warns and names the layers it saw.

## Measured

One GB10, 121 GiB unified memory, and [KyleNeverGivesUp/FastH3-4-step-Preview-v1-r16](https://huggingface.co/KyleNeverGivesUp/FastH3-4-step-Preview-v1-r16), which is FastH3 Preview v1 with the rank-reduced AdaLN converter applied, at 22.09B parameters. Every number below can be reproduced against that checkpoint.

| | before | after |
|---|---|---|
| quantized parameters | 8.01B | 20.04B |
| quantized linear layers | 208 | 312 |
| resident weights | 44.23 GiB | 32.99 GiB |

**This is not a speed lever, and the patch does not claim to be one.** Generation is 1587 s against 1643 s at 512x896 with 124 frames, which is noise. That agrees with the existing guidance in `docs/getting_started/installation/spark_performance.md` that quantizing linear layers buys almost nothing on an attention-bound video model. What changes is what fits: on this machine the 11.24 GiB is the difference between a workload that is killed and one that completes.

## Output quality

Same seed, resolution and frame count, with `--fp8` as the only variable.

SSIM is 0.669 and PSNR is 19.3 dB against the bf16 run, but neither measures degradation here. FP8 perturbs the model's prediction at each step, so the trajectory diverges and the result is a different video rather than a degraded copy of the same one.

What does transfer: mean luma is 106.4 against 106.2 and standard deviation is 61.2 against 61.1 across three sampled frames, so there is no exposure or contrast drift. Visual inspection of those frames shows no blocking, no banding, no texture collapse and no structural error. The chroma channels track closely, PSNR u 32.1 and v 34.8, while luma carries the divergence, which is where the difference in sampled content lives.

Callers who want tighter accuracy already have `granularity="channel"`, documented in `docs/inference/optimizations.md`.

## Tests

`fastvideo/tests/ops/quantization/test_fp8_ffn_suffix_wiring.py`, six CPU-only cases in a directory `unit_test.sh` already collects.

- both real H3 block prefixes get an `FP8QuantizeMethod`, `transformer_blocks.N.ff` and `token_refiner.refiner_blocks.N.ff`
- a real `MiniMaxH3TokenRefinerBlock` is constructed, so removing the `quant_config` or `.ff` propagation from the block fails even with the tuple intact
- WanVideo's `ffn.fc_in` and `ffn.fc_out` still match
- an unrelated projection still falls back to `UnquantizedLinearMethod`
- a root prefix of `tenant.ff.fc_in` does not drag `adaln_proj.linear` into FP8
- the conversion log reports the right layer counts, parameter totals and unmatched names, asserted on the `LogRecord` arguments rather than the rendered string, which formats both totals to `0.00B` at test scale

Reverting the two new entries turns three of the six red and leaves the three that should stay green.

On the GB10: 34 passed across `fastvideo/tests/ops/quantization/`, 5 passed in `fastvideo/tests/loader/test_fsdp_load_releases_checkpoint.py`, mypy clean, and an end-to-end H3 generation confirming the new log and unchanged component residency.

## Not in this patch

The underlying shape is a hardcoded allowlist where partial coverage is indistinguishable from full coverage. A deeper fix would have models declare their own quantizable layers, but that touches WanVideo, LTX-2 and Kandinsky5 and belongs in its own change. The reporting added here makes the current mechanism's misses visible in the meantime.

This module uses `logging.getLogger` rather than `init_logger`, matching all five quantization modules beside it. Switching it would make this file the odd one out, so that also belongs in its own change.


fastvideo/tests/ops/quantization/test_fp8_ffn_suffix_wiring.py
```diff
@@ -0,0 +1,169 @@
...
+    """Guard the wiring, not just the tuple.
+
+    ``MiniMaxH3FeedForward`` is reached through a block that must pass both
+    ``quant_config`` and a ``.ff`` prefix down. Dropping either would leave the
+    tuple correct and the model unquantized, so build the block itself.
+    """
+    from fastvideo.models.dits.minimax_h3 import MiniMaxH3TokenRefinerBlock
+    from fastvideo.platforms import AttentionBackendEnum
+
+    block = MiniMaxH3TokenRefinerBlock(
```
SolitaryThinker, line 97: **Medium:** This test is not CPU-only as written. At exact head `6fd11aa`, the suite gives `5 passed, 1 failed` on a CPU host: constructing `MiniMaxH3TokenRefinerBlock` reaches `get_attn_backend(...)` and raises `ValueError: Invalid attention backend for CPU` before either FP8 assertion runs. The exact `minimax_h3.py` and selector blobs match the versions used in the reproduction. Please mock the backend/platform selection or isolate the block-to-FF wiring so this test honors the file's CPU-only contract.

SolitaryThinker (changes requested):
The FP8 matching and coverage-reporting code looks sound in focused review: five of the six exact remote tests passed on this CPU host, a separate zero-coverage probe emitted the expected warning, and fastcheck/pre-commit are green. I am requesting changes for the sixth test, which is described as CPU-only but fails before its FP8 assertion on a CPU-only machine. Please also amend the commit to remove the `Co-Authored-By: Claude Opus 5` trailer; FastVideo maintainer policy does not allow AI agents as co-authors.

fastvideo/layers/quantization/fp8_config.py
```diff
@@ -43,6 +46,21 @@
...
+# prefix containing ".ff.fc_in" would make a substring rule match every
+# descendant, including ``adaln_proj.linear`` whose forward reads
+# ``self.linear.weight.dtype`` after the converter has removed that weight.
+# The tuple above keeps substring matching because it relies on it: "to_q"
+# is also how Kandinsky5's "to_query" is reached.
+_FP8_EXACT_SUFFIXES = (
+    ".ff.fc_in",
+    ".ff.fc_out",
+)
+
```
SolitaryThinker, line 63: **High:** These new H3 suffixes expose the existing post-load FP8 converter to sharded FSDP inference. `maybe_load_fsdp_model` calls `fully_shard` before `_maybe_quantize_model`; conversion then quantizes only `weight.to_local()`, registers that local shard as an ordinary `_fp8_weight` buffer, and removes the FSDP-managed parameter. `FP8QuantizeMethod.apply` reads the local buffer directly, so no FSDP all-gather occurs. I reproduced this on two CPU ranks: logical `(8,4)` / `(4,8)` DTensor weights became local `(4,4)` / `(2,8)` buffers, and both ranks failed the FF forward with `mat1 and mat2 shapes cannot be multiplied (2x4 and 8x2)`. `use_fsdp_inference=True` is a documented multi-GPU path, so please reject FP8 when the FSDP shard dimension is greater than one until supported, or keep the quantized representation under FSDP unshard/all-gather semantics. Add a multi-rank regression test before reporting these layers as successfully quantized.

PR #1784, opened 2026-08-29
[feat] Add streaming and GPU-accelerated LoRA extraction

shaoxiongduan:
## Purpose

Add a streaming, GPU-accelerated path to `scripts/lora_extraction/extract_lora.py` for large video transformers while preserving the existing exact CPU workflow and mixed LoRA/dense adapter format.

Previously, extraction materialized both transformer state dictionaries in host memory and performed full SVD on CPU. This is prohibitively expensive for models such as MiniMax-H3. The updated extractor streams indexed tensors, supports exact or randomized factorization on a selected device, and adds controls for accuracy, storage precision, resumability, and adapter validation.

## Changes

- Stream indexed safetensors one base/fine-tuned tensor pair at a time.
- Download only `transformer/*` when resolving Hugging Face repositories.
- Preserve the existing loading behavior through `--load-mode auto`:
  - try indexed loading first;
  - fall back to the legacy FastVideo pipeline loader;
  - allow either path to be selected explicitly.
- Add configurable CPU/GPU factorization:
  - `--device`
  - `--svd-method exact|randomized`
  - `--randomized-q`
  - `--oversample`
  - `--niter`
  - `--seed`
- Add storage dtype controls for:
  - low-rank factors;
  - exact `.diff` / `.diff_b` / `.diff_param` payloads;
  - fine-tuned-only `.set_weight` / `.set_param` parameters.
- Add repeatable `--exact-tensor-pattern` rules for matrices that should remain exact dense deltas instead of being rank-truncated.
- Document that matrix selection is runtime-agnostic and keep Wan condition embedders and its unwrapped output projection exact in the validated extraction command.
- Add resumable per-tensor extraction through `--work-dir` and `--resume`.
- Validate checkpoint key sets, tensor shapes, generated adapter keys, and output factor shapes.
- Preserve changed standalone parameters such as Wan2.2 `scale_shift_table` through `.diff_param`/`.set_param` payloads instead of silently dropping them.
- Write an adjacent extraction report containing settings, tensor counts, and reconstruction residuals.
- Omit unnecessary per-layer `lora_rank` and `lora_alpha` tensors when `alpha == rank`, which is already the loader default.
- Retain the existing mixed-adapter behavior:
  - changed matrices → `lora_B @ lora_A`;
  - changed base weights/biases → `.diff` / `.diff_b`;
  - changed standalone parameters such as `scale_shift_table` → `.diff_param`;
  - fine-tuned-only weights such as VSA gates → `.set_weight`;
  - other fine-tuned-only parameters → `.set_param`;
  - bit-identical parameters → omitted.
- Update the LoRA extraction documentation and add CPU/GPU unit coverage.

Exact CPU SVD remains the default for backward compatibility. GPU and randomized SVD are opt-in.

## Test Plan

```bash
source /mnt/lustre/vlm-s4duan/FastVideo/.venv/bin/activate

# CPU unit coverage
PYTHONPATH="$PWD" pytest \
  --confcutdir=fastvideo/tests/lora_extraction \
  fastvideo/tests/lora_extraction/test_streaming_lora_extraction.py -q

# Run the same test on a GPU node to activate CUDA coverage
PYTHONPATH="$PWD" pytest \
  --confcutdir=fastvideo/tests/lora_extraction \
  fastvideo/tests/lora_extraction/test_streaming_lora_extraction.py -q

# Existing dense LoRA loader coverage
PYTHONPATH="$PWD" pytest \
  fastvideo/tests/loader/test_lora_patch.py -q

# Wan2.2 extraction through the real FastVideo DMD LoRA loader on GPU
PYTHONPATH="$PWD" pytest \
  --confcutdir=fastvideo/tests/lora_extraction \
  fastvideo/tests/lora_extraction/test_lora_extraction.py \
  -q -s

# Full repository pre-commit suite
pre-commit run --all-files
```

A full MiniMax-H3 extraction was also run on one GB200:

```bash
python scripts/lora_extraction/extract_lora.py \
  --base /mnt/lustre/vlm-s4duan/models/MiniMax-H3 \
  --ft /mnt/lustre/vlm-s4duan/models/FastVideo-FastH3-8-step-Preview-v1-VSA-DataFree-full-local \
  --out /tmp/fasth3-generic-r64/adapter_model.safetensors \
  --rank 64 \
  --min-delta 0 \
  --load-mode indexed \
  --device cuda:0 \
  --svd-method randomized \
  --randomized-q 320 \
  --niter 4 \
  --seed 42 \
  --factor-dtype float16 \
  --dense-dtype float32 \
  --replacement-dtype source \
  --exact-tensor-pattern '^audio_proj_(in|out)\.weight$' \
  --exact-tensor-pattern '^context_embedder\.weight$' \
  --exact-tensor-pattern '^proj_(in|out)\.weight$' \
  --exact-tensor-pattern '^time_embedder\.'
```

The extracted adapter was then loaded through the MiniMax-H3 LoRA inference path and used to generate an eight-step FastH3 video.

## Test Results

<details>

```text
CPU extraction tests:
5 passed, 2 skipped

GPU extraction tests:
7 passed

Dense LoRA loader tests:
30 passed

Wan2.2 GPU extraction and real-loader integration test:
1 passed in 250.23s

Wan2.2 adapter application:
525 exact dense parameters fully applied
300 LoRA layers converted and applied
0 unmatched factor keys
WanDMDPipeline selected

Pre-commit (`pre-commit run --all-files`):
yapf passed
ruff passed
codespell passed
PyMarkdown passed
actionlint passed
mypy passed
filename check passed
suggestion check passed
```

Full MiniMax-H3 extraction:

```text
LoRA layers:             362
Exact additive tensors:   78
VSA replacement tensors:  50
Unchanged tensors:       198
Weighted residual:       0.8755630793425291
```

Comparison against the adapter previously produced by the specialized MiniMax-H3 extractor:

```json
{
  "old_keys": 852,
  "new_keys": 852,
  "key_sets_equal": true,
  "unequal_tensors": 0,
  "max_abs_difference": 0.0
}
```

End-to-end FastH3 inference:

```text
LoRA dense payload fully applied: 128 parameters
LoRA adapter applied to 362 layers
Video generation completed successfully
```

The existing Wan2.2 integration example now extracts on `cuda:0`, retains runtime-unsupported boundary matrices as exact deltas, constructs the real FastVideo `WanDMDPipeline`, and asserts that all 300 factorized matrices reach a layer.

</details>

## Checklist

Not applicable; this PR changes the extraction utility and does not modify model or pipeline numerics.


scripts/lora_extraction/merge_lora.py
```diff
@@ -214,7 +273,17 @@ def merge_lora_into_base(base_sd: dict, adapter: dict) -> dict:
         merged_sd[weight_key] = merged_weight.to(base_sd[weight_key].dtype)
         merged_count += 1
 
-    LOG.info(f"Merged {merged_count} layers, skipped {skipped_count}")
+    additive, replacement, unrecognized = group_dense_keys(adapter)
+    dense_merged, dense_skipped = merge_dense_into_base(merged_sd, additive, replacement)
```
SolitaryThinker: **High:** Indexed extraction emits Hugging Face parameter names, but both the factor loop above and this dense merge look them up directly in the FastVideo custom state dict without applying the model mappings. For Wan, `blocks.0.attn1.to_q.weight` maps to `blocks.0.to_q.weight`; a focused extract-shaped probe merged **0 LoRA layers and 0 dense tensors**. Please apply `lora_param_names_mapping` (where applicable) followed by `param_names_mapping`, or perform the merge wholly in the HF namespace. Add an indexed-extract -> merge integration test using real Wan mappings; the current test uses custom names on both sides and cannot catch this.

scripts/lora_extraction/extract_lora.py
```diff
@@ -130,292 +258,439 @@ def load_transformer_state_dict_from_model(
...
+            return
 
-    # cleanup
-    try:
-        del pipeline, transformer
-    except Exception:
-        pass
-    torch.cuda.empty_cache()
-    return state_dict_cpu
+    base_state = load_transformer_state_dict_from_model(base)
```
SolitaryThinker, line 377: **High:** `base_revision` and `finetuned_revision` are dropped in explicit pipeline mode and in the automatic fallback, while the output metadata later records those requested revisions. This can extract default-branch weights while claiming pinned commits. Resolve each revision to a pinned snapshot before pipeline loading, thread the revision through the loader, or reject revision flags for a mode that cannot honor them. The immutable revision also needs to participate in resume identity.

scripts/lora_extraction/extract_lora.py
```diff
@@ -130,292 +258,439 @@ def load_transformer_state_dict_from_model(
...
+    device = torch.device(config.device)
+    if device.type == "cuda" and not torch.cuda.is_available():
+        raise RuntimeError(f"CUDA device requested but CUDA is unavailable: {device}")
+    factor_dtype = _torch_dtype(config.factor_dtype)
+    exact_patterns = _compile_patterns(config.exact_tensor_patterns)
+
+    for index, key in enumerate(tqdm(sorted(finetuned.keys), desc="extracting LoRA", unit="tensor")):
+        tensor_file = work_dir / "tensors" / f"{index:05d}.safetensors"
+        existing = manifest["layers"].get(key)
+        if existing is not None and (existing.get("tensor_file") is None or tensor_file.is_file()):
```
SolitaryThinker, line 576: **High:** Resume decides this shard is reusable from source/config strings plus file presence; it does not validate checkpoint identity. I extracted once, overwrote the fine-tuned safetensors in place with the same keys/shapes, then resumed: the adapter retained the old `[[1,1],[1,1]]` delta while a fresh extraction produced `[[2,2],[2,2]]`. A partial run can therefore mix old and new weights silently. Persist and validate immutable revisions or checkpoint/shard fingerprints before skipping completed layers.

scripts/lora_extraction/extract_lora.py
```diff
@@ -130,292 +258,439 @@ def load_transformer_state_dict_from_model(
...
+                "delta_frobenius_norm": delta_fro_sq**0.5,
+                "relative_residual": residual,
+                "max_abs_delta": max_abs_delta,
+            }
+            del lora_a, lora_b, singular_values, payload
+        else:
+            output_key = dense_payload_key(key)
+            if config.dense_payload:
+                output_dtype = _resolve_output_dtype(config.dense_dtype, finetuned_tensor.dtype)
+                _save_layer_payload(tensor_file, {output_key: delta.to(output_dtype)}, key)
```
SolitaryThinker, line 664: **Medium:** The module describes these as exact dense deltas, but the default `dense_dtype='source'` rounds the FP32 difference back to FP16/BF16 and does not always reconstruct the fine-tuned tensor. One FP16 probe reconstructed `-0.0260009765625` instead of `-0.026031494140625`. Default exact deltas to FP32, or round-trip-check and promote/store a replacement whenever source-dtype addition cannot reproduce the target. Please cover FP16/BF16, not only the current float32 toy state.

fastvideo/models/loader/lora_patch.py
```diff
@@ -49,11 +50,11 @@
...
 
-# Suffix -> the parameter suffix it targets. ``.diff``/``.diff_b`` are additive,
-# ``.set_weight`` replaces. Ordered longest-first so ``.diff_b`` is tested before
-# ``.diff`` would match a truncated key.
-ADDITIVE_SUFFIXES: dict[str, str] = {".diff_b": ".bias", ".diff": ".weight"}
-REPLACEMENT_SUFFIXES: dict[str, str] = {".set_weight": ".weight"}
+# Suffix -> the parameter suffix it targets. An empty target suffix preserves the
+# full parameter name for standalone nn.Parameters. Ordered longest-first so the
+# generic spellings are tested before the shorter weight/bias spellings.
+ADDITIVE_SUFFIXES: dict[str, str] = {".diff_param": "", ".diff_b": ".bias", ".diff": ".weight"}
```
SolitaryThinker, line 56: **Medium:** These new dense suffixes are later resolved with only `param_names_mapping`, unlike the low-rank loader, which first applies `lora_param_names_mapping`. For Wan, official-style `blocks.0.self_attn.q.weight` remains unchanged under the regular mapping but the two-step mapping correctly reaches `blocks.0.to_q.weight`; factors can therefore apply while their matching dense payload stays unmatched. Please compose the same two mappings for dense keys and add an official-style mixed-adapter test.

SolitaryThinker (changes requested):
Requesting changes for several silent correctness failures in the documented extraction and merge paths. I reproduced the namespace mismatch, stale-resume behavior, ignored revisions, and low-precision dense round-trip locally. The focused CPU tests and pre-commit pass, but the current tests do not exercise these production cases; Buildkite fastcheck is also still red across all microscope lanes.

scripts/lora_extraction/merge_lora.py
shaoxiongduan: Fixed in 3906bd84. Offline merging now applies `lora_param_names_mapping` followed by `param_names_mapping` to factor and dense payloads, handles mapped fused slices, and fails on unapplied payloads by default (`--allow-unmatched` is the explicit opt-out). I added an indexed extraction -> offline merge regression using the real Wan mappings (`blocks.0.attn1.to_q` -> `blocks.0.to_q` and `condition_embedder.time_proj` -> the FastVideo time-modulation parameter), plus fused-source accumulation coverage.

scripts/lora_extraction/extract_lora.py
shaoxiongduan: Fixed in 3906bd84. Pipeline mode now rejects revision flags, local paths reject Hugging Face revisions, and auto mode re-raises an indexed-loading failure instead of falling back whenever either revision was requested. Indexed HF resolution still passes the revisions to `snapshot_download`. Tests cover explicit pipeline rejection, local-path rejection, and preventing revision-dropping auto fallback; checkpoint fingerprints also participate in the resume configuration.

scripts/lora_extraction/extract_lora.py
shaoxiongduan: Fixed in 3906bd84. Indexed readers now SHA-256 fingerprint the index plus all referenced transformer shards, persist both base/fine-tuned fingerprints in the resume configuration, and reject a manifest when either checkpoint changed. File stats are checked around fingerprinting and again after extraction to catch in-flight mutation. The regression overwrites a fine-tuned shard in place with the same keys/shapes and verifies resume is rejected.

scripts/lora_extraction/extract_lora.py
shaoxiongduan: Fixed in 3906bd84. `dense_dtype` now defaults to FP32 in both the API and CLI, while lower precision remains an explicit choice. Parameterized FP16 and BF16 regressions use values for which source-dtype delta storage would fail and assert that the default FP32 payload reconstructs the exact fine-tuned source value after casting.

fastvideo/models/loader/lora_patch.py
shaoxiongduan: Fixed in 3906bd84. `maybe_load_fsdp_model` now supplies both mapping callables to `DenseLoRAPatch`; dense names are resolved through the LoRA-specific mapping first and the checkpoint/model mapping second, matching factor loading. The loader regression uses official-style Wan `blocks.0.self_attn.q.diff` and verifies it reaches `blocks.0.to_q.weight`. The full Wan extraction/runtime lane also passes with all 525 dense parameters and 300 factorized layers applied.

PR #1785, opened 2026-08-29
[misc]: surface GenerationResult.peak_memory_mb in the MiniMax H3 examples

KyleNeverGivesUp:
## Problem

Every generation already computes its own peak. The worker calls `torch.cuda.max_memory_allocated()` at `fastvideo/worker/multiproc_executor.py:699` and the value reaches the caller as `GenerationResult.peak_memory_mb`. Both MiniMax H3 examples receive it and print only the generation time.

On a memory-constrained device that number is the one that decides whether a configuration runs at all. It is also what a reader has to reproduce in order to check any memory claim about this model, and right now they cannot: the field is populated, handed over, and dropped on the floor.

`examples/inference/optimizations/spark_benchmark.py` already prints it, so this only brings the H3 examples in line with an existing convention.

## What this changes

Two lines in each example, guarded on the value being present so nothing changes for a backend that does not report one.

```
Generation time: 902.31s
Peak memory: 70880.0 MB
```

`basic_minimax_h3_t2v.py` prints it for the first request and for each `--repeats` iteration, next to the generation-time line those iterations already emit. `basic_fasth3.py` prints it inside the same per-run block that already reports wall time, generation time and denoise time.

## Why it matters right now

Two open PRs report peak memory on one GB10 and neither is reproducible without a local patch to these files:

- #1761 reports component residency and the ceiling between 222,208 and 347,200 packed tokens.
- #1780 reports 44.23 GiB of resident weights falling to 32.99 GiB once FP8 reaches the feed-forward stack.

Every one of those numbers was read from a three-line local edit that has never been committed. With this change a reader can run the same command and see the same line.

## Testing

No test parses the output of either example, so nothing is invalidated. `grep -rn "Generation time:" fastvideo/tests/` returns nothing.

Checked on one GB10, 768x1344, 124 frames, MiniMax H3 r16:

```
Generation time: 902.31s
Peak memory: 70880.0 MB
```

which matches `torch.cuda.max_memory_allocated()` as reported by the worker for the same run.


SolitaryThinker:
Pushed `de7f77083`: documented what the surfaced `peak_memory_mb` actually measures, corrected the doc that called it per-run, and added a test guard. The branch was 39 commits behind `main` (which had touched both example files), so it was rebased onto `origin/main` first — clean, no conflicts, hence the force-push.

**Fixed**
- `Peak memory` is the worker's allocator *high-water mark*: `torch.cuda.max_memory_allocated()` is never reset between generations (`fastvideo/worker/multiproc_executor.py:712`), so the per-repeat lines are not per-run peaks -> scope documented next to both print sites (`examples/inference/basic/basic_fasth3.py:386`, `examples/inference/basic/basic_minimax_h3_t2v.py:142`, `:150`) and in `examples/inference/basic/README.md:121`.
- `docs/getting_started/installation/spark_performance.md:82` called `peak_memory_mb` a "per-run figure" -> reworded to "worker-measured figure" and spelled out that it includes model load and every prior run.
- The new print branch had no test coverage -> `fastvideo/tests/inference/test_basic_fasth3_profile.py` now returns `peak_memory_mb=42.0` from the fake generator and asserts the line is printed once per measured run (`:219`, `:260`).

**Verified**
- `python3 -m py_compile` on the three changed `.py` files: OK; 120-column check on them: clean.
- Static trace of writer -> reader: the field is populated in the mp executor, survives `from_legacy_result`, and the `is not None` guard is load-bearing — the Ray backend never populates the key, so that path stays quiet.
- CI re-triggered on the new head; the previous head was green on `fastcheck-passed`, `pre-commit`, and `trigger` (only the Mergify merge-protection gate was red, waiting on review + full suite).

**Left open**
- The value is still the worker process's lifetime peak (includes model load) and rank 0's only (`responses[0]`). Making it a true per-generation peak means resetting `torch.cuda.reset_peak_memory_stats()` before each `execute_forward`, which changes the metric for `bench_serving`, the OpenAI `X-Peak-Memory-MB` header, and the performance tests — a design decision beyond this PR, so the printed label was left as-is (it matches the field name and the output documented in the PR body).
- `basic_minimax_h3_t2v.py` still duplicates the guarded metric prints between the first generation and the `--repeats` loop; a shared helper would dedupe it, but the duplication predates this PR.
- PR body / commit message cite `multiproc_executor.py:699`; the call is at `:712` in both the PR head and current `main`.

PR #1803, opened 2026-09-01
[feat]: run FastH3 across two DGX Sparks with Ray sequence parallel

aryan5v:
## Summary

Rebased onto current `main` so this PR **does not fight #1793**. Sequential MiniMax H3 start (`h3_sequential_load`, GPU-direct DiT load) stays as merged. This branch adds the rest of the GB10 / dual-Spark work on top of that.

### From #1761 (authors preserved)

Kyle's `lazy_module_load` / `LazyModule` path: load each opted-in component on first use and free it after its last stage, with in-process reload for a later `generate()`. MiniMax-H3 opts in. Satyam's follow-ups: compile across lazy reloads, LoRA setup must not materialize lazy DiTs, share compiled VSA graphs across H3 layers. Geometry is read from checkpoint `config.json` so input prep / unpatchify do not hold a live DiT through VAE decode.

GPU-direct DiT load is **not** re-landed here; it already merged in #1793.

### Dual-Spark FastH3 (this PR's original work)

- Ray executor: driver-local `set_log_queue` / `clear_log_queue` (Queue is not picklable across nodes) and copy `NCCL_*` onto workers.
- `basic_fasth3.py --execution-backend {mp,ray}` (auto-ray when `RAY_ADDRESS` is set).
- QSFP / Ray bring-up, `spark_pair_env.sh`, generate YAML, cookbook recipe.

Sequence parallel **replicates** the DiT (~66 GiB/node), so each Spark still needs `h3_sequential_load` and `lazy_module_load` (both auto on unified memory).

Measured on two GB10s over QSFP RoCE (~21 GB/s NCCL busbw), same alpine 768×1344 FastH3 recipe, Triton VSA, no FA4, sequential + lazy load, parallel VAE, cold process:

| Run | E2E | Denoise (incl. DiT load) | Decode |
|---|---:|---:|---:|
| 1 GPU, 124 frames | 374–393 s | 180–188 s | 151–156 s |
| **SP=2, 124 frames** | **292 s** | **122 s** | **102 s** |
| **SP=2, 345 frames (~14.4 s)** | **587 s** | **351 s** | **173 s** |

That 292 s is faster than the ~330 s one-Spark clip from earlier bring-up. It is not a lower bound: first decode still pays VAE `torch.compile`; TAEH3 (#1795) is a separate opt-in and was not used.

## Test plan

- [ ] `pytest fastvideo/tests/worker/test_ray_distributed_executor.py`
- [ ] `pytest fastvideo/tests/stages/test_minimax_h3_sequential_start.py fastvideo/tests/stages/test_lazy_module_load.py`
- [ ] Two Sparks: `source examples/inference/optimizations/spark_pair_env.sh`, `ray start` on QSFP IPs with `FASTVIDEO_HOST_IP` matching `--node-ip-address`, then `python examples/inference/basic/basic_fasth3.py --num-gpus 2 --execution-backend ray --vsa-kernel triton --no-fa4 ...`

### DIFF
--- a/docs/assets/cookbook-recipes.json
+++ b/docs/assets/cookbook-recipes.json
@@ -666,7 +666,7 @@
       "evidence": "Verified",
       "expected_artifact": "MP4 under outputs/fasth3_spark_pair/",
       "modes": ["T2VA", "2-Spark SP"],
-      "limitations": ["Requires a two-node Ray cluster on the QSFP interconnect. See docs/getting_started/installation/spark_pair.md."]
+      "limitations": ["Requires a two-node Ray cluster on the QSFP interconnect. Height, width, frames, and steps in the YAML are examples. Edit them or pass CLI flags. See docs/getting_started/installation/spark_pair.md."]
     },
     {
       "id": "matrix-game-2",
--- a/docs/assets/cookbook-recipes.json
+++ b/docs/assets/cookbook-recipes.json
@@ -646,6 +646,14 @@
       "limitations": ["The upstream checkpoint must be converted to Diffusers layout via scripts/checkpoint_conversion/convert_mmaudio_to_diffusers.py unless loaded from the FastVideo converted repo as done here."]
     },
     {
+      "id": "fasth3-spark-pair",
+      "task": "Text to video",
+      "label": "FastH3 on two DGX Sparks (sequence parallel)",
+      "model": "FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-DataFree",
+      "source": "examples/inference/basic/basic_fasth3_spark_pair.yaml",
+      "command": "source examples/inference/optimizations/spark_pair_env.sh && FASTVIDEO_VSA_SM100A=0 FASTVIDEO_FA4=0 FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN_H3 FASTVIDEO_VAE_PARALLEL_DECODE=1 fastvideo generate --config examples/inference/basic/basic_fasth3_spark_pair.yaml"
+    },
+    {
       "id": "matrix-game-2",
       "family": "matrixgame",
       "stage": "inference",
--- a/docs/assets/cookbook-recipes.json
+++ b/docs/assets/cookbook-recipes.json
@@ -647,11 +647,26 @@
     },
     {
       "id": "fasth3-spark-pair",
-      "task": "Text to video",
+      "family": "minimax_h3",
+      "stage": "inference",
+      "task": "Few-step text to video (with audio)",
       "label": "FastH3 on two DGX Sparks (sequence parallel)",
+      "summary": "Run one FastH3 clip across two GB10s with Ray sequence parallel over QSFP RoCE. Sequential load and lazy module load stay on because SP replicates the DiT on each node.",
       "model": "FastVideo/FastVideo-FastH3-4-step-Preview-v1-VSA-DataFree",
       "source": "examples/inference/basic/basic_fasth3_spark_pair.yaml",
-      "command": "source examples/inference/optimizations/spark_pair_env.sh && FASTVIDEO_VSA_SM100A=0 FASTVIDEO_FA4=0 FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN_H3 FASTVIDEO_VAE_PARALLEL_DECODE=1 fastvideo generate --config examples/inference/basic/basic_fasth3_spark_pair.yaml"
+      "command": "source examples/inference/optimizations/spark_pair_env.sh && FASTVIDEO_VSA_SM100A=0 FASTVIDEO_FA4=0 FASTVIDEO_ATTENTION_BACKEND=VIDEO_SPARSE_ATTN_H3 FASTVIDEO_VAE_PARALLEL_DECODE=1 fastvideo generate --config examples/inference/basic/basic_fasth3_spark_pair.yaml",
+      "gpu_types": ["NVIDIA"],
+      "hardware": {
+        "platform": "cuda",
+        "gpu_count": 2,
+        "accelerator": "NVIDIA GB10 (DGX Spark pair)",
+        "evidence": "validated",
+        "evidence_url": "https://github.com/hao-ai-lab/FastVideo/pull/1803"
+      },
+      "evidence": "Verified",
+      "expected_artifact": "MP4 under outputs/fasth3_spark_pair/",
+      "modes": ["T2VA", "2-Spark SP"],
+      "limitations": ["Requires a two-node Ray cluster on the QSFP interconnect. See docs/getting_started/installation/spark_pair.md."]
     },
     {
       "id": "matrix-game-2",
--- a/docs/cookbook/index.md
+++ b/docs/cookbook/index.md
@@ -461,7 +461,10 @@ hide:
       Inference is the first complete stage. Distillation, fine-tuning,
       training, evaluation, optimization, and deployment will reuse the same
       family-first structure as their recipes land. Each family page shows
-      which stages are available and which are planned.
+      which stages are available and which are planned. Two DGX Sparks: bring up
+      the QSFP Ray cluster first
+      (<a href="../getting_started/installation/spark_pair/">pair two Sparks</a>),
+      then pick the FastH3 two-Spark recipe.
     </p>
   </section>
 </div>
--- a/docs/design/inference_schema_parity_inventory.yaml
+++ b/docs/design/inference_schema_parity_inventory.yaml
@@ -30,6 +30,7 @@ surfaces:
       image_encoder_cpu_offload: generator.engine.offload.image_encoder
       vae_cpu_offload: generator.engine.offload.vae
       pin_cpu_memory: generator.engine.offload.pin_cpu_memory
+      lazy_module_load: generator.engine.offload.lazy_module_load
       enable_torch_compile: generator.engine.compile.enabled
       enable_torch_compile_text_encoder: generator.engine.compile.text_encoder_enabled
       enable_torch_compile_vae: generator.engine.compile.vae_enabled
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -4,8 +4,9 @@
 FastVideo supports the following hardware platforms:
 
 - [NVIDIA CUDA](installation/gpu.md)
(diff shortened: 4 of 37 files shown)


KyleNeverGivesUp:
#1761 has now been independently rebased onto main and is mergeable with no conflicts. #1803 currently includes the #1761 commit stack because it was built on top of that work. Please merge #1761 first, then rebase #1803 onto the updated main so that #1803 contains only its follow-up changes.

This would leave #1803 focused on the geometry fix, the dual-Spark Ray work, and the accompanying recipe documentation, which should be reviewed separately.

SolitaryThinker:
## Review: lazy module load + dual-Spark Ray path

Recall-oriented review of the full diff against `main` (merge-base `a28f2bab`). Each item below was verified against the PR-head tree; several were reproduced empirically (noted inline). Ordered most severe first.

### Correctness

**1. `fastvideo/pipelines/basic/minimax_h3/minimax_h3_pipeline.py:74` — the PR's own test fails at head.**
`_apply_h3_checkpoint_arch_configs` unconditionally reads `pipeline_config.dit_config.patch_size` (and `vae_config.arch_config.*`) for its log line, so `post_init` raises `AttributeError: 'DiTConfig' object has no attribute 'patch_size'` whenever `FastVideoArgs.pipeline_config` is the generic default. `test_minimax_h3_sequential_start.py::test_inference_defers_dit_and_vae_until_after_conditioning` — which this PR extends and lists in its test plan — passes 6/6 at base and fails 1/6 at PR head.

**2. `fastvideo/pipelines/lora_pipeline.py:281` — LoRA + `lazy_module_load` silently drops the adapter and never frees the DiT.**
`convert_to_lora_layers` materializes the DiT proxy and pins the real blocks in `self.lora_layers` / `LoRAModelLayers.block_mapping`, so `LazyModule.release()` frees nothing; the next `materialize()` re-runs the plain loader and produces a base-weight DiT while `lora_initialized` stays True, and `set_lora_adapter` / `merge_lora_weights` edit orphaned layers. Triggered by `basic_fasth3_lora_preview.py --num-gpus 1` on a discrete GPU (it inherits the `num_gpus == 1 → lazy on` override). Reproduced: a block parameter survives `release()`.

**3. `fastvideo/worker/ray_distributed_executor.py:80` — copying `NCCL_SOCKET_IFNAME` / `NCCL_IB_HCA` / `GLOO_SOCKET_IFNAME` from the driver breaks pairs with asymmetric NIC names.**
The push goes through `update_environment_variables`, which overwrites (warning only) the per-node value the user set before `ray start` — exactly what `spark_pair_env.sh` and `spark_pair.md` tell them to do when `ibdev2netdev` differs per box. Remote worker then binds a non-existent interface: NCCL "no socket interface found" / Gloo "Unable to find address for". Reproduced that per-node exports do reach Ray actors and only this push clobbers them; `GLOO_SOCKET_IFNAME` is exact-match so no single driver value covers both boxes. Base copied no `NCCL_*`/`GLOO_*` vars. Suggest `setdefault` semantics for the interface-binding trio, or drop them from the list.

**4. `minimax_h3_pipeline.py:227` — text encoder re-read from disk on every request on discrete GPUs.**
`forward` now always routes through `_run_condition_then_denoise`, which unconditionally calls `_release_text_encoder()`; the base fast path `if self._denoise_stages_ready: return super().forward(...)` is gone. With both `h3_sequential_load` and `lazy_module_load` off (the `--num-gpus 4` default), request 2's `_ensure_text_encoder` finds the encoder missing and re-reads Qwen3-VL per rank. `text_encoder_cpu_offload` parking is thrown away, contradicting the `fastvideo_args.py:164-165` contract. Reproduced: PR head reloads every request, base 0.

**5. `minimax_h3_pipeline.py:187` — `compile.vae_enabled: true` in the shipped Spark YAML is a no-op; text-encoder compile applies to request 1 only.**
Pipeline-level compile (`_maybe_compile_pipeline_module` / `set_materialize_transform`) runs only in `post_init` on modules present then. On the sequential path `vae`/`audio_vae`/`transformer` are absent, so the call returns at `if module_name not in self.modules` — while post_init still logs "Torch Compile enabled for VAE". Proxies later added by `_load_denoise_modules` and the new `_ensure_text_encoder` (every path) get no transform. Reproduced: `vae` transform=None, `prepare_for_compile` calls: 0. `offloading.md:148` ("compile setup is reapplied after each reload") and `spark_pair.md:181` ("later generate() calls are cheaper") don't hold for this model.

**6. `fastvideo/pipelines/composed_pipeline_base.py:438` — release schedule pins proxy identity; `add_module` never rebuilds it.**
`_release_text_encoder` pops the original proxy and `_ensure_text_encoder` installs a new one via `add_module`, so from request 2 the conditioning stage's `_lazy_modules_to_release` is a stale released proxy (non-sequential) or `()` (sequential). The reloaded ~66 GiB encoder is freed only by refcount inside `_release_text_encoder`; if conditioning raises on request 2, nothing releases it (see 7). "Released deferred module ... freed X GiB" accounting silently stops covering the encoder.

**7. `minimax_h3_pipeline.py:306` — the new abort-path cleanup is dead for the only pipeline that opts in.**
`ComposedPipelineBase.forward`'s `except BaseException: self._release_all_lazy_modules()` never runs for H3 because `MiniMaxH3BasePipeline.forward` overrides `forward` without `super()` or an equivalent wrapper. A denoise failure on GB10 leaves `audio_vae` (T2VA) or `vae`+`audio_vae` (FL2VA/Ref2VA) materialized — the retry the docstring says this protects starts with less headroom. `_release_all_lazy_modules` has no other caller.

**8. `examples/inference/basic/basic_fasth3.py:285` (and `basic_minimax_h3_t2v.py:92`) — examples force lazy on for any single GPU regardless of device.**
`True if args.lazy_module_load is None and args.num_gpus == 1 else ...` bypasses the tri-state policy this PR adds in `finalize_device_offload_policy`. On a single discrete GB200/B200 every `generate()` now releases and re-reads the DiT (~35 s by the PR's numbers), both VAEs, and the encoder for no memory benefit; `--warmup --repeats N` measures reloads. The two help strings already disagree with `add_cli_args`.

**9. `fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_latent_preparation.py:320` — audio VAE materialized two stages early and held through denoise.**
`patch_size` moved off the live DiT, but `forward` still reads `self.audio_vae.latent_channels` (320/323/334, T2VA path) and `self.vae.latent_channels` (196/244). Under lazy load the audio VAE's last user is `audio_decoding_stage`, so denoise peak = DiT + audio VAE (+ video VAE for FL2VA/Ref2VA), not the "largest single component" `offloading.md` promises. Same fix pattern as `h3_dit_patch_size`: read from `pipeline_config.*_config.arch_config.latent_channels`.

**10. `composed_pipeline_base.py:705` — four WARNINGs + schedule rebuilds on the first request of the default GB10 / shipped-YAML path.**
Post_init installs hooks with only the two condition stages; `_add_denoise_stages` then hits `add_stage`'s "added after the deferred-release schedule was built; rebuilding" WARNING four times per worker (8 across the pair). Reproduced: 4 on request 1, 0 on request 2, 0 with either flag off. The designed happy path is logged as an anomaly.

**11. `composed_pipeline_base.py:667` — WARNING on every non-H3 model on unified memory.**
`lazy_module_load=None` auto-resolves True on GB10/Jetson/MPS for every pipeline, but only H3 declares `_lazy_module_names`, so Wan/Hunyuan/Cosmos/LTX/... all log "lazy_module_load is on but no deferred module is held by a stage ... may load its modules eagerly" on the documented opt-out state. Skip hook install (or INFO) when `_lazy_module_names` is empty.

**12. `fastvideo/pipelines/stages/base.py:165` — `BaseException` in output verification skips release.**
`_execute` is wrapped in `except BaseException` but output verification in `except Exception`, so a KeyboardInterrupt/SystemExit inside `verify_output` never releases the stage's deferred modules — and H3 has no outer net (7). A single `try/finally` closes it. (`torch.cuda.OutOfMemoryError` is an `Exception`, so OOM is covered.)

**13. `ray_distributed_executor.py:42` (and `:242`) — the Gloo-loopback change is a no-op with a misleading test.**
Base `node_gpus` was keyed by Ray node id, so two one-GPU Sparks already gave `len(node_gpus) == 2`; with the pre-existing `n_nodes != n_ips` guard, `len(set(worker_ips)) <= 1` is equivalent in every configuration that passes. The docstring, commit message, and `test_gloo_loopback_follows_worker_ips_not_node_count` describe a case that never occurred, so whatever actually failed (likely 3) is misattributed. The rewritten hunk also keeps a literal `"127.0.0.1"` instead of `get_loopback_ip()` (used by `MultiprocExecutor`), so `FASTVIDEO_LOOPBACK_IP` is ignored on Ray.

**14. `minimax_h3_pipeline.py:250` — audio VAE sampling rate hardcoded on the Ref2VA input path.**
`_input_audio_vae(ref2va=True)` now always returns `_default_audio_geometry()` (32000); neither the overlay nor this helper reads `audio_vae/config.json` or `pipeline_config.audio_vae_config`. A checkpoint declaring a different rate would resample reference audio to the wrong rate before `audio_vae.encode` uses the real one. All three published checkpoints declare 32000, so nothing shipped is affected today; base only fell back on the deferred Spark path.

### Docs

**15. `docs/inference/offloading.md:28`** still says "Sequential load currently cannot re-encode a new prompt on that worker; start a new generator", which this PR's `_ensure_text_encoder` makes false and which `spark_pair.md:115/195` and the extended sequential-start test directly contradict.

### Below the cut (verified, lower impact)
`_regional_compile_layer_idx` is now a dead per-block tensor and `prepare_for_compile` runs twice on the regional path (`video_sparse_attn_h3.py:500`); generic/training compile now does VSA pooling/top-k on dense layers before OR-ing an all-True mask (`:624`); the enumerated `ADDITIONAL_ENV_VARS` list silently drops any unlisted NCCL knob; the activation-trace WARNING at `composed_pipeline_base.py:361` fires without checking `FASTVIDEO_TRACE_ACTIVATIONS`; `_apply_h3_checkpoint_arch_configs` duplicates the loader overlay (no drift today); `set_log_queue` is a third copy of the base-class body; `h3_dit_patch_size` re-validates per keyframe; `exclude_lora_layers` is readable from `pipeline_config` without materializing the DiT.

One candidate was refuted: `_compile_layer_idx` landing on CPU under `dit_cpu_offload` — the non-FSDP inference loader materializes on the local CUDA device regardless of that flag.

https://claude.ai/code/session_01G6j6Dce9rizzdLuecd4sxY

PR #1826, opened 2026-09-07
[feat] Add phase-aware modular training performance metrics

Jack47:
## Summary

- add per-step optimizer/sample throughput and per-role Transformer call accounting to the modular trainer
- estimate Wan Transformer-core TFLOP/s and MFU using actual BDir/VSA, block-causal AR, streaming KV-cache, and DMD multi-role/multi-rollout geometry
- expose an optional peak BF16 override, document metric semantics and limitations, and keep all performance hooks configurable

## Why

The existing standalone MFU helper assumes one dense Wan forward. That does not represent causal chunking, sparse attention, DMD teacher/critic calls, or repeated student rollouts. Counting each role at the Transformer boundary makes the metric follow the algorithm actually executed.

## Validation

- 40 targeted CPU tests passed, including existing Trainer managed-optimization coverage
- pre-commit passed: YAPF, Ruff, codespell, PyMarkdown, mypy, filename and suggestion checks

GPU validation was not run locally; FLOP geometry is covered with CPU unit tests.


SolitaryThinker:
Pushed `8726d1526` (rebased onto `main` @ `37d06a832`): bound pre-loop performance records, harden the metric hooks, and document the remaining metric caveats.

**Fixed**
- Pre-loop Transformer forwards (KD teacher cache generation, step-0 validation) accumulated one `_ForwardWork` per call until the first step's `reset()` — ~6.5 GB host RSS for the shipped 600k-sample KD config. The monitor is now reset after the pre-loop phase and before the step loop (`fastvideo/train/trainer.py:156`).
- An estimator exception inside the forward hook could abort training. Estimation is now isolated so metrics can never kill a run (`fastvideo/train/utils/performance.py:293`).
- Peak-TFLOPs inference ran even with `training.performance.enabled: false`; it now only runs when metrics are enabled (`fastvideo/train/trainer.py:123`).
- `_teacher_forcing_frame_pairs` duplicated `_blockwise_causal_frame_pairs`; it now composes it (`fastvideo/train/utils/performance.py:98`).
- `_vsa_attention_pairs` re-implemented `compute_topk` and had a stale docstring; it now imports the shared helper from the VSA backend (`fastvideo/train/utils/performance.py:112`).
- Removed the dead `_ForwardWork.backward_expected` field.
- Docs: added `perf/peak_tflops_per_gpu` and `perf/role/<role>/causal_chunks`, plus caveats for managed-optimization sample throughput, per-role MFU aggregation, `tp_size`, and the MatrixGame2 window (`docs/training/train_infra.md`).
- Added `fastvideo/tests/train/trainer/test_performance_wiring.py`: asserts pre-loop forwards are dropped before the first step and that `enabled: false` logs no `perf/*` keys.

**Verified**
- `pytest fastvideo/tests/train/` — 185 passed (CPU venv + `fastvideo_kernel` stub).
- `pytest fastvideo/tests/contract/test_ci_test_collection.py fastvideo/tests/contract/test_merge_ci_plan.py` — 50 passed.
- The new pre-loop test fails when the reset is reverted, passes with it.
- `py_compile` on changed files; no >120-column Python lines.

**Left open**
- `perf/samples_per_sec` under-reports for `DiffusionNFT`: its outer step consumes `num_batches_per_epoch` dataloader batches, but the metric divides one configured batch by `step_time`. A proper fix needs a method-reported sample count, so the docs now point at `nft/num_sampled`; the FLOP metrics still aggregate every forward.
- `tp_size > 1` is not modeled in the data-parallel scaling (all shipped training configs use `tp_size: 1`); documented as a limitation.
- MatrixGame2's 15-frame compatibility window is not modeled, so its `perf/attention_density` is an upper bound; documented.
- `_output_requires_grad` does not handle dataclass outputs, so Kandinsky5 `backward_forwards` reports 0. Changing it changes logged values, so it is left for the author.
- CI: the previous head's `microscope-unit-tests` failure (exit status 2 after 86 min) matches the same lane failing on several unrelated PRs in the same window (builds #1155/#1164/#1165/#1171/#1176), and the new tests pass locally. Fastcheck is re-running on the new head.

PR #1828, opened 2026-09-07
[perf] MiniMax H3: return uint8 frames from the decode worker

stayinalive181:
## Purpose

The MiniMax-H3 decode stage returns a CPU fp32 pixel buffer that crosses the multiprocess-executor boundary (4.27 GB for a 345-frame 1344x768 clip), and the main process casts it to uint8 immediately afterwards. Casting inside the worker ships 1.07 GB instead. Lossless: the clamp matches the post-decode path, so the pixels are unchanged.

This is one of the runtime patches published in [hlander-ai/minimax-h3](https://github.com/hlander-ai/minimax-h3) (Apache-2.0), where the post-decode/encode/save stages went from 3.09 s to 1.98 s per 345-frame clip on 8xH200 (together with a worker-side encode change that is not part of this PR). It applies to every hardware target, not only Hopper.

## Changes

- `fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py`: on the output rank, quantize the decoded fp32 buffer to uint8 (`mul_(255).clamp_(0, 255)`) before handing it to the executor. The TAEH3 preview path and `output_type="latent"` are untouched.
- `fastvideo/utils.py`: new `pixels_to_uint8(pixels)` — uint8 passes through untouched, float pixels are clamped to `[0, 1]` and cast. Every consumer of decoded frames goes through it, so a pipeline returning either dtype produces the same bytes.
- `fastvideo/train/callbacks/validation.py` (review follow-up): the `ValidationCallback` frame loop reads `output_batch.output` directly and did `(x * 255)`, which wraps uint8 modulo 256 (`[63, 127, 191, 255]` became `[193, 129, 65, 1]`); it now goes through `pixels_to_uint8`. `examples/train/configs/overfit_minimax_h3_t2va.yaml` selects this callback with the H3 pipeline.
- `fastvideo/training/training_pipeline.py`, `fastvideo/training/distillation_pipeline.py`, `fastvideo/entrypoints/streaming_generator.py`: the three other direct consumers with the same `(x * 255)` loop use the helper too.
- `fastvideo/entrypoints/video_generator.py`: post-decode uses the same helper instead of an inline dtype switch (float output keeps the on-device quantize path from #1362); the optional `samples` mirror stays on its `[0, 1]` float contract (`copy_` then `div_(255)`, or `float().div_(255)` on the shape-mismatch path).
- `fastvideo/tests/stages/test_minimax_h3_vae_streaming.py` (review follow-up): `test_decode_stage_uses_cpu_output_buffer` asserts the VAE streamed into a float32 CPU buffer and that the stage returns its CPU uint8 quantization, pixel-exact (`0.25 -> 63`, `0.5 -> 127`, `1.0 -> 255`, overshoot `1.25 -> 255`, `-0.25 -> 0`); `test_parallel_decode_runs_on_every_rank` asserts `127` uint8 on the leader while keeping the rank-participation checks.
- `fastvideo/tests/train/callbacks/test_validation.py` (review follow-up): `TestH3ValidationContract` gains `test_run_validation_for_steps_passes_uint8_frames_through` (uint8 `[63, 127, 191, 255]` reaches the frames byte for byte through `_run_validation_for_steps`) and `test_run_validation_for_steps_quantizes_float_frames` (`[0.25, 0.5, 0.75, 1.0]` still maps to the same bytes).
- `fastvideo/tests/entrypoints/test_video_generator.py`: `test_generate_single_video_accepts_uint8_worker_output` checks that frames come back byte-identical and that `samples` is `[0, 1]` float.

## Test Plan

```bash
SRC="fastvideo/utils.py fastvideo/train/callbacks/validation.py fastvideo/training/training_pipeline.py \
     fastvideo/training/distillation_pipeline.py fastvideo/entrypoints/streaming_generator.py \
     fastvideo/entrypoints/video_generator.py fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py"
ruff check $SRC
yapf --diff $SRC
codespell --toml pyproject.toml $SRC fastvideo/tests/stages/test_minimax_h3_vae_streaming.py \
  fastvideo/tests/train/callbacks/test_validation.py fastvideo/tests/entrypoints/test_video_generator.py
# as the pre-commit hook runs it: a venv with only mypy 1.15.0 + the hook's type-stub packages
mypy --python-version 3.10 --follow-imports skip --disable-error-code union-attr --disable-error-code override $SRC
python -m pytest fastvideo/tests/stages/test_minimax_h3_vae_streaming.py \
  fastvideo/tests/train/callbacks/test_validation.py fastvideo/tests/entrypoints/test_video_generator.py -q
```

## Test Results

<details>

```
ruff:      All checks passed!
yapf:      no diff
codespell: clean
mypy:      Success: no issues found in 6 source files
pytest:    85 passed in 4.34s   (CPU, macOS arm64, torch 2.14.0)
           = 4 stage tests + 48 validation-callback tests + 33 video-generator tests
```

Before the follow-up commit, the two stage tests failed exactly as described in review (`result.output == 0.25` / `== 0.5` against uint8 `63` / `127`).

</details>

The same worker-side cast has been running in a downstream 4xH100 deployment (Windflow streaming worker on FastVideo `b2db0c0`, which consumes the uint8 frames through `return_frames=True`), including a 30-minute continuous run of 51 clips.

## Checklist

- [x] ruff, yapf, codespell and mypy run with the repository configuration on the touched files (`pre-commit` itself is not installed on the authoring machine)
- [x] I updated documentation if needed (none needed)
- [x] I considered GPU memory impact of my changes (none: the cast happens on the CPU buffer; the worker briefly holds the 1.07 GB uint8 tensor next to the 4.27 GB fp32 buffer before the latter is released)

- [ ] SSIM regression not run here; the frames are byte-identical by construction (same `*255`, clamp and cast, moved across the process boundary) and the unit tests assert exact equality
- [x] Support matrix unchanged


fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py
```diff
@@ -153,6 +153,13 @@ def forward(self, batch: ForwardBatch, fastvideo_args: FastVideoArgs) -> Forward
                     decode_to_pixels_parallel(self.vae, latents, output, sp_group, strategy=strategy)
                 else:
                     self.vae.decode_to_pixels(latents, output)
+            if is_output_rank and output is not None and output.dtype == torch.float32:
+                # Quantize to uint8 here, before the frames cross the
+                # multiprocess-executor boundary. A 345-frame 1344x768 clip is
+                # 4.27 GB as fp32 and 1.07 GB as uint8, and the main process
+                # cast it to uint8 immediately anyway; the clamp matches the
+                # post-decode handling of VAE output slightly outside [0, 1].
+                output = output.mul_(255).clamp_(0, 255).to(torch.uint8)
```
SolitaryThinker, line 162: [P1] Handle uint8 output in training validation before changing the stage contract

This also changes direct pipeline calls, which bypass the new VideoGenerator handling. The training validation callback reads output_batch.output directly and still applies `(x * 255).numpy().astype(np.uint8)` in fastvideo/train/callbacks/validation.py:1575. The shipped examples/train/configs/overfit_minimax_h3_t2va.yaml selects this H3 pipeline, so validation videos now suffer uint8 overflow: decoded values [0.25, 0.5, 0.75, 1.0] produce pixel bytes [63, 127, 191, 255], which validation transforms into [193, 129, 65, 1]. Please update this consumer to pass uint8 through and add regression coverage.

fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py
```diff
@@ -153,6 +153,13 @@ def forward(self, batch: ForwardBatch, fastvideo_args: FastVideoArgs) -> Forward
...
             batch.output = output if is_output_rank else placeholder
```
SolitaryThinker, line 163: [P2] Update existing stage tests for the new uint8 output contract

fastvideo/tests/stages/test_minimax_h3_vae_streaming.py still asserts that the serial result is the original float output buffer with values of 0.25 (lines 112–114), and that the parallel result contains 0.5 (line 192). This conversion instead returns a new uint8 tensor containing 63 or 127, respectively, so both tests contradict the new behavior. Please update them to verify quantization while preserving their CPU-buffer and rank-participation checks; the added entrypoint test does not exercise either stage path.

SolitaryThinker (changes requested):
Two issues need addressing: training validation still assumes normalized float output, and existing decode-stage tests contradict the new uint8 contract.

Validation: reproduced uint8 overflow with a CPU PyTorch arithmetic check. Attempted the entrypoint and H3 streaming-stage pytest suites, but collection was blocked by a Triton dependency requiring an active GPU driver. No GPU inference or SSIM run was performed.

fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py
stayinalive181: Done in af5025e. Added `fastvideo.utils.pixels_to_uint8` (uint8 passes through untouched, float is clamped to [0, 1] and cast) and routed the `ValidationCallback` frame loop through it. The same `(x * 255)` loop also lives in the legacy `TrainingPipeline`/`DistillationPipeline` validation paths and in `StreamingVideoGenerator._process_output_batch`, so those three direct consumers use the helper as well, and `VideoGenerator` now calls it instead of its inline dtype switch.

Regression coverage in `TestH3ValidationContract`: `test_run_validation_for_steps_passes_uint8_frames_through` feeds uint8 [63, 127, 191, 255] through `_run_validation_for_steps` and asserts the frames come out byte for byte (the old loop produced [193, 129, 65, 1]); `test_run_validation_for_steps_quantizes_float_frames` checks [0.25, 0.5, 0.75, 1.0] still map to the same bytes.

fastvideo/pipelines/basic/minimax_h3/stages/minimax_h3_decoding.py
stayinalive181: Done in af5025e. `test_decode_stage_uses_cpu_output_buffer` now asserts the VAE streamed into a float32 CPU buffer and that the stage returns its CPU uint8 quantization, pixel-exact: 0.25 -> 63, 0.5 -> 127, 1.0 -> 255, and the overshoot cases 1.25 -> 255 and -0.25 -> 0 for the clamp. `test_parallel_decode_runs_on_every_rank` asserts uint8 127 on the leader and keeps the rank-participation and non-leader placeholder checks. Both failed on the previous head exactly as you described (0.25 / 0.5 against 63 / 127) and pass now.

stayinalive181:
Pushed af5025e addressing both review comments (threads answered inline). Summary of the follow-up:

- `fastvideo.utils.pixels_to_uint8`: uint8 passes through, float is clamped and cast. The `ValidationCallback` frame loop, the legacy `TrainingPipeline`/`DistillationPipeline` validation loops, `StreamingVideoGenerator` and `VideoGenerator` all use it, so no direct consumer of `output_batch.output` multiplies uint8 by 255 any more.
- Stage tests assert the uint8 contract on both the serial and the parallel decode path (including the clamp of VAE overshoot); callback tests cover uint8 pass-through and float quantization through `_run_validation_for_steps`.
- Local run (CPU, macOS arm64): ruff / yapf / codespell clean, mypy clean in a venv set up like the pre-commit hook, `pytest` 85 passed across the three touched test files. Commands and output are in the updated description.

One note for the fastcheck status: the `microscope-unit-tests` lane also fails on recently merged PRs (#1824, #1825, #1831), so it looks unrelated to this change; happy to look further if you see something specific in that lane.

stayinalive181:
@SolitaryThinker both review points are addressed in the second commit (threads answered inline), and I've just rebased onto current `main` (`37d06a8`) with no code changes; `main` had not touched any file in this PR since the old base. Could you take another look when you have a moment? The unit-test lane failure is the same unrelated one you noted on #1829.

PR #1829, opened 2026-09-07
[feat] MiniMax H3: block-FP8 text encoder on Hopper, its converter, and a 4xH100 FastH3 profile

stayinalive181:
## Purpose

FastVideo gates the serialized block-FP8 Qwen3-VL text encoder on Blackwell, but only the FlashInfer groupwise GEMM is Blackwell-specific. On Hopper the encoder therefore cannot stay resident beside an FSDP-sharded DiT on 80 GB and is offloaded on every request: 1.82 s per clip on 8xH200 versus 0.04 s resident ([hlander-ai/minimax-h3](https://github.com/hlander-ai/minimax-h3)). This PR lets sm_90 load the same FP8 checkpoint: the weights stay FP8-resident (35.5 GB instead of 66.7 GB) and each language-model linear dequantizes its block-scaled weight to BF16 right before a plain matmul, which is the arithmetic the FlashInfer path performs internally. FP8 buys residency here rather than speed. The route and the converter recipe come from hlander-ai/minimax-h3 (Apache-2.0).

With that in place a 4xH100 FastH3 profile becomes runnable: FSDP-sharded DiT, ThunderKittens sm_90a VSA kernel, resident FP8 encoder.

## Changes

- `fastvideo/models/encoders/minimax_h3_checkpoint_fp8.py`: `get_min_capability` 100 -> 90; `validate_runtime` no longer requires the Triton quantizer or the FlashInfer GEMM on sm_90; `apply` dequantizes through `_dequantize_block_fp8_weight` on sm_90. The Blackwell path is unchanged.
- `scripts/checkpoint_conversion/quantize_minimax_h3_text_encoder_fp8.py`: writes the serialized checkpoint the loader already reads (E4M3 `weight` plus `[N/128, K/128]` `weight_scale_inv`, `quantization_config` in `config.json`, regenerated index). There was no producer for this format in-tree.
- `examples/inference/basic/basic_fasth3.py`: `--vsa-kernel tk` (ThunderKittens sm_90a, `FASTVIDEO_VSA_TK=1`), `--text-encoder-weights`, `--offload-text-encoder` / `--offload-vae` (defaults unchanged), and an sm_90 extension check in `validate_profile_dependencies`.
- `examples/inference/basic/basic_fasth3_h100.yaml` and a "Hopper (H100 80 GB)" README section with the measured points.
- Tests: `validate_runtime` accepts (9, 0) without the Blackwell dependencies; the dequantizer is bit-exact against the reference broadcast and the FP8 round trip stays within E4M3 precision; the `tk` profile's environment, config and dependency check.

## Test Plan

```bash
ruff check fastvideo/models/encoders/minimax_h3_checkpoint_fp8.py \
  fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py \
  fastvideo/tests/inference/test_basic_fasth3_profile.py \
  examples/inference/basic/basic_fasth3.py \
  scripts/checkpoint_conversion/quantize_minimax_h3_text_encoder_fp8.py
yapf --diff / codespell --toml pyproject.toml on the same files
python -m pytest fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py -q
python -m pytest fastvideo/tests/inference/test_basic_fasth3_profile.py -q
```

## Test Results

<details>

```
ruff:      All checks passed! (the only report is the pre-existing SIM102 at minimax_h3_checkpoint_fp8.py:381, untouched)
yapf:      touched hunks clean (the file is outside pre-commit and was never yapf-formatted upstream)
codespell: clean
pytest fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py: 13 passed in 0.57s
pytest fastvideo/tests/inference/test_basic_fasth3_profile.py:                15 passed in 0.07s
(CPU, macOS arm64, torch 2.14.0)
```

</details>

Measured on FastVideo `b2db0c0` with this encoder path (warm, loading and compilation excluded, 345 frames = 14.375 s of playback):

| GPUs | Resolution | Generation | Source |
|---|---|---|---|
| 8x H100 80 GB | 1344x768 | 13.506 s (13.503 / 13.506 / 13.554) | hlander-ai/minimax-h3, Triton VSA |
| 4x H100 80 GB, NVLink | 960x544 | 12.98-14.02 s over 8 consecutive clips | Windflow streaming worker: TK VSA, FSDP DiT, resident FP8 encoder |

The 4xH100 row was measured through a downstream streaming worker rather than `basic_fasth3.py`; `basic_fasth3.py` numbers for that shape on a fresh 4xH100 box can be added to this PR on request.

## Checklist

- [x] ruff, yapf and codespell run with the repository configuration on the touched files (`pre-commit` itself is not installed on the authoring machine)
- [x] I updated documentation if needed (README section, example YAML, `scripts/checkpoint_conversion/AGENTS.md`)
- [x] I considered GPU memory impact of my changes (sm_90 only: one dequantized BF16 weight is materialized per linear call, at most ~0.26 GB for the largest layer, and released after the matmul; the resident encoder drops from 66.7 GB to 35.5 GB)

- [ ] SSIM regression not run here; the FP8 encoder is a precision change relative to the stock BF16 encoder on every architecture, and this PR does not change the Blackwell numerics
- [x] Support matrix unchanged (Hopper is a new runtime for an existing model)


SolitaryThinker:
Pushed 3247699f2 (rebased onto `main`): the H100 profile now keeps the FP8 text encoder resident, and the converter fails closed instead of writing shards the loader rejects.

**Fixed**
- `examples/inference/basic/basic_fasth3_h100.yaml:37,50`: `lazy_module_load: true` and `h3_sequential_load: true` made the pipeline free the 35.5 GB FP8 encoder after conditioning and reload it on the next request, contradicting the profile's "nothing is offloaded between requests" claim and the CLI route the README documents (both flags default to auto-off on discrete GPUs). Both are now explicitly `false`, with a header note explaining why.
- `scripts/checkpoint_conversion/quantize_minimax_h3_text_encoder_fp8.py:54`: the converter padded non-128-divisible linears and wrote ceil-shaped `weight_scale_inv`, which `MiniMaxH3SerializedFP8LinearMethod.process_weights_after_loading` rejects. It now raises at conversion time.
- `fastvideo/models/encoders/minimax_h3_checkpoint_fp8.py:486-502`: dropped the now-unreachable padding branch in `_dequantize_block_fp8_weight` and made the block-scale multiply in-place (`mul_`), halving the peak transient per linear.
- `fastvideo/models/encoders/minimax_h3_checkpoint_fp8.py:467-475`: corrected the "exact-arithmetic substitution" comment; the Blackwell path also quantizes activations per token group and runs an FP8 GEMM, so the Hopper route is a different (more accurate) arithmetic, not the same one.
- `fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py:203,219,228`: added a converter -> dequantizer round-trip test, a non-divisible rejection test, and a CUDA-gated test that the sm_90 `apply` branch dequantizes instead of calling the FlashInfer GEMM.

**Verified**
- `pytest fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_checkpoint_fp8.py`: 15 passed, 1 skipped (the new CUDA test; CPU-only sandbox)
- `pytest fastvideo/tests/inference/test_basic_fasth3_profile.py`: 15 passed
- `pytest fastvideo/tests/encoders/test_minimax_h3_qwen3_vl_truncation.py`: 13 passed
- `build_generate_run_config` on the YAML resolves `lazy_module_load=False`, `h3_sequential_load=False`, `text_encoder/vae offload=False`
- converter -> dequantizer round trip on 8192x5120 and 25600x5120: max abs error 0.203 (E4M3 precision)
- `py_compile` and 120-column check on the touched files; rebase onto `origin/main` (37d06a832) clean

**Left open**
- `fastvideo/tests/inference/test_basic_fasth3_profile.py` is collected by no CI lane (`.buildkite/scripts/unit_test.sh` omits `fastvideo/tests/inference/`; only its `lora/` and `vmoba/` subdirs have lanes), so the two profile tests this PR adds there do not run in CI. Wiring the file into a lane is a CI-ownership decision (planner path ownership + private lane table), so it is not changed here; the encoder-side tests above do run in the encoder lane.
- `--vsa-kernel tk` does not change kernel selection on sm_90 (the kernel's `_force_tk()` branch is identical to its default, and `--vsa-kernel triton` does not set `FASTVIDEO_VSA_TRITON`); the flag still sets `FASTVIDEO_VSA_SM100A=0` and preflights the extension. The selection logic lives in `fastvideo-kernel/`, outside this PR.
- The unit-test lane is red on this PR, but not for a diff-caused reason: the diff touches no unit-lane path, and unrelated PRs from the same base (1826, 1827, 1828, 1830, 1832) fail the same lane. The rebase gives CI a current-base run.

stayinalive181:
Thanks for taking this over and fixing both problems. You're right about the profile: with `lazy_module_load` and `h3_sequential_load` enabled, the encoder was freed and reloaded on every request, which is exactly what this profile is meant to avoid. Failing at conversion time is also the right behaviour for shapes the serialized loader cannot take.

One housekeeping note: I force-pushed this branch once to clean up my own commit message. Both commits keep their trees and authors, your commit message is unchanged, and the head is now `a2e7fba8`, so the content CI sees is identical to your push.

On the two points you left open:

1. **Profile tests not collected.** `test_basic_fasth3_profile.py` locates the repo root with `parents[3]`, so it can move unchanged into a directory the unit lane already collects, such as `fastvideo/tests/entrypoints/`. I can push that move if it fits your path ownership; otherwise it can stay where it is.
2. **`--vsa-kernel tk`.** Agreed. Since `_force_tk()` matches the default branch on sm_90, the flag does not choose a kernel on Hopper; it only turns off the sm_100a route and preflights the extension. I suggest removing the description that says otherwise: the `tk` help text should say exactly that, and the Hopper section of the README should stop presenting the flag as the way to select ThunderKittens. Our H100 timings should not be read as a kernel comparison either. I can push that wording change, or drop the `tk` choice entirely if you prefer.

PR #1840, opened 2026-09-14
[bugfix] Add OpenAI-compatible image generation endpoint

sudhirpol522:
Adds `POST /v1/images/generations`, the endpoint used by the OpenAI Python client, while preserving `POST /v1/images`.

Regression tests cover both routes, base64 responses, and URL downloads. Before the fix, the two standard-route cases returned 404; all four cases now pass.

Validation: changed-file pre-commit checks passed. Tests ran on macOS against the real router with a mocked generator and a lightweight import bootstrap that bypasses GPU package initialization; no GPU inference was run.


SolitaryThinker (approved):

SolitaryThinker:
Pushed `fededc675`: documented `POST /v1/images/generations` in the server-contract table and alias note, renamed the route test to `test_image_generation_route_compatibility.py`, and gave the parametrize axes explicit IDs. Local run: 4/4 pass. Note `bench_serving.py` already posts text-to-image to this route, so the fix also unblocks it.

PR #1841, opened 2026-09-14
[bugfix] Validate image response format before generation

sudhirpol522:
Rejects unsupported image `response_format` values with HTTP 400 before generation or output-directory creation. Preserves the existing default, case normalization, and supported formats.

Regression tests verify zero generator calls for invalid formats and unchanged behavior for supported values. The two invalid-format cases fail before the fix; all eight cases pass afterward.

Validation: changed-file pre-commit checks passed. Tests ran on macOS against the real router with a mocked generator and a lightweight import bootstrap that bypasses GPU package initialization; no GPU inference was run.


SolitaryThinker:
## Review — `[bugfix] Validate image response format before generation`

**Verdict:** the `generations` fix is correct and tested; one sibling endpoint with the same bug class is left unfixed.

### Findings

**1. `[medium]` `/v1/images/edits` still silently accepts invalid `response_format`** — `fastvideo/entrypoints/openai/image_api.py:222`
The PR validates `response_format` only in `generations`. `edits` does `resp_format = (response_format or "b64_json").lower()` and then `if resp_format == "b64_json": ... else: <url>`, so `response_format=invalid` (or any typo) silently returns a URL response instead of a 400 — the same bug this PR fixes, minus even the late 400. Since `edits` is the sibling image-generation endpoint, it should get the same pre-generation validation (3 lines), or the PR body should state it is deliberately out of scope.

**2. `[low]` Omitted `response_format` is not covered by the test** — `fastvideo/tests/entrypoints/test_image_response_format.py:49`
`ImageGenerationsRequest.response_format` defaults to `"url"` (`protocol.py:32`), while the endpoint's `or "b64_json"` fallback maps explicit `null`/`""` to `b64_json`. The parametrization covers `None` and `""` but never omits the key, so the actual default path (omitted → URL) is untested. Add a case without `response_format` and assert the URL response.

**3. `[nit]` Test filename convention** — `fastvideo/tests/AGENTS.md` asks for `test_<feature>_<expected_behavior>.py`; `test_image_response_format.py` is missing the behavior half (e.g. `test_image_response_format_validation.py`).

**4. `[nit]` Duplicated supported-format set** — `{"b64_json", "url"}` is now hardcoded at `image_api.py:91` alongside the protocol comment at `protocol.py:32`. A module-level constant would keep them in sync (a `Literal` on the model would change the status code to 422, so it is not a drop-in).

### Verified

- Validation runs before `_build_generation_kwargs`, so no output directory is created and no generator call happens for invalid formats — backed by `assert not (tmp_path / "images").exists()` and `assert_not_awaited()`.
- 400 + lowercased detail preserved from the old late check; case normalization and the `b64_json`/`url` branches are unchanged.
- The new test runs in the unit lane (`unit_test.sh` includes `fastvideo/tests/entrypoints/`); the diff also triggers the `api-server` merge lane via `plan_merge_ci.py:467`. `buildkite/pr-fastcheck` was still pending at review time.
- PR-body claim of 8 cases (2 invalid + 6 supported) matches the parametrization.

### Not verified

- Could not run pytest locally (no torch/fastapi in this sandbox); the CI unit lane is the gate. `py_compile` and the 120-col check pass on both changed files.

SolitaryThinker:
## Simplification / efficiency pass

Follow-up to the correctness review above — same head (`7b5c99fed`).

**S1. `[medium]` Fold normalize + validate into one shared helper** — `fastvideo/entrypoints/openai/image_api.py:90-92`, `:222`
The new check sits on top of the existing `(value or "b64_json").lower()` normalization, which is already duplicated in `edits`. One helper dedupes it and closes the `edits` gap at the same time:

```python
_SUPPORTED_RESPONSE_FORMATS = frozenset({"b64_json", "url"})


def _normalize_response_format(value: str | None) -> str:
    fmt = (value or "b64_json").lower()
    if fmt not in _SUPPORTED_RESPONSE_FORMATS:
        raise HTTPException(status_code=400, detail=f"response_format={fmt} is not supported")
    return fmt
```

Then `resp_format = _normalize_response_format(request.response_format)` at the top of `generations`, and `resp_format = _normalize_response_format(response_format)` at the top of `edits` (before the uploads dir is created and before generation), deleting the assignment at line 222. Net: one named constant instead of an inline literal, no duplicated normalization, and the sibling endpoint gets the same 400 for free.

**S2. `[low]` Parametrize the expected format instead of re-deriving it** — `fastvideo/tests/entrypoints/test_image_response_format.py:49-61`
`if response_format and response_format.lower() == "url"` recomputes the endpoint's normalization inside the assertion. Put the expectation in the table instead:

```python
@pytest.mark.parametrize("response_format,expected", [
    (None, "b64_json"), ("", "b64_json"),
    ("b64_json", "b64_json"), ("B64_JSON", "b64_json"),
    ("url", "url"), ("URL", "url"),
])
```

and branch on `expected`. The table then documents the contract (including the null/empty → `b64_json` special case) instead of encoding it as logic.

**S3. `[info]` No efficiency work needed in the production change.**
- The check is O(1), and CPython compiles `x not in {"b64_json", "url"}` to a `frozenset` constant in the code object (verified with `dis`), so there is no per-request set allocation.
- `.lower()` allocates one short string per request — negligible.
- The fix is itself an efficiency win: invalid requests now skip model generation and output-dir creation entirely.
- In the test, a fresh `FastAPI()`/`TestClient` per parametrized case is cheap (no lifespan) and buys function-scoped isolation; hoisting it would trade clarity for nothing.

**S4. `[nit]` Response-data construction is duplicated between `generations` and `edits`** — `image_api.py:122-138` vs `:223-243`
The b64_json/url branches are near-identical but not identical: `generations` checks the file exists and omits `file_path` from the b64 payload, while `edits` does neither. A shared builder would need flags or would change behavior, so it is not worth doing in this bugfix — flagging only so it stays a conscious decision.

**S5. `[info]` Dead code handled correctly.** The old trailing `else: raise` in `generations` is gone and the remaining `else` at line 122 is exhaustive under the new validation; no unreachable branch or stale comment is left behind.

SolitaryThinker:
## Fixes applied (maintainer push)

Pushed `594b8e4cc` addressing the review findings:

- **Finding 1 / S1**: extracted `_normalize_response_format` + `_SUPPORTED_RESPONSE_FORMATS` and used it at the top of both `generations` and `edits`. `/v1/images/edits` now rejects unsupported formats with 400 before the uploads dir is created or the generator runs.
- **Finding 2 / S2**: the supported-format test now parametrizes `(payload, expected)` and covers the omitted-field case (omitted → protocol default `url`; explicit `null`/`""` → `b64_json`).
- **Finding 3**: renamed the test to `test_image_response_format_validation.py` per the `test_<feature>_<expected_behavior>.py` convention.
- **Finding 4**: the inline set is now the named `_SUPPORTED_RESPONSE_FORMATS` constant.
- Added `test_edits_invalid_response_format_does_not_generate` (fails on the previous head, passes now) and `test_edits_supported_response_format_still_generates`.

Local verification: 11/11 tests in the file pass against the real router with a mocked generator (heavy imports stubbed, no GPU). The edits regression test fails on `7b5c99fed` and passes on `594b8e4cc`. `py_compile` and the 120-col check pass. CI is re-running.

PR #1843, opened 2026-09-14
[bugfix]: respect serialized tensor dtype in parquet dataloader

Yaegaki1Erika:
## Purpose

Fix Parquet tensor deserialization incorrectly treating every serialized tensor as `float32`.

The Parquet schema stores each tensor as a `*_bytes`, `*_shape`, and `*_dtype` triplet, but the dataloader previously ignored `*_dtype`. This caused tensors such as `uint8` PIL images to be decoded with an incorrect element count and fail during reshape, for example:
  ```text
  ValueError: cannot reshape array of size 299520 into shape (3,480,832)
```
## Changes

- Decode Parquet tensor buffers using their serialized `*_dtype` metadata.
- Support NumPy and PyTorch dtype strings, including `uint8`, `float16`, `float32`, `float64`, and `bfloat16`.
- Preserve raw `bfloat16` storage by decoding it through `uint16` before reinterpreting it as `torch.bfloat16`.
- Validate that the serialized byte length matches the declared shape and dtype before reshaping.
- Add regression tests covering mixed tensor dtypes, `bfloat16`, invalid byte lengths, and legacy aliases.

## Test Plan

```bash
python -m pytest -p no:cacheprovider \
  fastvideo/tests/dataset/test_parquet_dtype_roundtrip.py -vv -s
```

## Test Results

<details>

```text
collected 4 items

fastvideo/tests/dataset/test_parquet_dtype_roundtrip.py::test_i2v_collate_respects_each_serialized_tensor_dtype PASSED
fastvideo/tests/dataset/test_parquet_dtype_roundtrip.py::test_decode_tensor_bytes_accepts_torch_bfloat16_dtype PASSED
fastvideo/tests/dataset/test_parquet_dtype_roundtrip.py::test_decode_tensor_bytes_rejects_mismatched_byte_length PASSED
fastvideo/tests/dataset/test_parquet_dtype_roundtrip.py::test_legacy_collate_helper_respects_dtype_for_alias_keys PASSED

4 passed, 14 warnings in 2.71s
```

</details>


SolitaryThinker:
Thanks for the fix, @Yaegaki1Erika — this correctly addresses a real data-path bug (uint8 `pil_image` was being decoded as float32). I pushed a small follow-up commit to your branch (`c222e653`) that keeps the behavior identical while tightening the implementation:

**Efficiency**
- `_decode_tensor_bytes` now takes a keyword-only `zero=True` path: CFG-dropped `text_embedding` rows still get their byte length validated, but return a correctly typed zero tensor instead of decoding the payload and zeroing it afterwards. This restores the old code's "no copy on drop" behavior while keeping the new dtype awareness.
- `_normalize_tensor_dtype` is memoized with `functools.lru_cache`, and the alias table is now a module-level constant, so repeated dtype strings across rows skip the string work.
- `math.prod(shape)` replaces `int(np.prod(shape, dtype=np.int64))`.
- Per-tensor `shape`/`bytes`/`dtype` key lookups are hoisted out of the row loop.

**Simplifications**
- Single `is_bfloat16` flag instead of comparing the dtype name twice.
- `str.removeprefix` for the `torch.` / `numpy.` / `np.` prefixes.
- Removed a dead `serialized_key = key` initialization in the alias-key branch. I kept the existing last-match-wins alias behavior rather than flipping it to first-match, since that would be an unrelated behavior change.

**Tests / docs**
- Added `test_decode_tensor_bytes_zero_returns_typed_zeros` and `test_collate_cfg_drop_zeros_text_embedding_with_serialized_dtype`.
- Fixed the now-stale "reads the byte payload as float32" comment in `preprocess_minimax_h3_overfit.py`.

I did not switch to `torch.frombuffer` (which would drop the numpy round-trip and the bfloat16 special case) because it needs runtime verification on the pinned torch 2.12; happy to do that as a follow-up.

CI should re-run on the new commit. Thanks again for the contribution!

SolitaryThinker (approved):

PR #894, opened 2025-11-21
[feat] Support sequence packing and shard after pachification for USP

loaydatrain:
Previously, sequence parallelism was implemented by sharding full latent frames along the temporal dimension. This update refactors the implementation to shard after patchification within the DiT block.

Additionally, it introduces padding logic to handle scenarios where the sequence length is not a perfect multiple of the sequence parallel size (sp_size), ensuring robustness across varying input lengths.

PR #1844, opened 2026-09-14
[bugfix]: fix Wan I2V DMD conditioning under sequence parallelism

Yaegaki1Erika:
## Purpose

Fix Wan I2V DMD conditioning being pre-sharded incorrectly when sequence parallelism is enabled.

The distillation pipeline previously split only `image_latents` along the latent temporal dimension before concatenating them with the unsharded noisy video input. This could either fail when the latent temporal length was not divisible by the SP world size, or produce incompatible temporal dimensions during concatenation.

`WanTransformer3DModel` already performs sequence-parallel sharding after patch embedding, over the flattened spatiotemporal token sequence. The pipeline should therefore construct and pass the complete model input instead of pre-sharding only its image-conditioning component.

## Changes

- Remove pipeline-level sequence-parallel sharding of Wan I2V image conditioning.
- Keep the mask and first-frame latent conditioning at the full temporal length until it is concatenated with the noisy video latent.
- Let `WanTransformer3DModel` perform sequence-parallel sharding after patch embedding.
- Add explicit validation that the noisy latent and image conditioning have matching batch, temporal, and spatial dimensions before concatenation.
- Add regression tests covering:
  - preservation of the complete temporal dimension when `sp_world_size > 1`;
  - the expected mask-plus-latent conditioning layout;
  - rejection of mismatched latent temporal lengths.

## Test Plan

```bash
python -m pytest -p no:cacheprovider \
  fastvideo/tests/training/distill/test_wan_i2v_distillation.py \
  -vv -s
```

## Test Results

<details>

```text
collected 3 items

fastvideo/tests/training/distill/test_wan_i2v_distillation.py::test_i2v_prepare_inputs_does_not_pre_shard_conditioning PASSED
fastvideo/tests/training/distill/test_wan_i2v_distillation.py::test_i2v_conditioning_preserves_full_temporal_length_for_sequence_parallelism PASSED
fastvideo/tests/training/distill/test_wan_i2v_distillation.py::test_i2v_conditioning_rejects_temporal_length_mismatch PASSED

3 passed, 14 warnings in 4.67s
```

</details>


SolitaryThinker (approved):
Fix is correct. The pipeline-level temporal pre-sharding of image conditioning was stale after WanTransformer3DModel moved to sharding the flattened token sequence after patch embedding (#894): the full-length noise input was concatenated with a temporally sharded conditioning tensor, which cannot match for sp_world_size > 1. Removing the pre-shard and passing full-length mask+latent conditioning matches the non-distillation WanI2VTrainingPipeline and the T2V WanDistillationPipeline. Mask math, device/dtype handling, and the new shape validation all check out.

PR #1848, opened 2026-09-15
[bugfix]: isolate cancelled streaming requests from GPU result readers

Gxj230958:
Cancelling `SubprocessGpuPool.run()` currently propagates cancellation through
`asyncio.wrap_future` to the pending concurrent future. When the worker replies,
the result reader tries to resolve that cancelled future and exits with
`InvalidStateError`, breaking subsequent jobs on the same worker. Cancellation
while waiting for executor-backed queue dispatch also leaves a pending entry.

Always remove the pending entry when `run()` exits, cancel unfinished futures,
and ignore completed futures in the result reader to cover cancellation callback
ordering. The worker may still finish an abandoned generation; its late success
or error is discarded without stopping result delivery for later requests.

Added six CPU regression cases using the existing thread worker factory and real
multiprocessing queues. Four exercise cancellation during dispatch/generation
with late success/error responses; two force the cancelled-future callback race.
All six fail against the unchanged production baseline. The patched component
suite passes all 23 tests. Local execution bypassed eager package initializers
and distributed conftest; full package/GPU validation is still pending.
The repository's pre-commit command for both changed files passes, including
yapf, ruff, codespell, and mypy. Markdown/workflow hooks are skipped by their
configured file filters. Component tests still pass after the hook run.

Both failure modes are still present on current `main` (checked 2026-09-15):
`run()` awaits the wrapped future without cleanup, and `_drain_results` resolves
futures without a done-guard in its hot path. Code and tests were developed
with AI assistance.


SolitaryThinker:
Pushed `4b2406d48`: log dropped late replies in the pool reader and tidy the cancellation-test imports.

**Fixed**
- `_drain_results` dropped replies for abandoned/unknown jobs silently, so an operator could not tell "abandoned result discarded" from "result lost" -> added a `logger.debug` on the discard path (`fastvideo/entrypoints/streaming/gpu_pool.py:425-427`).
- Function-local imports in the new race test -> hoisted `Future` and `_PendingJob` into the module import block and reworded the docstring to say what the test actually asserts (`fastvideo/tests/entrypoints/streaming/test_gpu_pool.py:15,32,470`).

**Verified**
- `python3 -m py_compile` on both changed files and a 120-column check: clean. No tests run locally (no project env here); CI is the gate.
- Three independent review lenses (correctness / efficiency / simplification) read the diff plus the writer/reader paths. All three agree the fix is correct: `CancelledError` is not an `Exception`, so the old `except Exception` never popped `_pending`, and the reader then hit `InvalidStateError` on the cancelled future, killing the reader and hanging every later job on that worker. The `finally` pop plus the `done()` guard closes both paths.
- The `done()` guard is load-bearing, not dead code: `_chain_future` registers `_call_check_cancel` before the task's `__wakeup`, so the source future can already be cancelled while its `_pending` entry is still present. Kept as-is.
- `future.cancel()` in the `finally` is not a no-op on the cancel-during-dispatch path (the task is awaiting the executor future there, so the source future is still pending). Kept.
- The new tests are collected by the unit lane (`./fastvideo/tests/entrypoints/` in `.buildkite/scripts/unit_test.sh`), one of the six fastcheck lanes; `fastvideo/entrypoints/` also selects the `api-server` merge-gate lane.

**Left open**
- `InProcessGpuPool.run` (`fastvideo/entrypoints/streaming/gpu_pool.py:174-180`) releases `_gen_lock` as soon as the awaited executor future is cancelled while `generate()` keeps running in the thread, so a second `run()` can enter `generate()` concurrently on one GPU. Pre-existing on `main` and outside this PR's `SubprocessGpuPool` scope; fixing it changes cancellation semantics for the default pool, so it needs its own PR and a design decision.
- The reader resolves futures without a `try/except InvalidStateError`; the `done()` guard is correct today because every `_pending`/future mutation happens on the loop thread, so I did not add a broad `except` that could mask real bugs.
- `run()` still has no timeout, and a dead reader is neither restarted nor surfaced in `health()`; out of scope.
- CI: `pre-commit` passed on the previous head. `buildkite/pr-fastcheck` has been queued ("scheduled") for 20+ minutes — the fastcheck queue is backed up across open PRs, so the unit lane that runs these tests has not executed yet. It re-triggers for `4b2406d48`.

SolitaryThinker (approved):

PR #1849, opened 2026-09-15
[bugfix] ROCm: detect the platform from a HIP torch build when amdsmi is unavailable

stayinalive181:
## Purpose

On AMD GPUs running the official `rocm/pytorch` images, FastVideo resolves the **CPU** platform. `rocm_platform_plugin()` only recognizes ROCm through the `amdsmi` Python package, and those images ship a ROCm (HIP) torch build but no `amdsmi` package (no `/opt/rocm/share/amd_smi`, no `libamd_smi.so`). Installing the PyPI `amdsmi` wheel as a workaround is worse: it breaks `import torch` itself (`torch/cuda/__init__.py` imports `amdsmi`, which fails with `KeyError: 'libamd_smi.so'`).

## Changes

- `fastvideo/platforms/__init__.py`: when the amdsmi check does not find a device, accept a HIP torch build (`torch.version.hip`) together with the AMD compute device node `/dev/kfd`. The check reads only `torch.version.hip` and whether the process can open `/dev/kfd`, so detection still does not initialize the GPU runtime. The amdsmi path is unchanged and still runs first.
- `fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py`: HIP torch plus an accessible device node resolves `RocmPlatform`; HIP torch without it, and a CUDA torch build, do not. CPU-only, amdsmi simulated as missing.

## Test Plan

```bash
ruff check fastvideo/platforms/__init__.py fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py
yapf --diff fastvideo/platforms/__init__.py fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py
codespell --toml pyproject.toml fastvideo/platforms/__init__.py fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py
mypy --python-version 3.10 --follow-imports skip --disable-error-code union-attr --disable-error-code override fastvideo/platforms/__init__.py
python -m pytest fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py -q
```

On an AMD Instinct MI355X (gfx950) in `rocm/pytorch:rocm7.14.1_ubuntu24.04_py3.12_pytorch_release_2.12.0`, without amdsmi installed:

```python
from fastvideo.platforms import current_platform
print(type(current_platform).__name__)
```

## Test Results

<details>

```
ruff:      All checks passed!
yapf:      no diff
codespell: clean
mypy:      Success: no issues found in 1 source file
pytest:    3 passed, 15 warnings in 0.01s
```

Platform resolution on the MI355X box, FastVideo `37d06a8` without and with this change:

```
amdsmi importable: False
37d06a8 (before): current_platform -> CpuPlatform
this change:      current_platform -> RocmPlatform (AMD Instinct MI355X VF)
```

</details>

## Checklist

- [x] ruff, yapf, codespell and mypy run with the repository configuration on the touched files
- [x] I updated documentation if needed (none needed)
- [x] I considered GPU memory impact of my changes (none)


SolitaryThinker:
Pushed a9ba77225: covered the real ROCm device-node probe with tests and corrected stale docs/comments.

**Fixed**
- Docstring overclaimed what the probe proves -> now "has read/write permission on the AMD compute device node", with a note that permission bits are not proof an `open()` succeeds (`fastvideo/platforms/__init__.py:92-99`).
- Inline comment said "the check above always fails", but the gate is `if not is_rocm`, which also covers amdsmi present with zero handles -> reworded to "the amdsmi check above finds no device" (`fastvideo/platforms/__init__.py:117-122`).
- Redundant function-local `import os` in `cuda_platform_plugin` after the new module-level import -> deleted (`fastvideo/platforms/__init__.py:39`).
- The real `os.access("/dev/kfd", os.R_OK | os.W_OK)` line was never executed by the suite (all three tests stubbed the helper) -> added `test_device_node_probe_checks_the_amd_compute_node` (stubs `os.access`, asserts path and mask) and `test_amdsmi_device_wins_over_the_torch_fallback` (fails if the torch fallback runs when amdsmi reports a device), so a wrong path/mask or a reordered fallback now fails CI (`fastvideo/tests/platforms/test_rocm_detection_without_amdsmi.py:44-67`).

**Verified**
- `python3 -m py_compile` on both changed files: OK.
- 120-column check on both changed files: clean.
- Static review of the new tests against the code paths they exercise; no local pytest run (this sandbox has no project env) — CI is the gate.
- CI: `pre-commit` passed on the previous head. `buildkite/pr-fastcheck` stayed queued ("Build #1236 scheduled") for ~20 min of polling and never started, so it is reported pending, not red. Fresh runs are queued for a9ba77225.

**Left open**
- `os.access` proves permission bits, not that the node opens or that a GPU is usable: a HIP torch build with `HIP_VISIBLE_DEVICES=""`, or a container with `/dev/kfd` bind-mounted while the device cgroup denies access, now resolves `RocmPlatform` where it previously fell back to CPU. Hardening this (also requiring `/dev/dri/renderD*`, or honoring empty `*_VISIBLE_DEVICES`) is a detection-semantics decision, and the existing amdsmi path has the same blind spot, so it is left as a follow-up rather than changed here.
- `docker/rocm7_1.Dockerfile:46` installs the `.[rocm]` extra (`amdsmi`), the wheel this PR says breaks `import torch` on images without `libamd_smi.so`. Whether the base image ships that library is unverified here; dropping the extra is a separate change.
- The defensive `try/except` around `import torch` in the fallback is unreachable (torch is already imported via `fastvideo.utils`), but it matches the sibling `mps_platform_plugin`/`npu_platform_plugin` style, so it is kept.
- `hip_version` truthiness vs `torch.version.hip is not None` (as in `fastvideo/utils.py:75`) is cosmetic; left as written.
- The new test is not added to the macOS MLX workflow's explicit platform test list; the Linux unit lane runs `fastvideo/tests/platforms/` via `.buildkite/scripts/unit_test.sh:12`.

PR #1842, opened 2026-09-14
[bugfix] Allow MiniMax H3 frame padding at the 15-second limit

LeslieWylie:
## Purpose

An exact 15-second MiniMax H3 request (`num_frames=360` at 24 fps) is currently rejected: causal-VAE alignment rounds it to 362 frames, then the duration check compares 362 / 24 against 15. Inputs 346 through 362 therefore fail even though they are within the advertised duration.

## Changes

Derive the upper internal frame bound by applying the same alignment to `MAX_DURATION * FPS`, and hoist both aligned bounds next to `align_num_frames` in `packing.py`. This allows the final 362-frame bucket, including already-aligned requests, while rejecting the next bucket (363 rounds to 379). Keep the public 15-second constant, existing lower-bound behavior, and output geometry/trimming behavior unchanged.

The same bound is mirrored in the MLX runtime (`fastvideo/mlx_runtime/minimax_h3.py`, `minimax_h3_pipeline.py`) and the LoRA example (`examples/inference/lora/minimax_h3_lora_inference.py`) so a 15-second request is accepted on every entry point. Error messages now state the accepted frame range.

Add focused boundary cases for the exact maximum, an already-aligned maximum, idempotence, out-of-range buckets and non-integer input, plus `seconds="15"` at the OpenAI admission layer and the MLX `resolve_geometry` bound. Docs that advertised the 345-frame cap are updated to 362.

## Test Plan / Results

- Confirmed the failure on main bfc9c017977d1f428d43b6e580a0ff20503d442b using its pure frame helpers.
- Boundary assertions were re-validated after the cross-runtime changes against AST-extracted **actual** helper functions/constants (plain Python; the local sandbox has no torch/pytest). This bypasses GPU-dependent package imports; it does not validate package import or end-to-end inference.
- `uvx pre-commit run --files ...` passed on the original revision; the follow-up commits are checked by CI.
- GPU inference/SSIM was not rerun. A historical RTX PRO 5000 deployment accepted this same final bucket through a local maximum-duration workaround; that is supporting operational evidence, not a current-main GPU result.

For normal FastVideo CI environments, run:

```bash
pytest fastvideo/tests/stages/test_minimax_h3_input_preparation.py fastvideo/tests/entrypoints/test_openai_serving_engine.py -q
```

## Known gaps

[...]

PR #1857, opened 2026-09-16
[bugfix] Preserve video frames when decoded audio is shorter

LeslieWylie:
## Problem and change

The ffmpeg pipe writer uses `-shortest`, so decoded audio shorter than the video truncates generated video frames. This surfaced during H3 first/last-frame experiments: preserving the model's generated tail matters independently of input frame-count validation (#1842).

Make video duration authoritative: pad short audio with silence and bound output by the supplied frame count and frame-count/fps duration. Long audio is trimmed to that duration. Encoding quality defaults and the other save backends are unchanged.

## Validation

- A minimal ffmpeg reproduction with 48 frames at 24 fps and one second of audio saves only 24 frames with the old flags; the new flags preserve all 48 frames.
- Added a regression calling `_save_video_with_audio_ffmpeg_pipe` with short/equal/long audio (1/2/3 seconds) against a two-second video and checking actual ffprobe frame counts and both stream durations.
- Locally executed the unchanged source methods (`_audio_to_int16`, `_write_pcm_wav`, and the pipe writer) and this regression through an AST-isolated harness using NumPy inputs and real ffmpeg/ffprobe. All three cases pass. Running the same short-audio test against the base method fails (ffmpeg exits early / broken pipe). This isolates the CPU save path; the full FastVideo package test module and GPU generation were not run locally.
- The repository's targeted pre-commit hooks, including mypy, pass; diff whitespace checks pass.


LeslieWylie:
Could a maintainer review this focused video-tail preservation fix and help with the pending Fastcheck/required merge checks? Pre-commit passes, and the PR documents the short-audio reproduction plus short/equal/long-audio save-path validation and its limits. I would appreciate getting this merged once review and the required checks are satisfied. Thanks!

LeslieWylie:
CI follow-up: Buildkite pr-fastcheck build 1255 is still pending on the unchanged head ef4e2bb, while pre-commit has passed. Could a maintainer inspect/retry that Buildkite job and review the patch? The repository merge-gate workflow requires the ready label after review; this external contributor account has no write/triage permission to set it or trigger privileged checks. I am not treating skipped trigger jobs or the waiting Mergify protection as a test failure. The short/equal/long audio regression and local validation limits are documented in the PR.

PR #1859, opened 2026-09-16
[bugfix] FP8: gate the ROCm _scaled_mm path on CDNA4 (gfx950) instead of the capability tuple

stayinalive181:
## Purpose

`_supports_fp8_compute()` decides between the FP8 `torch._scaled_mm` path and the bf16 dequant fallback with the CUDA capability tuple (`>= (8, 9)`). On ROCm that tuple is the GFX generation: gfx942 (MI300X) reports `(9, 4)` and gfx950 (MI355X) reports `(9, 5)`, so every CDNA part passes the sm89 test. gfx950 does run the e4m3fn `_scaled_mm` through hipBLASLt (numbers below), but gfx942 only exposes the `fnuz` FP8 formats and gfx90a has no FP8 at all: those would fail inside the GEMM instead of taking the documented fallback.

This gates the ROCm branch on the architecture name (`gcnArchName`), admitting CDNA4 only, and states the requirement next to the sm89 one in the FP8 docs. CUDA behaviour is unchanged.

## What FP8 does on MI355X today (why the gate is worth having)

`torch._scaled_mm` (e4m3fn, 8192^3, ROCm 7.14.1 / torch 2.12.0+rocm7.14.1): bf16 1401 TFLOPS, FP8 tensorwise 2720 TFLOPS, FP8 rowwise 2379 TFLOPS.

FastH3 Preview v1 (MiniMax-H3, 4 forwards, VSA 0.9) at 1344x768, 345 frames, 4x MI355X, sequence-parallel 4, replicated DiT, regional DiT compile + compiled VAE, median of 3 windows (same seed and prompt; the bf16 pipeline is bit-deterministic run to run):

| DiT weights | E2E s | denoising s | video decode s | peak VRAM GB/GPU |
|---|---|---|---|---|
| bf16 | 20.94 | 14.74 | 3.23 | 152.9 |
| FP8 per-tensor, attention projections only (what `transformer_quant="FP8"` covers on H3 today, see #1780) | 20.72 | 14.52 | 3.24 | 144.2 |
| FP8 per-tensor, attention + feed-forward (#1780's suffix fix applied) | 20.14 | 13.95 | 3.25 | 133.0 |
| FP8 per-channel, attention + feed-forward | 20.58 | 14.40 | 3.21 | 132.9 |

The GEMMs themselves are 2x faster, but the eager activation quantization (`abs`/`amax`/`div`/`clamp`/`to`) costs about as much as the FP8 GEMM it feeds at H3's shapes (65,536 x 5,376: 1.06 ms of quantization next to a 1.25 ms `_scaled_mm`), which is why the window only moves 4%. A follow-up PR fuses that quantization into Triton kernels.

FP8 changes the sampled video the way any perturbation of a 4-step distilled trajectory does (PSNR vs the bf16 sample ~14 dB, i.e. a different sample of the same prompt); sharpness statistics of the frames are equal (variance of the Laplacian 375 vs 375-400).

## Test plan

- `pytest fastvideo/tests/ops/quantization/test_fp8_compute_gate.py` (12 tests: gfx950 admitted, gfx942 / gfx90a / RDNA rejected regardless of the capability tuple, CUDA sm89 boundary unchanged, no-accelerator case).
- Existing `fastvideo/tests/loader/test_fsdp_inference_quantization.py` and `fastvideo/tests/ops/quantization/` (minus the CUDA-only MXFP8 file) pass with the same 10 pre-existing NVFP4-wiring failures as `main` on a CPU-only checkout.
- Pre-commit (ruff, yapf, codespell, mypy) passes on the changed files.
- GPU: the table above ran on the branch's gate (4x MI355X, Daytona); MI300X was not available, so its fallback behaviour is by construction (the arch check) rather than measured.


SolitaryThinker:
Pushed a fix on top of your branch (`0b1a34de`) — the ROCm CDNA4 gate was applied to `fp8_config.py` only, but the same gate exists a second time in `fastvideo/layers/fp8linear.py` and protects a second live `torch._scaled_mm` call site.

**Why it matters:** `fp8linear.py:16-21` still used the bare capability tuple, so on MI300X (gfx942 reports `(9, 4)`) it returned `True` and the `fp8_qat_train` path (`fp8_qat_train_config.py:47` → `_LinearFWD8BWD16Fn.forward` → `fp8linear.py:88`) ran `torch._scaled_mm` with e4m3fn instead of taking the bf16 fake-quant fallback — the exact failure this PR fixes in `fp8_config.py`.

**The change** (2 files, +13/−11):

- `fastvideo/layers/fp8linear.py`: drop the duplicated `FP8_DTYPE` / `FP8_MAX` / `FP8_MIN_SCALE` and the local `_supports_fp8_compute`; import them from `fp8_config` instead. Docstring updated to say the fallback covers pre-CDNA4 ROCm, not just pre-sm89 CUDA.
- `fastvideo/tests/ops/quantization/test_fp8_compute_gate.py`: add `test_fp8linear_reuses_the_same_gate_and_constants`, asserting both modules share the same gate object and constants so the two paths cannot drift again.

**Verified locally:**

- With a fake gfx942 device, both gates now return `False` (the old formula returned `True`); gfx950 still returns `True`; the CUDA sm89 boundary is unchanged in both.
- `ruff check` passes on both files; no import cycle (`fp8_config` imports only `base_config` and `models.utils`).

**Not verified here:** the pytest run itself needs a GPU driver (`fastvideo/__init__.py` → `vmoba` → `fastvideo_kernel` → triton), so CI will be the first real execution of the new test.

If you would rather keep the gate local to `fp8linear.py`, the alternative is duplicating the `gcnArchName` branch there — but sharing one function is what keeps the two paths from drifting again.

SolitaryThinker (approved):

PR #1862, opened 2026-09-16
[bugfix] Enable SP for LongCat BSA refinement

Livinfly:
## Purpose

LongCat BSA refinement with `sp_size=2` fails in the first attention block: Q/K/V retain the full token sequence, but the BSA branch divides the declared H/W grid by the SP partition. On **2× NVIDIA H20**, each rank passes **184320 tokens** with a grid describing only **92160 tokens**, triggering `assert Tq * Hq * Wq == Sq`.

The refinement bucket also depends on SP size: the same 704×1280 request selects 768×1280 at SP2. The fix enables actual token sharding and keeps the output resolution consistent with SP1.

## Changes

- Split patch-embedded tokens using LongCat's balanced 2D H/W partitioning. Use FastVideo's existing all-to-all helpers to exchange sequence and head dimensions, restore global THW order before the existing RoPE/BSA, reverse the exchange after attention, and gather projected tokens before unpatchify.
- Keep bucket alignment based on the global BSA grid; broadcast initial refinement latents so all SP ranks start the scheduler from the same sample.
- Keep the layout helper in `longcat.py` and update `longcat_refine_init.py`. Release the temporary rank-major QKV buffer after restoring the global layout.

The new token sharding is scoped to BSA refinement inference with `num_cond_frames=0`. Other modes retain their existing execution paths without new unsupported-mode errors; this PR does not establish their SP scaling or training correctness. The existing weights, refinement LoRA and BSA kernels are reused.

### Environment

- **2× NVIDIA H20** (NVLink), BF16; one GPU for SP1.
- **Python 3.12.13**, **PyTorch 2.12.0+cu126** (CUDA 12.6).
- FlashAttention **2.8.3.post1**, Triton **3.7.0**.
- CUDA Graph and torch.compile disabled.

Unpatched FastVideo commit: `39718cd54d74e3ae869a18772ff5a24fb7920606`.

### Reproduction configuration

Use a LongCat checkpoint in FastVideo format with its refinement LoRA, and 93 stage1 RGB frames at 480×832. The following is the public-API configuration used by the test; `model_path` identifies that checkpoint, `frames` contains those decoded input frames, and `prompt` is held constant across runs.

<details>
<summary>SP2 configuration and generation request</summary>

```python
from pathlib import Path
from fastvideo.api import (
    CompileConfig, ComponentConfig, EngineConfig, GenerationRequest,
    GeneratorConfig, InputConfig, OffloadConfig, OutputConfig,
    ParallelismConfig, PipelineSelection, SamplingConfig,
)

config = GeneratorConfig(
    model_path=model_path,
    pipeline=PipelineSelection(
        components=ComponentConfig(
            lora_path=str(Path(model_path) / "lora/refinement"),
            lora_nickname="refinement",
        ),
        experimental={
            "attention_backend": "FLASH_ATTN",
            "enable_bsa": True,
            "bsa_sparsity": 0.9375,
            "bsa_chunk_q": [4, 4, 4],
            "bsa_chunk_k": [4, 4, 4],
        },
    ),
    engine=EngineConfig(
        num_gpus=2,
        parallelism=ParallelismConfig(
            tp_size=1, sp_size=2,
            hsdp_replicate_dim=1, hsdp_shard_dim=2,
        ),
        offload=OffloadConfig(
            dit=False, dit_layerwise=False, text_encoder=False,
            vae=False, pin_cpu_memory=False, lazy_module_load=False,
        ),
        compile=CompileConfig(
            enabled=False, vae_enabled=False,
            text_encoder_enabled=False, audio_vae_enabled=False,
        ),
    ),
)
request = GenerationRequest(
    prompt=prompt,
    negative_prompt="",
    sampling=SamplingConfig(
        height=704, width=1280, num_frames=186, fps=30,
        num_inference_steps=50, seed=42, guidance_scale=1.0,
        batch_cfg=False,
    ),
    inputs=InputConfig(stage1_video=frames),
    output=OutputConfig(output_path="output", save_video=True, return_frames=False),
    extensions={
        "t_thresh": 0.5, "spatial_refine_only": False, "num_cond_frames": 0,
    },
)
# Inside the script's __main__ entrypoint; FastVideo starts its own workers.
generator = VideoGenerator.from_config(config)
try:
    generator.generate(request)
finally:
    generator.shutdown()
```

The SP1 test uses `num_gpus=1`, `sp_size=1`, and `hsdp_shard_dim=1`. Worker CPU/CUDA RNGs are reset to seed 42 before each generation because VAE posterior/noise sampling uses worker-local RNGs.

</details>

### Correctness and timing checks

1. **Unpatched SP2:** run a real generation. A pass-through BSA wrapper records tensor/grid shapes, calls the original function with unchanged arguments, and preserves the original exception.
2. **Layout:** 14 CPU cases check each shard against independently sliced H/W regions, both layout permutations, and invalid grids. Coverage includes the full `(48, 44, 80)` grid and local H=22 at SP4/SP8, which must remain valid because BSA sees the restored global grid.
3. **Full checkpoint:** run unpatched SP1, patched SP1, and patched SP2 with the same checkpoint, LoRA, input frames, prompt and seed. Capture the first input and first/last full DiT predictions. A hook on the first transformer block checks the actual token count on every rank.
4. **Same-input reference:** after patched SP2 generation, both ranks run one additional SP1 forward on the exact saved first input, with the same weights and LoRA. This comparison is outside timing.
5. **Video:** compare every decoded frame of each patched output against the unpatched SP1 output using FFmpeg SSIM.
6. **Route regression:** construct a small two-block BF16 LongCat DiT directly on each GPU, with identical parameters and inputs. Compare dense, conditioned, single-frame and KV-cache forwards against the unpatched model under the same distributed runtime. Compare BSA SP2 against an SP1 BSA reference with the same weights. Also exercise BSA cache return, mixed BSA/dense blocks and a grad-enabled forward; these last checks do not validate training backward.

<details>
<summary>Key assertions used to inspect the generated artifacts</summary>

The saved tensor dictionaries contain `first_input`, `first_prediction`, and `last_prediction`. The following checks the observed exact agreement across the three full-checkpoint runs:

```python
import torch

native = torch.load("native-g1/measured-predictions.pt", map_location="cpu", weights_only=True)
sp1 = torch.load("integrated-g1/measured-predictions.pt", map_location="cpu", weights_only=True)
sp2 = torch.load("reviewed-g2/measured-predictions.pt", map_location="cpu", weights_only=True)
for reference, candidate in ((native, sp1), (sp1, sp2)):
    for key in ("first_input", "first_prediction", "last_prediction"):
        assert torch.isfinite(reference[key]).all()
        assert torch.isfinite(candidate[key]).all()
        torch.testing.assert_close(candidate[key], reference[key], rtol=0, atol=0)
```

The block pre-hook records `values[0].shape[1]` on every forward. Each rank checks:

```python
# Five actual refinement forwards in this configuration.
assert local_block_tokens == [168960 // sp_size] * 5
```

Video comparison:

```bash
ffmpeg -i patched-measured.mp4 -i native-sp1-measured.mp4 \
  -lavfi ssim -f null -
```

For the small BSA check, both ranks use a `(T, H, W) = (8, 12, 16)` token grid. A pre-hook observes the first block's token count. Selecting the model's SP1 path leaves the actual RoPE and BSA kernels unchanged:

```python
from unittest.mock import patch
import fastvideo.models.dits.longcat as longcat

# Inside the initialized two-rank runtime and forward context.
with torch.no_grad():
    with patch.object(longcat, "get_sp_world_size", return_value=1):
        reference = model(**inputs)
    output = model(**inputs)
torch.testing.assert_close(output, reference, rtol=0, atol=0)
assert observed_sp2_block_tokens == [768]  # 1536 tokens split across two ranks
```

</details>

Each successful configuration performs **one full warmup followed by one measured generation**, with five actual refinement steps. CUDA events bracket each native denoising iteration, including scheduler updates, and are read after generation. Reported DiT step time is the median of the five per-step maximum-rank times. Video time is the measured request's wall-clock duration, including VAE work and video saving. Tensor export and the extra SP1 reference forward are excluded.

### Actual unpatched SP2 failure

Excerpt from the worker's actual SP2 generation traceback. Paths are relative to the repository; log prefixes and caret markers are removed. Omitted frames are marked with `...`, including PyTorch dispatch and the pass-through shape-diagnostic wrapper.

```text
Traceback (most recent call last):
  ...
  File "fastvideo/pipelines/stages/longcat_denoising.py", line 126, in forward
    noise_pred = self.transformer(
  ...
  File "fastvideo/models/dits/longcat.py", line 1084, in forward
    block_out = block(
  ...
  File "fastvideo/models/dits/longcat.py", line 819, in forward
    attn_result = self.self_attn(
  ...
  File "fastvideo/models/dits/longcat.py", line 398, in forward
    out = flash_attn_bsa_3d(q_bsa,
  ... (pass-through diagnostic wrapper frame omitted)
  File "fastvideo/third_party/longcat_video/block_sparse_attention/bsa_interface.py", line 710, in flash_attn_bsa_3d
    assert Tq * Hq * Wq == Sq
AssertionError
```

Both worker failures propagate to the caller:

```text
RuntimeError: RPC 'execute_forward' failed: worker 0: AssertionError: ; worker 1: AssertionError:
```

The pass-through diagnostic wrapper recorded the following inputs in the same run, then called the original BSA function with unchanged arguments:

```text
NATIVE_SP2_DIAGNOSTIC={"rank": 0, "boundary": "bsa_input", "q_shape": [1, 32, 184320, 128], "k_shape": [1, 32, 184320, 128], "v_shape": [1, 32, 184320, 128], "latent_shape_q": [48, 48, 40], "latent_shape_k": [48, 48, 40], "declared_grid_tokens": 92160, "actual_q_tokens": 184320}
NATIVE_SP2_DIAGNOSTIC={"rank": 1, "boundary": "bsa_input", "q_shape": [1, 32, 184320, 128], "k_shape": [1, 32, 184320, 128], "v_shape": [1, 32, 184320, 128], "latent_shape_q": [48, 48, 40], "latent_shape_k": [48, 48, 40], "declared_grid_tokens": 92160, "actual_q_tokens": 184320}
```

Both ranks pass Q/K/V of shape `[1, 32, 184320, 128]`, but `48 × 48 × 40 = 92160`. The native bucket selected 768×1280; it was not overridden during this reproduction.

### Full-checkpoint results

The final SP2 run and route regressions use the submitted source tree.

| Implementation | Configuration | DiT step (s) | Video generation (s) |
|---|---|---:|---:|
| Unpatched | 1× H20 / SP1 | 48.679 | 287.646 |
| Unpatched | 2× H20 / SP2 | BSA assertion | No video |
| Patched | 1× H20 / SP1 | 48.715 | 287.550 |
| Patched | 2× H20 / SP2 | 25.202 | 169.960 |

Patched SP2 achieves **1.93× DiT-step speedup** and **1.69× video-generation speedup** relative to patched SP1. These are single measured generations after warmup.

- Initial latents and first/last DiT predictions: **bitwise identical** across unpatched SP1, patched SP1 and patched SP2; maximum error and RMSE are 0.
- Additional same-input SP1 reference: maximum error and RMSE are **0 on both ranks**.
- Both patched videos versus unpatched SP1: **SSIM 1.0 (infinite dB)** across all 186 decoded frames, at 704×1280.
- Block tokens per rank: **168960 at SP1**, **84480 at SP2**.
- Peak allocated memory: **67.10 GiB at SP1**, **61.37 GiB per rank at SP2**.

### Final route regression

All **9 cases pass on both ranks**, with exact output agreement. BSA SP2 processes **768 tokens per rank**, compared with **1536** in its SP1 reference. The other eight cases execute their existing full-sequence routes without the newly added errors. This verifies forward compatibility on the small model, not full-pipeline support or SP acceleration for every mode.

## Checklist

- [x] Reproduced the original SP2 failure on 2× H20.
- [x] Validated full-checkpoint SP1/SP2 outputs and all decoded video frames.
- [x] Passed 14 CPU layout cases.
- [x] Passed nine two-rank route cases.
- [x] Passed `pre-commit run --files fastvideo/models/dits/longcat.py fastvideo/pipelines/stages/longcat_refine_init.py`.
- [x] Measured per-rank memory usage.

This PR was developed with assistance from OpenAI Codex and has been personally reviewed by Livinfly.


aryan5v:
@greptileai

PR #1866, opened 2026-09-17
[kernel] sm_100a CUDA backward for VSA block-sparse attention, 128-token blocks

lpc0220:
## Purpose

Adds the 128-token-block sm_100a CUDA backward for VSA block-sparse attention, the sibling of the 64-token backward in #1819. With `FASTVIDEO_VSA_SM100A=1`, 128-token metadata (currently forward-only on the native route, backward raised) now gets a native backward that is 2-4.5x faster than FastVideo's default 128 route (the 64-token Triton kernels over the expanded map).

## Changes

- New `csrc/attention/block_sparse_bwd_kernel_sm100a_blk128.cuh`, `block_sparse_bwd_launch_sm100a_blk128.cuh`, `block_sparse_bwd_blk128_sm100a.cu` (namespace `vsa_bwd_blk128`, exports `block_sparse_sm100a_blk128_bwd`); one CTA per (batch, head, kv128 block), preprocess / main / postprocess chained with programmatic dependent launch; `variable_block_sizes` masks padded kv rows; empty kv blocks get exact-zero dK/dV rows in-kernel.
- `block_sparse_bwd_sm100a.cu`: `VSA_BLK128` switch (same pattern as the forward's `block_sparse_blk128_sm100a.cu`); 64-token behaviour unchanged.
- `primitives.cuh`: additions only (+27 lines: cp.async.cg, griddepcontrol); forward / 64-token SASS untouched. CMake + `common_extension.cpp` register the new TU / op.
- Python: `block_sparse_attn_bwd_sm100a` picks the op by the metadata's block size; `_backward_sm100a` no longer refuses 128-token metadata when the op is built (unsupported 128 still raises: the Triton backward is 64-token only).
- Tests: `test_block_sparse_bwd_sm100a.py` runs every case for both block sizes; `test_block_sparse_sm100a_dispatch.py::test_blk128_backward` checks the route runs the CUDA backward (Triton backward monkeypatched to fail) and matches the Triton 128 wrapper's grads.

## Test Plan

```bash
TORCH_CUDA_ARCH_LIST=10.0a pip install . --o-build-isolation
python -m pytest tests/test_block_sparse_bwd_sm100a.py -v
python -m pytest tests/test_block_sparse_sm
python -m pytest fastvideo-kernel/tests/test_vsa_forward.py -v
```

## Test Results

GB200 (sm_100a), torch 2.13 cu130: bwd 32 pblock count), dispatch 11 passed, forward 1passed.

Backward only, B=1 H=8 D=128, 25 % density, shared inputs, same window, TFLOPS:

| S | triton-64 | sm100a-64 | triton-128 | sm100a-128 | sm100a-128 / triton-128 | sm100a-128 / sm100a-64 |
|---|---|---|---|---|---|---|
| 4096 | 92.9 | 283.1 | 90.3 | 403.8 | 4.5x | 1.43x |
| 16384 | 343.5 | 688.1 | 335.5 | 1058.4 | 3.2x | 1.54x |
| 32768 | 403.7 | 771.7 | 400.7 | 1243.5 | 3.1x | 1.61x |
| 65536 | 426.4 | 780.4 | 422.0 | 1158.6 | 2.8x | 1.48x |
| 131072 | 435.1 | 696.7 | 431.9 | 1006.8 | 2.3x | 1.45x |
| 524288 | 422.0 | 734.2 | 409.7 | 835.0 | 2.0x | 1.14x |

sm100a-128 and triton-128 grads agree to reat every size.

<details>

```
tests/test_block_sparse_bwd_sm100a.py
tests/test_block_sparse_sm100a_dispatch.py     11 passed
fastvideo-kernel/tests/test_vsa_forward.py
```

</details>

## Checklist

- [x] I considered GPU memory impact of my one fp32 Delta scratch, same as the 64-tokenbackward minus its Q^T/dO^T transposes)

sm_103a: compiles through the same arch guard as the 64-token backward; validated on GB200 only.


fastvideo-kernel/csrc/attention/block_sparse_bwd_launch_sm100a_blk128.cuh
```diff
@@ -0,0 +1,231 @@
...
+  if (!args.dq || !args.dk || !args.dv) {
+    return cudaErrorInvalidValue;
+  }
+  if (!args.k2q_idx || !args.k2q_num || !args.variable_block_sizes) {
+    return cudaErrorInvalidValue;
+  }
+  if (!args.dqaccum || !args.delta) {
+    return cudaErrorInvalidValue;
+  }
+  return cudaSuccess;
```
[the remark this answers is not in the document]
lpc0220: `k2q_idx`/`k2q_num` come from FastVideo's own `invert_indices`, which cannot produce a count above the row width or an id outside `[0, num_blocks)`; the op is not a user-facing API, and a host-side value check would cost a device-to-host sync per call. This is the same contract as the merged 64-token backward (#1819) and the Triton kernels, which index with these tensors unchecked as well. Keeping it as is.

tests/test_block_sparse_bwd_sm100a.py
```diff
@@ -20,32 +20,38 @@
...
     not torch.cuda.is_available() or torch.cuda.get_device_capability() not in {(10, 0), (10, 3)}
     or not bwd._HAS_VSA_BWD_SM100A,
     reason="requires a data-center Blackwell GPU (sm_100a/sm_103a) and a fastvideo_kernel "
-    "extension built with block_sparse_sm100a_bwd",
+    "extension built with block_sparse_sm100a_bwd / block_sparse_sm100a_blk128_bwd",
 )
 
 
-def make_case(num_blocks=8, topk=4, heads=4, batch=1, ragged=False, seed=0, kv_pool=None):
+@pytest.fixture(params=bwd.BLOCKS, ids=lambda b: f"blk{b}")
```
alexzms, line 40: [P2] Please include the new blk128 regression coverage in GB200 CI.

Both this module and `tests/test_block_sparse_sm100a_dispatch.py` live under the root `tests/` directory, while the active kernel lane (`.buildkite/scripts/lanes/kernel_tests.sh`) runs only `pytest fastvideo-kernel/tests/ -vs`. These new gradient and dispatch cases therefore are not collected by that lane.

Please move the modules into the collected directory or explicitly add them to the existing GB200 kernel lane, and provide passing CI results for the updated commit. On that supported-hardware lane, also require `block_sparse_sm100a_blk128_bwd` to be present and exercised: the current fixture can skip every blk128 case when the op is missing, so simply collecting the files is not enough. Skips can remain for unsupported hardware/builds.

fastvideo-kernel/csrc/attention/block_sparse_bwd_launch_sm100a_blk128.cuh
```diff
@@ -0,0 +1,231 @@
...
+  if (make_tma_tile_units(&tq, args.q, B, H, S, Q_TILE) != cudaSuccess ||
+      make_tma_tile_units(&tk, args.k, B, H, S, KV_TILE) != cudaSuccess ||
+      make_tma_tile_units(&tv, args.v, B, H, S, KV_TILE) != cudaSuccess ||
+      make_tma_tile_units(&tdo, args.dout, B, H, S, Q_TILE) != cudaSuccess ||
+      make_tma_tile_units(&tdk, args.dk, B, H, S, KV_TILE) != cudaSuccess ||
+      make_tma_tile_units(&tdv, args.dv, B, H, S, KV_TILE) != cudaSuccess) {
+    return cudaErrorInvalidValue;
+  }
+
+  auto main_kernel        = vsa_bwd_main_kernel<VSA_BHSD, Sched::NON_PERSISTENT>;
```
alexzms, line 170: Design question: why retain the persistent/CLC scheduler variants in this PR when this launcher always instantiates `Sched::NON_PERSISTENT` and passes `nullptr` for `workitem_remap` below? The kernel also defines `KERNEL_SCHED`, but it is not used to select the instantiation.

Is there a current consumer or validation path for those variants that I missed? If they are only for future work, I would prefer deferring them until they are enabled and tested, to keep the synchronization code in this low-level kernel change easier to review. This is a request for clarification; the CI coverage gap is the reason for requesting changes.

alexzms (changes requested):
The low-level kernel scope makes sense. Requesting changes to make the new 128-token backward correctness and dispatch coverage part of GB200 CI, with passing results before merge. I also left a question about retaining the currently unreachable scheduler variants.

Validation from this review: the new kernel and launcher compiled in an isolated harness with CUDA 12.9 for both sm_100a and sm_103a. GPU numerical tests and performance measurements were not rerun.

SolitaryThinker (commented):
**Static review — pinned revision.** Head `a532d7ea810ffbff33315456cae388538d33af2a`, base `c4824c776433555ab655a0dfc43c25f4dac20829`. Read-only and static-only: **tests were not run** — no build, no code execution, no GPU, no benchmark. Every finding cites the pinned SHA. This is a COMMENT review, not merge authorization.

Scope: all 11 changed files (+1482/−105).

---

### What I checked and found consistent

I traced the new main kernel end to end against the proven 64-token backward (#1819) and the sm_100a forward. No correctness defect found in the following:

- **Barrier protocol.** Every producer/consumer pair, arrival count and phase-parity tracker lines up. The subtle cases hold: `sDST` is single-buffered and the `gemm4_dq` read for q block *j* is covered by the `full_bar_dpt` commit issued for *j+1*, so the softmax warp cannot overwrite `dS^T` mid-read; `tmem_dq` aliases `tmem_dpt` and the MMA warp waits on `empty_bar_dq` (the epilogue's post-`tcgen05_wait_ld` arrival) before the next dPT GEMM; `tmem_st` is rewritten for *j+1* only after the previous iteration's `full_bar_pt` wait.
- **Descriptor arithmetic.** The ST/dPT GEMMs use the forward's `bmm1` convention (`tb=false`, `LBO=16`, K sub-tiles stepped through the descriptor address), the dV/dK GEMMs use the forward's `bmm2` convention (`tb=true`, `LBO` = N sub-tile offset), and `gemm4_dq` matches the 64-token backward (`ta=tb=true`, `LBO=DST_TILE_BYTES`). The 32-byte K-atom stepping inside a 128B-swizzled row is the same idiom the forward and blk64 already use, and the bf16x2 tmem A-operand column map (`q_half*64 + ki*8`) matches the P^T/dS^T store offsets.
- **Budgets.** `SMEM_TOTAL` = 231 200 B against the 232 448 B sm_100a limit (1 248 B headroom; the smem layout ends at 230 660 B). `TMEM_TOTAL` = 512 columns exactly. `setmaxnreg` allocations sum to 60 672 registers against the 65 536 budget. Both are enforced loudly (`cudaFuncSetAttribute` + `TORCH_CHECK`), not silently.
- **`dqaccum` layout.** The epilogue's `[hd_slice][v4][row][4]` push, the post-process read and the final swizzled `dq` write agree; the post-process swizzle/unswizzle pair is an exact inverse.
- **Empty-kv-block path.** The epilogue zeroes the two dQ staging buffers and TMA-stores them to dk/dv; the `static_assert` ties the buffer size to one 128×64 bf16 tile and the 128-thread × 16-`uint4` zeroing covers exactly 32 768 B.
- **PDL usage is correct.** I initially suspected the preprocess's `griddepcontrol_launch_dependents()` (kernel line 910, placed before the delta loop) raced the main kernel's delta read. The PTX ISA refutes that: `griddepcontrol.wait` "causes the executing thread to wait until all prerequisite grids in flight have completed and all the memory operations from the prerequisite grids are performed and made visible to the current grid". `launch_dependents` is a scheduling hint only. The same guarantee covers the main→postprocess `dqaccum` hand-off.
- **64-token path untouched.** The `#if !VSA_BLK128` guards keep `qt`/`dot`/`order` out of the 128-token TU; `primitives.cuh` additions are purely additive (no duplicate symbols).

---

### Findings

#### 1. PDL visibility covers only the load warp; three other warp roles read caller metadata before any wait

`fastvideo-kernel/csrc/attention/block_sparse_bwd_kernel_sm100a_blk128.cuh:221, 463, 609, 653, 784, 790`

`griddepcontrol_wait()` is executed only by the load warp (line 221). The MMA (463), softmax (609) and epilogue (784) warps call `decode_workitem()` → `k2q_num[item]`; the softmax warp also reads `variable_block_sizes[it.kv_block_id_in_seq]` (653) and the epilogue warp reads `it.local_k2q_idx[j]` (790) — all before the first barrier that is transitively gated by that wait. Because all three launches carry `cudaLaunchAttributeProgrammaticStreamSerialization`, the main kernel's CTAs are resident and running while the preprocess grid is still executing, so those reads sit outside the visibility guarantee `griddepcontrol.wait` establishes.

A stale read of `k2q_num`/`k2q_idx`/`variable_block_sizes` would silently mis-drive the work item (wrong q-block list, wrong padded-row mask) and produce wrong gradients with no error. **This is an unresolved hypothesis, not a demonstrated failure** — the metadata is written several ops before the preprocess so the writes are complete by then, and your GB200 runs did not show one. Note the 64-token backward has no PDL at all, so the hazard is new with this file.

Fix: hoist the wait above the warp-role dispatch so every warp executes it, or move the metadata reads behind the first barrier each warp already waits on (compute `kv_row_valid` after the first `full_bar_st` wait; read `local_k2q_idx[j]` after the `full_bar_dq` wait). Dropping the PDL attribute for this kernel — matching blk64 — is also acceptable.

#### 2. ~200 lines of CLC/persistent scheduler scaffolding is unreachable in every build

`fastvideo-kernel/csrc/attention/block_sparse_bwd_kernel_sm100a_blk128.cuh:22, 537`; `block_sparse_bwd_launch_sm100a_blk128.cuh:170`

Same conclusion as @alexzms's design question — I reached it independently, so recording the measurement: the file carries `enum class Sched` + `KERNEL_SCHED`, a CLC scheduler warp (~30 lines at 537), persistent work-item loops in all four warp roles, `empty_bar_kv`/`empty_bar_epi`, `workitem_remap`/`total_workitems` and the `clc_*` barriers. The only instantiation is `vsa_bwd_main_kernel<VSA_BHSD, Sched::NON_PERSISTENT>` (launcher:170), nothing in the repo sets `KERNEL_SCHED`, and the launcher passes `workitem_remap = nullptr`. So `PERSISTENT` and `CLC` are compile-time false everywhere: ~136 lines inside `if constexpr` blocks plus ~70 lines of surrounding scaffolding — roughly 20 % of the new file — are dead and untested. Several non-persistent behaviours are only correct *because* a `PERSISTENT` branch is compiled out, which makes the live path harder to review.

#### 3. Capability documentation still says the backward is 64-token-only / forward-only

`fastvideo-kernel/python/fastvideo_kernel/block_sparse_attn_sm100a.py:11`, `fastvideo-kernel/python/fastvideo_kernel/block_sparse_attn.py:55`, `fastvideo/attention/backends/video_sparse_attn_h3.py:422`

Three places describe the sm_100a backward as 64-token-only or the extension as forward-only: the forward module docstring ("`block_sparse_attn_bwd_sm100a`, 64-token blocks, sm_100a devices only"), `_force_sm100a`'s docstring ("the extension is forward-only, so this routing pairs it with the Triton backward"), and the H3 backend's gate message ("the sm_100a/sm_103a kernel is forward-only; grad paths keep Triton"). This PR updates the adjacent comment block in `block_sparse_attn.py` but not these, so a reader of the dispatch path is told the 128-token backward does not exist.

#### 4. `is_supported` does not mirror the launcher's new `batch * heads <= 65535` precondition

`fastvideo-kernel/python/fastvideo_kernel/block_sparse_attn_bwd_sm100a.py:87`; `block_sparse_bwd_launch_sm100a_blk128.cuh:87`

The 128-token launcher adds that check (needed because `gridDim.y = B*H`), but the Python predicate does not, so an over-limit input reaches the binding and raises instead of being refused by the predicate. Low impact — unrealistic configuration, loud failure, and 128-token metadata has no Triton fallback anyway — but the predicate is documented as the capability contract and should match it.

---

### Overlap with the existing review

- @alexzms already requested GB200 CI coverage for these tests and asked about the scheduler variants. I reached the same two conclusions independently; findings 1 and 3 are additional, and finding 4 is minor.
- @alexzms validated that the kernel and launcher compile for both sm_100a and sm_103a with CUDA 12.9 in an isolated harness. I built nothing, so my sm_103a caveat is now only about *numerical* validation.
- Greptile's metadata-bounds point was answered by the author on the `invert_indices` contract and is consistent with what I see: the kernel treats `k2q_num`/`k2q_idx` as trusted internal metadata, same as blk64 and Triton. I did not treat it as a finding.

---

### Limitations

- Tests NOT run. No build, no execution, no GPU, no benchmark. The TFLOPS table and the "grads agree" claim in the description are unverified.
- MMA smem-descriptor / instruction-descriptor encodings were checked for *consistency* against the forward and blk64, not executed.
- Numerical accuracy, the empty-kv-block zeroing, padded-row masking and dQ reduce-add ordering were reasoned about statically; only GPU tests can confirm them.
- No in-repo production caller reaches this backward: the H3 backend refuses the sm_100a route whenever grad is enabled, and `VSA_H3_TILE_SHAPES` only offers 256- and 64-token tiles, so 128-token metadata is not produced by FastVideo. Reachability is limited to direct kernel callers.
- The description's "Test Results" section reads "bwd 32 pblock count), dispatch 11 passed, forward 1passed" and the collapsed "Test output" block contains only the command lines, so the pass counts for the new parametrized cases are not visible; the test tolerance comment also asserts "the blk128 cases sit in the same range" without numbers.

**Assessment: ready to merge with fixes.** The dataflow, barrier protocol, descriptor arithmetic and resource budgets are internally consistent and I found no demonstrated correctness defect; the PDL usage that first looked racy is correct per the PTX ISA. The fixes are small — put the metadata reads behind the PDL wait (or drop PDL), remove the unreachable scheduler scaffolding, refresh the three docstrings, and close the CI coverage gap @alexzms asked for.

lpc0220:
> Assessment: ready to merge with fixes. The dataflow, barrier protocol, descriptor arithmetic and resource budgets are internally consistent and I found no demonstrated correctness defect; the PDL usage that first looked racy is correct per the PTX ISA. The fixes are small — put the metadata reads behind the PDL wait (or drop PDL), remove the unreachable scheduler scaffolding, refresh the three docstrings, and close the CI coverage gap @alexzms asked for.

- PDL: `griddepcontrol.wait` now runs in every warp right after barrier init, before any
  gmem read (k2q metadata, Delta, dqaccum); the trigger moved after the epilogue's final
  `bar_sync<10>`, so it fires once the CTA's work is complete (256dc37f).
- Docstrings: the three forward-only / 64-token-only claims are gone (a00e9bf1).
- `is_supported` refuses batch * heads > 65535, with a test (63816960).
- CI: both test modules moved into `fastvideo-kernel/tests/` so the GB200 lane collects
  them (a24b6cd3); from the repo root against the rebuilt wheel: 32 passed / 1 skipped and
  11 passed, `test_vsa_forward.py` 1 passed; pre-commit clean on the changed files.

Scheduling scaffolding is we'd like to keep. They are not speculative. As measured, CLC is at parity with NON_PERSISTENT on the 32k-64k grid; the persistent form is the base for L2-locality oriented scheduling we are preparing for S >= 262k, where the non-persistent grid loses L2 residency of dqaccum. Keeping them keeps future PRs small diffs instead of re-imports. Cost to this PR: none in the binary -- the `if constexpr` branches compile out and
the shipped SASS is the same as a stripped copy -- and nothing in the live path depends on a
PERSISTENT branch being present. WDYT?  If it helps review, we can add a header note stating that only
NON_PERSISTENT is instantiated here and the other two are validated out of tree.

